PR #11, opened 2014-08-05
Representing trials in elephant 

alperyeg:
This pull is meant to initiate a discussion on a central topic in the design of
elephant: How to represent trials (for analysis functions that are trial based)
using the neo framework, and how to efficiently and flexibly deal with trials
from the user's and method-programmer's perspective.

In previous discussions, it was stimulated that trials could be represented as
Segments. This PR contains code to propose one approach to deal with such a
trial representation by introducing a class (NeoInfo) that allows to access and
interpret trials stored in such a fashion.

Of course, other approaches are equally possible. E.g., trials could be encoded in
lists or dictionaries, or simply as EpochArrays accompanying a Block, Segment
or AnalogSignal. All of such representations are useful, and we believe it is
probably good to allow for some of these possibilities to coexist. The idea was
that something like the NeoInfo class could be used to unify access to trials
stored in these various representations, e.g., to have something like (in
pseudo code):
  x <- object # object containing trials represented in a supported way
  analogsig <- NeoInfo(x).get_analogsignal_trial(1)

(Currently, NeoInfo only supports trials as Segments).

A Fano factor function is supplied as a simple use case of the class for
illustration (see provisional test case in test_statistics.py, fano_TestCase).

The code in this PR is still "raw" in many ways, but is meant to serve as a
basis for discussion for a perspective on how trials may be represented. 

We realize that some of the functionality in this PR has overlap with some of the new functionalities of Neo and neo_tools.py introduced recently.

### DIFF
--- a/elephant/neoinfo.py
+++ b/elephant/neoinfo.py
@@ -1,0 +1,2172 @@
+import numpy as np
+from neo.core import Block, Segment, SpikeTrain, AnalogSignal, Epoch, \
+    RecordingChannel, RecordingChannelGroup, Unit
+
+
+class NeoInfo(object):
+    """
+    This class introduces a trial concept for elephant based on the Neo
+    framework, and provides general information about Neo objects.
+
+    In its simplest use, NeoInfo extracts information about arbitrary Neo
+    objects. From the given input it traverses the hierarchical Neo tree and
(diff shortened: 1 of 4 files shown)


alperyeg:
I am closing this PR now.

PR #14, opened 2015-01-19
coefficient of local variation (lv)

alperyeg:
This pull request concerns the coefficient of local variation (lv) as described in: 
Shinomoto, S., Shima, K., & Tanji, J. (2003). Differences in spiking patterns among cortical neurons. Neural Computation, 15, 2823–2842.

This index is a simple, frequently used measure of spike train variability that takes into account global rate fluctuations.


elephant/statistics.py
```diff
@@ -181,4 +181,56 @@ def fanofactor(spiketrains):
...
+    ..[1] Shinomoto, S., Shima, K., & Tanji, J. (2003). Differences in spiking
+    patterns among cortical neurons. Neural Computation, 15, 2823–2842.
+
+
+    """
+    # convert to array
+    v = np.array(v)
+
+    # ensure we have enough entries
+    if v.size<2:
```
toddrjen: should be `if v.size < 2:` (added space)

elephant/statistics.py
```diff
@@ -181,4 +181,56 @@ def fanofactor(spiketrains):
...
+        return np.nan
+
+    # calculate LV
+    lvar = np.mean(
```
toddrjen: A few things here:

You can avoid assigning the variable here by just doing `return np.mean(...`
`v[:-1] - v[1:]` is the same as `-1*np.diff(v)`, which simplifies to `np.diff(v)` due to squaring.
Mean is associative, so you can move the `3` outside the mean.
Further, due to power identities, you can move the division inside the power.
So this line would simplify to:
```Python
return 3.*np.mean(np.power(np.diff(v)/(v[:-1] + v[1:]), 2))
```
Which in my test, where `v=np.random.rand(10000)`, it reduced the calculation time by more than a half while giving the same result.

elephant/statistics.py
```diff
@@ -181,4 +181,56 @@ def fanofactor(spiketrains):
...
+
+    References
+    ----------
+    ..[1] Shinomoto, S., Shima, K., & Tanji, J. (2003). Differences in spiking
+    patterns among cortical neurons. Neural Computation, 15, 2823–2842.
+
+
+    """
+    # convert to array
+    v = np.array(v)
```
toddrjen: `np.asarray` is much, much faster than `np.array` when the result is already an array, and only slightly slower when it isn't.

elephant/test/test_statistics.py
```diff
@@ -301,5 +301,37 @@ def test_fanofactor_list_same(self):
...
+
+    def test_lv_with_plain_array(self):
+        seq = np.array(self.test_seq)
+        assert_array_almost_equal(es.lv(seq), self.target, decimal=9)
+
+    def test_lv_with_list(self):
+        seq = self.test_seq
+        assert_array_almost_equal(es.lv(seq), self.target, decimal=9)
+
+    def test_lv_empty(self):
```
toddrjen: You don't test for the case where `v.size == 1`.

elephant/statistics.py
```diff
@@ -181,4 +181,56 @@ def fanofactor(spiketrains):
...
+
+    Parameters
+    ----------
+
+    v : quantity array, numpy array or list
+        Consecutive time intervals
+
+    Returns
+    -------
+    lvar : float or nan
```
toddrjen: `nan` is a `float`.

elephant/statistics.py
```diff
@@ -181,4 +181,56 @@ def fanofactor(spiketrains):
...
+           The LV of the inter-spike interval of the input sequence.
+           If an empty list is specified, or if the sequence has less
+           than two one entries, F:=nan.
```
toddrjen: I think "two one" is a typo.  And "has less than two entires" includes an empty sequence, so I think the first part is redundant.  So I think it should be "If the sequence has less than two entries, F:=nan"

toddrjen:
Overall, I think it is good.  Some minor improvements and typos.

The one outstanding issue I see is that the behavior is undefined for multidimensional arrays, both in the documentation and the function itself.  It may or may not work, depending on the situation, but if it does work the behavior can be unexpected.  I think you should either properly handle the n-dimensional case (which isn't hard to do in this function), or you should specify that only 1-dimensional arrays are allowed and raise an exception if a higher-dimension array is provided.

toddrjen:
Also, please squash the commits into a single commit.

alperyeg:
@toddrjen Thanks for the input, I adjusted the function and the test. 
I
 decided to stick to 1-dim input and pointed out in the docstring that only vectors are allowed. If a matrix is supplied numpy will raise a ValueError , while calculating the LV.  
Another point came up, if the input consists only of integers, the calculation will not provide the right result, because of the (integer) division. So I cast the input to float in the beginning.

apdavison:
Note that if you use (`from __future__ import division`) (and I think probably every file in Elephant should use this, as a safety measure) the cast to float is not necessary.

toddrjen:
Integer division isn't an issue because we use `from __future__ import division`.  I tested the results and it works fine.  Plus you get a performance hit for using `np.asarray(v, float)`

```Python
>>> from __future__ import division, print_function
>>> import numpy as np
>>> 
>>> v = np.random.randint(-1000, 1000, 100000)
>>> v.dtype
dtype('int64')
>>> v2 = np.asfarray(v)
>>> (np.diff(v)/(v[:-1] + v[1:])).dtype
dtype('float64')
>>> (np.diff(v)/(v[:-1] + v[1:]) == np.diff(v2)/(v2[:-1] + v2[1:])).all()
True
>>> %timeit -n 1000 np.asarray(v);3.*np.mean(np.power(np.diff(v)/(v[:-1] + v[1:]), 2))
1000 loops, best of 3: 5.56 ms per loop
>>> %timeit -n 1000 np.asarray(v, float);3.*np.mean(np.power(np.diff(v)/(v[:-1] + v[1:]), 2))
1000 loops, best of 3: 5.98 ms per loop
```

apdavison:
Alper: are you still planning to squash the commits into a single commit, as @toddrjen requested?

alperyeg:
@apdavison yes will do. Do I have to use git rebase ? I am not sure how to do it properly.

alperyeg:
Sorry something went terribly wrong. I used `git rebase --i --root` and squashed everything. But after a conflict and solving it, it seems all the commit logs were reintroduced. What is the best way to go now? Try to squash again (if you can tell me how :)) or new PR?

alperyeg:
The problem is solved. If there are no objections anymore, can we merge the PR?

PR #15, opened 2015-01-28
Binning class and PSTH

alperyeg:
This pull request contains two components. First we integrate a binning class `Binned` in the `conversion` module, that provides a representation of binned (clipped, unclipped) spike trains based on sparse matrices, which is an important feature for large sets of spike trains in combination with fine-grained binning. This binning class therefore acts more as a specialized data type within elephant. The advantage of this class is that it takes care of efficiently handling the sparse matrix management for the user (optionally keeping the “non-sparse” matrices in memory) and handles the clipping, but also keeps the time axis in order (similar to the `AnalogSignal` class of `Neo`). Please note: The existing binning function in elephant (`conversion.binarize`) has intentionally not been touched in this PR.

Second, we added a short function `psth` to the to `statistics` module that calculates the peristimulus time histogram (PSTH). It is meant as a very simple example of how to use the binning class in a function (however, more complex functions such as the correlation coefficient can exploit the benefits of the sparse matrix representation offered by the Binning class more dramatically).

### DIFF
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -8,6 +8,9 @@ docstring goes here.
 
 from __future__ import division, print_function
 
+import neo
+import scipy
+import scipy.sparse as sps
 import numpy as np
 import quantities as pq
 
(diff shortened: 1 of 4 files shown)


elephant/conversion.py
```diff
@@ -152,3 +155,706 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        `t_stop` from.
+
+    Returns
+    -------
+    start : quantities.Quantity
+        Start point extracted from input :attr:`spiketrains`
+    stop : quantities.Quantity
+        Stop point extracted from input :attr:`spiketrains`
+    """
+    if isinstance(spiketrains, neo.SpikeTrain):
```
toddrjen: This should probably just be:

```Python
if hasattr(spiketrains, 't_start') and hasattr(spiketrains, 't_stop'):
```

elephant/statistics.py
```diff
@@ -132,12 +135,12 @@ def mean_firing_rate(spiketrain, t_start=None, t_stop=None, axis=None):
 
     if not axis or not found_t_start:
         return np.sum((spiketrain >= t_start) & (spiketrain <= t_stop),
-                      axis=axis) / (t_stop-t_start)
```
toddrjen: Please don't put any pep8 changes for other code in the commit.  These aren't default pep8 rules anyway (are you using an old version of pep8?).

elephant/statistics.py
```diff
@@ -240,3 +243,83 @@ def lv(v):
     # calculate LV and return result
     # raise error if input is multi-dimensional
     return 3.*np.mean(np.power(np.diff(v)/(v[:-1] + v[1:]), 2))
+
+
+def psth(sts, w, t_start=None, t_stop=None, output='counts', clip=False):
```
toddrjen: This should be a separate pull request.

toddrjen:
Please split this into two separate pull requests, one for Binned and one for psth.

elephant/statistics.py
alperyeg: Hm that's good to know. Neither the pep8 command evoked from terminal nor my IDE (latest intellij with pycharm plugin or emacs) complained about this. In fact it must be auto format... I will consider this with the new PR.

elephant/conversion.py
alperyeg: That is arguable. I want to check for spiketrain objects not for attributes in this case. One could even think about leaving the condition out, to use duck typing and make it more pythonic. But still I would prefer `isinstance()` over `hasattr()`.

elephant/statistics.py
alperyeg: Alright, I can make a separate PR. But the PSTH is using the binning class. 
That means, if separated,  it will not pass the tests. If that is acceptable, than fine with me.

toddrjen: It is important to get how the binning should work figured out first.  Once it is accepted, then you can submit the PSTH pull request.

PR #16, opened 2015-02-02
added binned class and tests

alperyeg:
This pull request integrates a binning class `Binned` in the `conversion` module, that provides a representation of binned (clipped, unclipped) spike trains based on sparse matrices, which is an important feature for large sets of spike trains in combination with fine-grained binning. This binning class therefore acts more as a specialized data type within elephant. The advantage of this class is that it takes care of efficiently handling the sparse matrix management for the user (optionally keeping the “non-sparse” matrices in memory) and handles the clipping, but also keeps the time axis in order (similar to the `AnalogSignal` class of `Neo`). Please note: The existing binning function in elephant (`conversion.binarize`) has intentionally not been touched in this PR.


elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
     else:
         return res, pq.Quantity(np.arange(t_start, t_stop+sampling_period,
                                           sampling_period), units=units)
+
+###############################################################################
+#
+# Methods to calculate parameters, t_start, t_stop, bin size, number of bins
+#
+###############################################################################
+def calc_tstart(num_bins, binsize, t_stop):
```
toddrjen: These are essentially one-line code snippets, used in exactly one place each.  So I don't think they should be functions at all, and certainly not public API.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+    """
+
+    if num_bins is not None and t_start is not None and t_stop is not None:
+        if t_stop < t_start:
+            raise ValueError("t_stop (%s) is smaller than t_start (%s)"
+                             % (t_stop, t_start))
+        return (t_stop - t_start) / num_bins
+
+
+def set_start_stop_from_input(spiketrains):
```
toddrjen: This doesn't set anything, it returns it.  So I think the function name and docstring are incorrect.  However, again, this is such a simple code snippet I don't think it should be a function, and again certainly not public API.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+    spike_indices
+    mat_clipped
+    mat_unclipped
+
+    Notes
+    -----
+    There are four cases the given parameters must fulfill.
+    Each parameter must be a combination of following order or it will raise
+    a value error:
+    * t_start, num_bins, binsize
```
toddrjen: What happens if four are defined, and they don't agree?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        if binsize is None and t_start is None and t_stop is None \
+                and num_bins is None:
+            raise ValueError(
+                "No arguments given. Please enter at least three arguments")
+        # Check if num_bins is an integer (special case)
+        if num_bins is not None:
+            if type(num_bins) is not int:
+                raise TypeError("num_bins is not an integer!")
+        # Check if all parameters can be calculated, otherwise raise ValueError
+        if t_start is None:
```
toddrjen: You don't actually raise a ValueError here.

The simplest approach would probably just be

```Python
if sum(x is None for x in [t_start, t_stop, num_bins, binsize]) > 1:
  raise ValueError('blah blah')
```

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        elif t_stop is None:
+            self.t_stop = calc_tstop(num_bins, binsize, t_start)
+        elif num_bins is None:
+            self.num_bins = calc_num_bins(binsize, t_start, t_stop)
+            if self.matrix_columns is None:
+                self.matrix_columns = self.num_bins
+        elif binsize is None:
+            self.binsize = calc_binsize(num_bins, t_start, t_stop)
+
+    def __calc_start_stop(self, spiketrains):
```
toddrjen: Since this has no class-breaking side effects, I would make this (and similar functions) single underscore (`_calc_start_stop` not `__calc_start_stop`) to disable name mangling.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        """
+        Calculates start, stop from given spike trains.
+
+         The start and stop points are calculated from given spike trains, only
+         if they are not calculable from given parameter or the number of
```
toddrjen: Indentation messed up here.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        count is greater or equal to `3`.
+
+        Returns
+        -------
+        bool :
+            True, if the count is greater or equal to `3`.
+            False, otherwise.
+
+        """
+        param_count = 0
```
toddrjen: Again, easier just to do:

```Python
return sum(x is None for x in [self.t_start, self.t_stop, self.binsize, self.num_bins) < 3
```

Also, you need to check for `None` explicitly, since t_start or t_stop being zero is acceptable, but would resolve to False and thus mess up the test.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+            param_count += 1
+        if self.t_stop:
+            param_count += 1
+        if self.binsize:
+            param_count += 1
+        if self.num_bins:
+            param_count += 1
+        return False if param_count < 3 else True
+
+    def __check_consistency(self, spiketrains, binsize, num_bins, t_start,
```
toddrjen: To keep consistency with `neo`, this should be public API.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+                'by t_stop')
+        elif self.matrix_columns < 1 or self.num_bins < 1:
+            raise ValueError(
+                "Calculated matrix columns and/or num_bins are smaller "
+                "than 1: (matrix_columns: %s, num_bins: %s). "
+                "Please check your input parameter." % (
+                    self.matrix_columns, self.num_bins))
+
+    @property
+    def edges(self):
```
toddrjen: I don't understand this.  What is an "edge"?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        -------
+        bin_edges : quantities.Quantity array
+            All edges in interval [start, stop] with :attr:`num_bins` bins
+            are returned as a quantity array.
+        """
+        return np.linspace(self.t_start, self.t_stop,
+                           self.num_bins + 1, endpoint=True)
+
+    @property
+    def left_edges(self):
```
toddrjen: Isn't this just `edges[:-1]`?  Does this really need its own property?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        -------
+        bin_edges : quantities.Quantity array
+            All edges in interval [start, stop) with :attr:`num_bins` bins
+            are returned as a quantity array.
+        """
+        return np.linspace(self.t_start, self.t_stop, self.num_bins,
+                           endpoint=False)
+
+    @property
+    def right_edges(self):
```
toddrjen: Again, isn't this just `edges[1:]`?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        -------
+        matrix: scipy.sparse.csr_matrix
+            Sparse matrix, counted, unclipped version.
+
+        See also
+        --------
+        scipy.sparse.csr_matrix
+        mat_unclipped
+        """
+        if self._sparse_mat_u is not None:
```
toddrjen: If this is `None`, then the property will return `None`.  Is this what you really want?  If so, it should be documented.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        mat_clipped
+        """
+        if self._sparse_mat_c is not None:
+            return self._sparse_mat_c
+        # Return sparse Matrix without storing
+        tmp_mat = self._sparse_mat_u.copy()
+        tmp_mat[tmp_mat.nonzero()] = 1
+        return tmp_mat
+
+    def store_sparse_mat_clip(self):
```
toddrjen: You need to document in this method and the previous property how these relate.

Also, is there no method to remove it from memory?  Perhaps a method to add or remove it from memory, depending on an (optional) parameter?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+
+        See also
+        --------
+        scipy.sparse.csr_matrix
+        mat_clipped
+        """
+        if self._sparse_mat_c is not None:
+            return self._sparse_mat_c
+        # Return sparse Matrix without storing
+        tmp_mat = self._sparse_mat_u.copy()
```
toddrjen: What if there is no `_sparse_mat_u`?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        >>> st = n.SpikeTrain([0.5, 0.7, 1.2, 3.1, 4.3, 5.5, 6.7] * pq.s, t_stop=10.0 * pq.s)
+        >>> x = conv.Binned(st, num_bins=10, binsize=1 * pq.s, t_start=0 * pq.s)
+        >>> print(x.spike_indices)
+        [[0, 0, 1, 3, 4, 5, 6]]
+        >>> print(x.sparse_mat_unclip.nonzero()[1])
+        [0 1 3 4 5 6]
+
+        """
+        spike_idx = []
+        for row in self._sparse_mat_u:
```
toddrjen: Again, what if there is no `_sparse_mat_u`?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        Converts neo.core.SpikeTrain objects to a sparse matrix
+        (`scipy.sparse.csr_matrix`), which contains the binned times.
+
+        Parameters
+        ----------
+        spiketrains : neo.SpikeTrain object or list of SpikeTrain objects
+           The binned time array :attr:`spike_indices` is calculated from a
+           SpikeTrain object or from a list of SpikeTrain objects.
+
+        Examples
```
toddrjen: This example does not invoke `__convert_to_binned`.  You really don't need an example for non-public API.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        if self.mat_u is not None:
+            return self.mat_u
+        if self.store_mat_u:
+            self.mat_u = self.sparse_mat_unclip.toarray()
+            return self.mat_u
+        # Matrix on demand
+        else:
+            return self._sparse_mat_u.toarray()
+
+    def store_mat_unclipped(self):
```
toddrjen: This is public API but undocumented.

More fundamentally, you have a public attribute `store_mat_u`, which if changed does absolutely nothing.  I would think it would be better to have this be a property, where the setter is used to store or erase the matrix.  Then `mat_u` should be made private.  Then `mat_unclipped` can just check if `mat_u` is None.  That way, changing the value has immediate effects.  

The same could be done for `store_sparse_mat_clip`.

elephant/conversion.py
alperyeg: The idea was to make the snippets public so that other functions could use them, too.
I can think about making them private inside the class, but I will let them stay as methods.

elephant/conversion.py
alperyeg: Yes, this function should be a getter. 
Then again, I do not agree to make this function a non-method. Besides that this is a function, which is used a lot.  In our lab nearly everybody has this kind of function somewhere in his own script. So it would be better to have such functions in a central place then scattered around. That does not mean, that this has to stay in the `conversion` module.  I would prefer it even in `neo`.

elephant/conversion.py
alperyeg: I believe, you mean the documentation should be more precise?

elephant/conversion.py
alperyeg: That looks nice, I will add it.

elephant/conversion.py
alperyeg: No, the above is easier ;) 
Do you mean `Zero` instead of `None`? If `t_start` and `t_stop` are both zero the test later will fail. Also with `None` it will fail later on.

elephant/conversion.py
alperyeg: Yes this is redundant, I will remove it. The one below, too.

elephant/conversion.py
alperyeg: Just leaving out the if condition is enough. `_sparse_mat_u` is always calculated in `__convert_to_binned` and should not be `None`.

elephant/conversion.py
alperyeg: I will add to the docs that the clipped version is calculated from the unclipped one. 

Yes a removing method should be there, but as a separate explicit method.

elephant/conversion.py
alperyeg: You would get an error, that a `NoneType` doesn't have a copy method. But as mentioned above it shouldn't be the case.

elephant/conversion.py
alperyeg: Not iterable error :-)

elephant/conversion.py
alperyeg: I think `mat_u` should be just private. Both remove and store options should be separated.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+
+    def __init__(self, spiketrains, binsize=None, num_bins=None, t_start=None,
+                 t_stop=None, store_mat=False):
+        """
+        Defines a binned spike train class
+
+        """
+        # Converting spiketrains to a list, if spiketrains is one
+        # SpikeTrain object
+        if type(spiketrains) == neo.core.SpikeTrain:
```
apdavison: `if isinstance(spiketrains, neo.core.Spiketrain):` would generally be considered more Pythonic, I believe.

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+    """
+    if isinstance(spiketrains, neo.SpikeTrain):
+        return spiketrains.t_start, spiketrains.t_stop
+    else:
+        start = max([elem.t_start for elem in spiketrains])
+        stop = min([elem.t_stop for elem in spiketrains])
+    return start, stop
+
+
+class Binned:
```
apdavison: should inherit from object, since we're still supporting Python 2

apdavison: I think a more explicit class name is needed, e.g. "BinnedSpikeTrainArray"

elephant/conversion.py
alperyeg: True, will change.

elephant/conversion.py
alperyeg: I think classes should have shorter names. What about `BinnedST`?

elephant/conversion.py
```diff
@@ -152,3 +155,704 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+    num_bins : int
+        Number of bins of the binned spike train.
+        Default is `None`
+    t_start : quantities.Quantity
+        Time of the first bin (left extreme; included).
+        Default is `None`
+    t_stop : quantities.Quantity
+        Stopping time of the last bin (right extreme; excluded).
+        Default is `None`
+    store_mat : bool
```
apdavison: I don't really want to store mats, carpets, rugs or any other type of floor covering (-;
I think `store_matrix` would be clearer here (see also `mat_clipped`, etc.)

elephant/conversion.py
apdavison: Shorter names are better, but not at the expense of readability. How about `TimeHistogram`? This class essentially defines a sparse histogram representation, with additional attributes `t_start` and `t_stop`.

apdavison:
Some general comments on this PR. 

(1) Alper, I'm sorry if some of the comments appear like nit-picking, but I think this is a very important class, which will be widely used within Elephant, so I think it's important to get it right.

(2) Second, a comment on method names. The `Binned` class (aka `TimeHistogram` if you like my latest name proposal!) seems very much like a scipy sparse matrix with some extra methods. I therefore think the method names should match the scipy names as closely as possible, e.g. `toarray()` instead of `sparse_mat_unclip()` and `tosparse()` instead of `sparse_mat_unclip()`.

(3) The names "clipped" and "unclipped" are very unintuitive to me. Without carefully reading the code and documentation, I would assume that "clipped" applied to the time axis, i.e. you were clipping values off the start and/or end of the time histogram. I'm not yet sure what a better name would be. Also, given that obtaining the "clipped" version is as simple as `abs(scipy.sign(self.mat_unclipped))` (actually, I'm not sure the `abs()` is desirable here), are the "clipped" methods really necessary? Finally, if we decide to keep the "clipped" methods, wouldn't it be more memory efficient to return a boolean array rather than an integer array containing only 1s and 0s?

(4) Giving the option to store the real array seems like a premature optimization. Do you already have use cases where profiling shows that converting the sparse array to a dense one is a major bottleneck?

elephant/conversion.py
alperyeg: `TimeHistogram` is fine for me.

elephant/conversion.py
alperyeg: :)), yes I will change that.

alperyeg:
@apdavison 1) I can understand that. Sometimes it is a little bit painful, but I hope it assures the quality of elephant. On the other hand I also learn a lot by discussion this things. 

2) I can change the names, it is just a short refactorisation command. But I wonder if we are bound to use the numpy naming schemes, I can't remember that. Or is just convention because we use a lot of numpy functions? 

3) Maybe I am biased, but in our lab if I say clipped or unclipped matrix everybody understands ^^  If there are better suggestion that makes it more readable or intuitive I am happy to change the name. 
The `abs` function is needed not to get negative 1's.  See in the `test_conversion` the `test_binned_neg_times_list` method. 
I would like to have a "clipped" version of the matrix, although it is just a simple command. At least it is more consistent that way. 

3.5) Why should be a boolean array more efficient than an integer bit mask?  

4) I do not have a performance profiling for the conversion. To build the dense matrix it should be in O(n^2), whereas a method call is one operation (assuming it was calculated before). Then storage is of course O(n^2).   
The idea here was ,that it is possible to pass around the class (to other functions) and to have everything inside if needed. The other option would be, that the user stores locally the dense matrix in a variable and passes the variable and the class if needed.

apdavison:
> But I wonder if we are bound to use the numpy naming schemes, I can't remember that. Or is just convention because we use a lot of numpy functions?

We are not bound to do it. I was suggesting it as a policy, as I think it will make code more consistent and hence easier to understand. @toddrjen do you have any comments on this?

apdavison:
> The abs function is needed not to get negative 1's. See in the test_conversion the test_binned_neg_times_list method.

Ah, I see. This is for negative spike times.

apdavison:
> Why should be a boolean array more efficient than an integer bit mask?

The `mat_clipped` property gives an integer array containing only 1 and 0:  e.g. `[1 1 0 1 1 1 1 0 0 0]` I'm not very familiar with NumPy internals, but I would expect this to take 32 or 64 bits per element, whereas `[True True False True True True True False False False]` should take only one bit per element, no?

apdavison:
> 4) I do not have a performance profiling for the conversion. To build the dense matrix it should be in O(n^2), whereas a method call is one operation (assuming it was calculated before). Then storage is of course O(n^2).
> 
> The idea here was, that it is possible to pass around the class (to other functions) and to have everything inside if needed. The other option would be, that the user stores locally the dense matrix in a variable and passes the variable and the class if needed.

I understand this, the question is whether `n` is ever large enough to make it worthwhile to store the dense matrix rather than regenerating it every time. The major cost of storing the dense matrix is increased code complexity; I was just wondering if you'd made any performance measurements to show at what values of `n` this increased complexity becomes justified.

apdavison:
> Maybe I am biased, but in our lab if I say clipped or unclipped matrix everybody understands ^^ If there are better suggestion that makes it more readable or intuitive I am happy to change the name.

I'd like to have other comments on this. My concrete proposal would be:
```
  sparse_mat_unclip --> tosparse()
  sparse_mat_clip --> astype(bool).tosparse() or tosparse().astype(bool)
  mat_clipped --> toarray()
  mat_unclipped --> astype(bool).toarray() or toarray().astype(bool)
```

toddrjen:
It seems numpy defaults to int32, at least on 32-bit systems.  It may be
int64 on 64-bit systems, but I am on my smartphone right now and it us
32-bit so that is all I can check.

However, due to how memory is addressed in computers, 1-bit values are not
technically possible. There are some hacky ways to make something like a
1-bit bool array, but numpy does not do that.  The standard bool, which
numpy uses, is 8-bit.  That is still a factor of 4 memory savings (or maybe
8 on 64-bit systems).

elephant/conversion.py
toddrjen: When I mean private, I mean with a single underscore.  So instead of `calc_tstart`, it would be `_calc_tstart`.  This hides it by default, but anyone and everyone who wants to use it can.  But it could be changed or removed at any time.

elephant/conversion.py
toddrjen: Yes, it isn't documented, but more importantly there is also no explicit handling of this in the code as far as I can tell.

toddrjen:
`asfoo` is the standard syntax in numpy.  I think it would be better to stick with that naming scheme.

apdavison:
numpy uses a mixture of `asfoo` (e.g. `astype`) and `tobar` (e.g. `tolist`, `tostring`).

toddrjen:
Oops, you are right.

elephant/conversion.py
alperyeg: Please see `_check_consistency`:
``` python
 elif num_bins != int(
((t_stop - t_start).rescale(binsize.units) / binsize).magnitude)
```
which raises a `ValueError` if all four attributes are not in consent.

alperyeg:
>sparse_mat_unclip --> tosparse() 
  sparse_mat_clip --> astype(bool).tosparse() or tosparse().astype(bool)
  mat_clipped --> toarray()
  mat_unclipped --> astype(bool).toarray() or toarray().astype(bool)

I am not 100% happy with this naming. 
>sparse_mat_unclip --> tosparse() 

This indicates a transformation (to a different object structure), which is not true. Since `sparse_mat_clip` is a getter and a property, something like `get_sparse` or just `sparse_matrix` would be more appropriate.
>sparse_mat_clip --> astype(bool).tosparse() or tosparse().astype(bool)

Looks like method chaining. Same argument as before. `get_sparse_asBool` or `sparse_matrix_asBool` would be my proposal.  
>mat_clipped --> toarray()

Here it is not clear what is going to be an array. How about `matrix_toBoolArray` or `matrix_asBoolArray`? Likewise `matrix_toArray` or `matrix_asArray`? 

I also would prefer to write mixed case, i.e., `asBoolArray`instead of `asboolarray` <- looks weird.

apdavison:
>>sparse_mat_unclip --> tosparse()

>This indicates a transformation (to a different object structure), which is not true. 

It is true. The object is a TimeHistogram, the method returns a scipy sparse matrix; this is a transformation.

>Since sparse_mat_clip is a getter and a property, something like get_sparse or just sparse_matrix would be more appropriate.

I think it should be a method, not a property.

>>sparse_mat_clip --> astype(bool).tosparse() or tosparse().astype(bool)

>Looks like method chaining. Same argument as before. get_sparse_asBool or sparse_matrix_asBool would be my proposal.

Yes, this is method chaining. I don't see a reason for having separate methods/properties to return integer ("unclipped") and boolean ("clipped") versions when the boolean version can be obtained so trivially.

>>mat_clipped --> toarray()

>Here it is not clear what is going to be an array. How about matrix_toBoolArray or matrix_asBoolArray? Likewise matrix_toArray or matrix_asArray?

Sorry, that was a typo. It should be `mat_unclipped --> toarray()`.

>I also would prefer to write mixed case, i.e., asBoolArrayinstead of asboolarray <- looks weird.

Weird is in the eye of the beholder (-:  Mixed case is a major [PEP8 violation](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables). A more modern Python naming convention would be `as_bool_array` (the NumPy convention is a little outdated).

alperyeg:
> The object is a TimeHistogram, the method returns a scipy sparse matrix; this is a transformation.

I am still not convinced. You don't transform the class into an other structure. You request to give a sparse matrix, which is calculated inside/via the class. 
Let's say we have the class `bin` and I call the method `bin.tosparse()` I would expect the returned object to be a sparse representation(?) of the class, which is not the case. What I want to say that the sparse method is not working on the class itself and changing its structure. 

> I think it should be a method, not a property.

Sounds ok. 
>Yes, this is method chaining. I don't see a reason for having separate methods/properties to return integer ("unclipped") and boolean ("clipped") versions when the boolean version can be obtained so trivially.

Ah, I thought you meant the name should look like this, that's why I said method chaining. Although the boolean version may be trivial calculated I think it should stay for convenience and it is not harming. We have such cases like the `cv`.  
> Weird is in the eye of the beholder (-: Mixed case is a major PEP8 violation. A more modern Python naming convention would be as_bool_array (the NumPy convention is a little outdated).

Agreed.

apdavison:
OK, compromise proposal:

```
  sparse_mat_unclip --> to_sparse_array()
  sparse_mat_clip --> to_sparse_bool_array()
  mat_unclipped --> to_array()
  mat_clipped --> to_bool_array()
```

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+
+    A binned spike train represents the occurrence of spikes in a certain time
+    frame.
+    I.e., a time series like [0.5, 0.7, 1.2, 3.1, 4.3, 5.5, 6.7] is
+    represented as [0, 0, 1, 3, 4, 5, 6]. The outcome is dependent on given
+    parameter such as size of bins, number of bins, start and stop points.
+
+    A boolean matrix represents the binned spike train in a binary (True/False)
+    manner.
+    It's rows represent the number of spike trains
```
apdavison: Typo: should be "Its" without the apostrophe.

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+
+    A matrix with counted time points is calculated the same way, but its
+    columns contain the number of spikes that occurred in the spike train(s).
+    It counts the occurrence of the timing of a spike in its respective
+    spike train.
+
+    Parameters
+    ----------
+    spiketrains : List of `neo.SpikeTrain` or a `neo.SpikeTrain` object
+        Object or list of `neo.core.SpikeTrain` objects to be binned.
```
apdavison: this line and the previous one seem to be redundant

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        if self._mat_u is not None:
+            return self._mat_u
+        if self._store_mat_u:
+            self._mat_u = self.to_sparse_array().toarray()
+            return self._mat_u
+        # Matrix on demand
+        else:
+            return self._sparse_mat_u.toarray()
+
+    def store_array(self):
```
apdavison: I think this and `remove_array` are new since the first version of the PR, no? What is the argument for having them? Aren't they redundant with the `store_matrix` argument?

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        See also
+        --------
+        remove_array
+
+        """
+        if self._mat_u is None:
+            self._mat_u = self.to_sparse_array().toarray()
+            self._store_mat_u = True
+
+    def remove_array(self):
```
apdavison: Should probably be `removed_stored_array` for clarity.

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        scipy.sparse.csr_matrix
+        to_bool_array
+
+        """
+        if self._sparse_mat_c is not None:
+            return self._sparse_mat_c
+        # Return sparse Matrix as a copy
+        tmp_mat = self._sparse_mat_u.copy()
+        tmp_mat[tmp_mat.nonzero()] = 1
+        return tmp_mat
```
apdavison: Doesn't this need `astype('bool')` appended?

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+    calculated from given SpikeTrain objects (max start and min stop point).
+    Missing parameter will also be calculated automatically.
+    All parameters will be checked for consistency. A corresponding error will
+    be risen, if one of the four parameters does not match the consistency
+    requirements.
+
+    """
+
+    def __init__(self, spiketrains, binsize=None, num_bins=None, t_start=None,
+                 t_stop=None, store_matrix=False):
```
apdavison: `store_matrix` should be `store_array`, for consistency with the renamed methods.

elephant/conversion.py
alperyeg: The Numpydoc "requires" to write the type and a short description. In this case it is straightforward.  Shall I leave the first line (type info) out?

elephant/conversion.py
alperyeg: Yes it is new, but only the remove method. According to the discussion above https://github.com/NeuralEnsemble/elephant/pull/16#discussion-diff-23917521 I inserted it. Here a commit history would be benefical, as discussed yesterday ;) 

The argument `store_matrix` can be specified when instantiating the class, but is optional. If the user wants to store the matrix later on, there is still the possibility to do it with the method. With the remove method the user can explicitly remove the stored matrix.

elephant/conversion.py
alperyeg: Did you mean `remove_stored_array` ?

elephant/conversion.py
```diff
@@ -152,3 +155,679 @@ def binarize(spiketrain, sampling_rate=None, t_start=None, t_stop=None,
...
+        bin_edges : quantities.Quantity array
+            All edges in interval [start, stop] with :attr:`num_bins` bins
+            are returned as a quantity array.
+
+        """
+        return np.linspace(self.t_start, self.t_stop,
+                           self.num_bins + 1, endpoint=True)
+
+    @property
+    def center_edges(self):
```
alperyeg: Sounds good.

elephant/conversion.py
apdavison: Perhaps give the type as just "list"? or make the second line more general, e.g. "spike train(s) to be binned"

elephant/conversion.py
apdavison: Well, I agree that the history should not be squashed _before_ the PR is ready to merge.

I think this would be better done by having `store_matrix` (or rather `store_array` for consistency) be a property. If it is set to False, the stored array is deleted.

elephant/conversion.py
apdavison: sorry, yes

elephant/conversion.py
alperyeg: > I think this would be better done by having store_matrix (or rather store_array for consistency) be a property. If it is set to False, the stored array is deleted.

I think especially the remove method, should be explicit. 
How about this. I remove the `store_matrix` (or `store_array`) parameter from the constructor and instead  put a parameter in the method `to_sparse_bool_array`, which indiciates that the matrix shall stored. Same goes for method `to_sparse_array`.

apdavison: `store_array = False` seems pretty explicit to me, but ok. I like your proposal, but shouldn't the argument be added to `to_array` and `to_bool_array`, not to the "sparse" variants? The class always stores the sparse array anyway, the "store_array" argument is for the dense array, no?

alperyeg: > store_array = False seems pretty explicit to me, but ok. I like your proposal, but shouldn't the argument be added to to_array and to_bool_array, not to the "sparse" variants? The class always stores the sparse array anyway, the "store_array" argument is for the dense array, no

Yes, sorry I meant the `to_array`. 
Besides that, I also had implemented a method for storing the boolean version of the sparse matrix. I will remove that.  

Let me summarize: The class calculates internally the **sparse non boolean** matrix when instantiated. 
There exists one method to : 
* get the non boolean matrix - `to_sparse_array`
* calculate the **boolean** matrix out of the non boolean and return it - `to_sparse_bool_array`
* return a dense matrix non boolean version - `to_array` (will get a parameter to store)
* store non boolean version in memory - `store_array` (will be removed)
* remove non boolean version from memory - `remove_stored_array`
* store boolean version in memory - `store_sparse_bool_array` (will be removed)
* remove boolean version - `remove_stored_sparse_bool_array` (will be removed)   
What do you think about the last two points (will be removed)? Should I remove them, so no storing of boolean sparse matrix?

apdavison: Removing storage of the boolean sparse matrix is ok with me, as it simplifies things. If profiling later shows that calculating the boolean sparse matrix is a bottleneck, we could add this feature back in.

elephant/conversion.py
mdenker: I believe `TimeHistogram` is not a good name, firstly, because a binned spike train can also be clipped (i.e., only a 0/1). In that case it is not a histogram anymore. 

Second, it is not a histogram of time, but more importantly specifically of a spike train. Thus the name is not very telling about what the class does.

Third, `TimeHistogram` sounds more like an analysis, but in fact this class provides more of a data representation.

I would suggest `BinnedSpikeTrain` (similar to the suggestion above), and would leave out "Array" since that is redundant information (and not consistent with the *_to_array functions of the class). This name would also be more compatible with the Neo naming scheme.

alperyeg:
@mdenker any further comment on this? 
@apdavison  Michael correct me if I am wrong, I think Michael meant in later processing steps the boolean representation will be converted to `int` implicit internally or explicit by the user. So there could be overhead when converting from `bool` to `int`, because most of the time matrix operations will be calculated with integers.

apdavison:
> in later processing steps the boolean representation will be converted to int implicit internally or explicit by the user.

This is what I disagree with, and why I asked for an example.

alperyeg:
Here is a fragment of an implicit conversion example taken from the psth function, which will be PRed soon, after the binning class is available. 
```python
    if clip is True:
        bin_hist = bs.sparse_mat_clip.sum(axis=0)
    else:
        bin_hist = bs.sparse_mat_unclip.sum(axis=0)
    bin_hist = bin_hist.getA1()
```

It is still not adjusted to the actual changes, I just copied it. You can see the function [here](https://github.com/NeuralEnsemble/elephant/blob/0c0536793d8b9f5c3e49dd6497bdba96786923d8/elephant/statistics.py#L248) within the whole context.  

On the other hand, is there any example which just uses the boolean representation? At some point there is always a conversion when doing (matrix) calculation.

mdenker:
@apdavison Yes, the question is whether it always makes sense for the class to convert int to bool, when the user then converts bool back to int.

A simple example that forms the basis of many analysis approaches, as Alper indicated above, is to calculate the spike complexity per bin assuming that individual spike trains in the object are different neurons:
>>>a=np.array([[True,False],[True, True]])
>>>np.sum(a,axis=0)
    array([2, 1])
>>>np.sum(a,axis=0).dtype
    dtype('int64')

Of course, it works like this, but the question is whether it would not be beneficial to offer both return types separately (bool and int) for "clipped" versions.

apdavison:
My feeling is that the added complexity of offering two different return types is not worth it, since a bool array will give the same results as a {0,1} array. There may be a performance penalty from extra conversions, but until we have profiling data for a real-world use case, it feels like premature optimisation.

That said, I don't feel strongly about it - you should do what seems best.

On the other hand, we do really need to resolve the issue of the class name. Maybe we could have `BinnedSpikeTrain` as an alias or subclass of `TimeHistogram`, or we could drop `TimeHistogram` for now, and resurrect it later if need be. What will the return type of the `psth()` function be?

mdenker:
@apdavison I see your point about TimeHistogram being useful for things other than spike trains and that this is an argument to remove the semantic connotation from the name. If it would turn out that we wanted an object to represent specifically spike trains in the future (e.g., in order to retain annotations of the source Spike Train object), it could still be done as a subclass of the more general TimeHistogram object. 

However, if we use that name, the elephant documentation must be very clear about how to represent a binned spike train with that class, because I fear many people will not automatically make the connection that a TimeHistogram object is designed (primarily) for that purpose

apdavison:
> e.g., in order to retain annotations of the source Spike Train object

This is a good argument. OK, we seem to be converging on the "subclass" approach.

mdenker:
Well, at least that could be an idea for the future if it turns out to be useful... in general, I think a major challenge for the future will be to develop methods to work with these objects in a useful way (meaning in particular, retaining metadata info in the annotations of the Neo).

mdenker:
I see your point, I don't really not know what these bool->int conversion imply internally (where I am more concerned about memory duplication than execution speed in the case of large 0/1 arrays). I can live with keeping a to_bool_array() for now and adding a to_binary_array() (or similar) returning int at a later stage if it seems necessary or is wanted by users.

alperyeg:
> What will the return type of the psth() function be?
`AnalogSignal`. I have to adapt it to `AnalogSignalArray`, according to the latest discussions in the other PRs. 
Should I change the name of the class to `BinnedSpikeTrain` and leave everything as it is?
We should definitely profile this class later on, to see if there are performance drops if there is a conversion from bool to int and to see if there are memory problems when just using int.

mdenker:
After various discussions I had about the naming of this class, I think we cannot make a final decision yet. Given the current input to the class are currently SpikeTrains objects, and given that the class currently cannot compute a histogram across the spiketrains, I propose to use the more specific term BinnedSpikeTrain for now, but simultaneously open a new issue regarding the naming of this function as "TimeHistogram" or other, possibly generalizing the class more (subclassing), just so we remember to be discuss again a bit later and further develop the API -- which also gives us the opportunity to collect user input regarding this name. This would allow us to restrict the terminology of "histogram" for the analysis functions that actually compute PSTH-like things for the time being.

apdavison:
ok with me.

alperyeg:
@apdavison  or @mdenker can you merge this? Thanks!

PR #18, opened 2015-02-24
updated the install.sh for travis ci

alperyeg:
Since a few days the Travis build fails and it times out when trying to compile scipy on Ubuntu distributions. 
I have changed the install script and tested it successfully on one of my own branches. 
Travis uses an older version of Ubuntu, thus older packages are installed, which may be not compatible with the requirements stated in elephant. In that case scipy and other packages are upgraded per pip.


alperyeg:
I was not sure if I should create a Pull Request for this small change or commit it with my actual Pull Request regarding the Binning Class. So it decided to put it here as a separate PR.

mdenker:
The changes seem reasonable to me. I will merge this PR by tonight unless there are objections, so that travis works again.

PR #19, opened 2015-02-25
Butterworth filtering function for AnalogSignal

Junji110:
Added a Butterworth filtering function `butter()` in a newly created module `signal`

PR #20, opened 2015-02-27
added signal.welchpsd(), a power spectrum estimation function for AnalogSignal

Junji110:
This PR addes a power spectrum estimation function for AnalogSignal as signal.welchpsd().
It takes AnalogSignal (or AnalogSignalArray) as input, and returns power spectrum (or spectra) estimated by Welch's method.

### DIFF
--- a/elephant/signal.py
+++ b/elephant/signal.py
@@ -11,8 +11,9 @@ import quantities as pq
 import neo
 
 
-def welchpsd(data, num_seg=8, len_seg=None, freq_res=None, overlap=0.5,
-             **kargs):
+def welch_psd(data, num_seg=8, len_seg=None, freq_res=None, overlap=0.5,
+              fs=1.0, window='hanning', nfft=None, detrend='constant',
+              return_onesided=True, scaling='density', axis=-1):
     """
     Estimates power spectrum density (PSD) of a given AnalogSignal using
     Welch's method, which works in the following steps:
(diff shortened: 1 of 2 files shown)


elephant/signal.py
```diff
@@ -0,0 +1,133 @@
...
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+
+import numpy as np
+import scipy.signal
+import quantities as pq
+import neo
+
+
+def welchpsd(data, num_seg=8, len_seg=None, freq_res=None, overlap=0.5,
```
apdavison: Perhaps add an underscore for easier comprehension, i.e. `welch_psd`?

elephant/signal.py
```diff
@@ -0,0 +1,133 @@
...
+             **kargs):
```
apdavison: Why use `**kargs` rather than an explicit list of keyword arguments?

elephant/signal.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # attach proper units to return values
+    if isinstance(data, pq.quantity.Quantity):
+        if 'scaling' in params and params['scaling'] is 'spectrum':
+            psd = psd * data.units * data.units
+        else:
+            psd = psd * data.units * data.units / pq.Hz
+        freqs = freqs * pq.Hz
+
+    return psd, freqs
```
apdavison: Should return `freqs, psd` for compatibility with `scipy.signal.welch`.

elephant/signal.py
```diff
@@ -0,0 +1,133 @@
...
+        if key in ('fs', 'nperseg', 'detrend', 'window', 'noverlap', 'nfft',
+                   'return_onesided', 'scaling'):
+            params[key] = val
+        else:
+            raise ValueError("Unsupported keyword argument '{}'".format(key))
+
+    # specify sampling frequency if the data is given as AnalogSignal(Array)
+    if isinstance(data, neo.AnalogSignal) or \
+            isinstance(data, neo.AnalogSignalArray):
+        params['fs'] = data.sampling_rate.rescale('Hz').magnitude
```
apdavison: `else:` ?

apdavison:
Note that `AnalogSignal` will be removed in Neo 0.4. This will happen soon, so it would be best to support only `AnalogSignalArray` here to avoid having to modify it later.

elephant/signal.py
Junji110: Parameter 'fs' can be given as a keyword argument (ll.76-78), and it is used only when the input is not AnalogSignal or AnalogSignalArray.
If 'fs' is not given, `scipy.signal.welch()` will use `fs=1.0` as its default.

elephant/signal.py
Junji110: I meant `**kargs` for keyword arguments to be just passed to `scipy.signal.welch()`.

elephant/signal.py
apdavison: My point was that according to the docstring, `data` _must_ be either an AnalogSignal or AnalogSignal array, so the "else" should be to raise a TypeError, or you just remove the "if" entirely. If `data` can be something else (e.g. a plain NumPy array) this should be stated in the docstring, and a test case added.

elephant/signal.py
apdavison: I think it would be better to have them as explicit keyword arguments, with documentation.

elephant/signal.py
Junji110: OK, I see the point.
Which option would be preferable, either accept NumPy array, list, etc. or raise an exception?

apdavison: If it is straightforward to accept a plain array, it is good to do so. (I guess if `data` is an `AnalogSignalArray` then `freq_res` should be a Quantity, and if `data` is a plain array then `freq_res` should be a float.)

Junji110:
I chose to accept plain array as input, and made fixes accordingly.
There are other changes according to the comments.
Please review the code again.

elephant/signal.py
```diff
@@ -0,0 +1,147 @@
...
+        the interval between adjacent frequency bins. When given as a float, it
+        is taken as frequency in Hz. Default is None (determined from other
+        parameters).
+    overlap: float, optional
+        Overlap between segments represented as a float number between 0 (no
+        overlap) and 1 (complete overlap). Default is 0.5 (half-overlapped).
+    fs: Quantity array or float, optional
+        Specifies the sampling frequency of the input time series. When the
+        input is given as an AnalogSignalArray, the sampling frequency is taken
+        from its attribute and this parameter is ignored. Default is 1.0.
```
apdavison: This is something to fix later, but it is dangerous to ignore arguments. If `fs` is specified explicitly and is different from the sampling frequency of the signal, the code should raise an Exception.

apdavison:
I only realized this after merging, but once #19 is merged we will have a `signal` module and a `signal_processing` module. Is this what you want, or did you forget to change the module name in this PR?

Junji110:
I forgot to correct the module name, sorry.
In fact, I and Michael once talked about this and thought about placing the PSD function in a separate `spectral` module rather than in `signal_processing` module, and we haven't made a decision.
What do you think about this?

apdavison:
I like the idea of `spectral`. Can you make a new PR for that, and then update #19 accordingly (currently it cannot be merged because of conflicts in `test_signal.py`)?

mdenker:
+1 for spectral for now.

In the future, maybe we need to become more specific in the name, if there is too much spectral analysis coming in (e.g., for AnalogSignal data and SpikeTrain Data,...), but for now I think that is good.

PR #22, opened 2015-03-04
Added function to z-score an analog signal.

mdenker:
This PR supplies a function to calculate the z-score of AnalogSignal or AnalogSignalArray objects, or lists thereof (in which case the normalization is performed by taking the mean and std across all list elements, such that each element is normalized equally).

The doc string contains a "Use Case" section which I would like to put up for discussion as an optional section that allows to provide more hands on information or suggestions on how a particular function could be useful in an imaginary (but realistic) analysis situation.


mdenker:
Broken for Python 3.4, will fix...

tbekolay:
Scipy contains a function for this: [`scipy.stats.zscore`](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.mstats.zscore.html). How does this differ from Scipy's implementation?

mdenker:
Indeed, the first thought was to provide only an alias to the scipy zscore function in elephant.

A common use case is that the zscore should be calculated for multiple AnalogSignals, but using the total mean and variance, e.g., across multiple trials or sometimes even electrodes. We thought to include this as an option, as this cannot be performed using the scipy zscore and requires users to code from scratch. Here, a list of signals can be passed, in which case a list of z-transformed AnalogSignals is output, but each transformed against the cumulative mean and stddev.

A second aspect of this function is the "inplace" option that allows to overwrite input signals, which is very handy to save memory if dealing with large data sets.

scipy zscore is used in some of the unit tests to double check.

An alternate option could be to alias the name zscore to the scipy function, write unit tests to ensure scipy zscore keeps working with AnalogSignal objects, and rename the function of this PR to a more specific name. However, I am not quite sure this is useful 

(Note that the current implementation can still be drastically improved in terms of memory consumption when dealing with lists, but that's for another iteration).

elephant/signal_proc.py
```diff
@@ -0,0 +1,130 @@
+'''
+signal_proc module
```
apdavison: I think we should avoid using abbreviations* in module or public function names, so I would make this `signal_processing`.

*widely-used acronyms (e.g. fft, sta) I think are ok, however.

apdavison:
This is not compatible with Neo 0.4, in which `AnalogSignal` will be removed.

elephant/signal_proc.py
Junji110: I added a Butterworth filtering function in "signal" module, which was meant as a module of signal processing functions.
We need to agree on the naming of this module and use it consistently.

mdenker: Yes, true. I am not a fried of long names, but I think it will be good to have explicit module names for a start to avoid confusion, so I would agree with signal_processing.

mdenker:
Ok, wasn't sure if to go 0.4 right away, but that would simplify things in the doc string, I'll adapt to stick to AnalogSignalArray (I still find the name quite cluncky to work with, I'll raise this on neo).

elephant/signal_proc.py
Junji110: "signal_processing" is fine with me as well.
I will fix the module name in my pull request for the butterworth function accordingly.

elephant/signal_processing.py
```diff
@@ -1,17 +1,136 @@
...
+    You may supply a list of AnalogSignalArray objects, where each object in
+    the list contains the data of one trial of the experiment, and each signal
+    of the AnalogSignalArray corresponds to the recordings from one specific
+    electrode in a particular trial. In this scenario, you will z-transform the
+    signal of each electrode separately, but transform all trials of a given
+    electrode in the same way.
+
+    Example
+    -------
+    >>> a=neo.AnalogSignalArray(
```
apdavison: This is not a valid way to create an AnalogSignalArray (Neo should complain, but doesn't) - it should always be 2D.

mdenker: Ok, didn't know that, will change that for the documentation. But I wonder if this is something that should be considered again in neo for the future -- it seems a bit clumsy to me given that AnalogSignal is dropped and given that 1-d AnalogSignals are a common thing to have.

apdavison: Yes, I agree - we should support this properly in Neo.

PR #23, opened 2015-03-06
Added spike-triggered average function

dholstein:


elephant/sta.py
```diff
@@ -0,0 +1,144 @@
...
+    Calculates the spike-triggered average of analog signals in a given time window around 
+    the spike times of a corresponding spiketrain for multiple signals each.
+
+    The function receives n analog signals and either one or n spiketrains. In case it is one spiketrain 
+    this one is muliplied n-fold and used for each of the n analog signals.
+
+    Parameters
+    ----------
+
+        signal: neo AnalogSignalArray object containing n analog signals
```
Junji110: This description style of parameters is not following NumPy documentation guideline.
I would suggest to revise it to comply with the guideline.

elephant/sta.py
```diff
@@ -0,0 +1,144 @@
...
+                "The spiketrain indexed by %i stops later than the analog signal." % i)
+
+    # *** Main algorithm: ***
+
+    # window_bins: number of bins of the chosen averaging interval
+    window_bins = int(np.round(
+        ((window_stoptime - window_starttime) * signal.sampling_rate).simplified))
+    # result_sta: array containing finally the spike-triggered averaged signal
+    result_sta = AnalogSignalArray(np.zeros(
+        (window_bins, num_signals)), sampling_rate=signal.sampling_rate, units=signal.units)
```
Junji110: I guess this row-column order is opposite from what users expect, in the sense that result_sta[0] refers to the first bins of multiple signals, rather than all bins of the first signal.

apdavison: I agree it is opposite to my expectations, but this is a Neo convention (originally introduced at the request of people working with EEG data), so we cannot change it within Elephant. There has been some discussion within the Neo community about reversing the convention, but obviously this would be a major backwards-incompatible change.

Junji110: That confuses me, because the following code obviously deals the first axis as signal identity and the second as time index...

>>> import numpy as np
>>> import quantities as pq
>>> import neo
>>> asa = neo.AnalogSignalArray(np.zeros((2,1000))*pq.mV, sampling_rate=1.0*pq.Hz)
>>> asa[0]
AnalogSignalArray in 1.0 mV with 1000 float64 values
channel index: None
sampling rate: 1.0 Hz
time: 0.0 s to 1000.0 s

apdavison: You've found a bug in Neo. `asa[0]` should I think give a Quantity array, not an `AnalogSignalArray`.

Note that indexing on the second dimension gives an `AnalogSignal`, as expected.
```
In[6]: asa[:, 0]
Out[6]:
AnalogSignal in 1.0 mV with 2 float64 values
channel index: None
sampling rate: 1.0 Hz
time: 0.0 s to 2.0 s
```

apdavison: Also, you can see by the `t_stop` attribute that the number of samples in `asa` is 2 (given a sampling rate of 1 Hz):
```
In [7]: asa.t_stop
Out[7]: array(2.0) * s
```

Junji110: I see, thanks for clarification.
I need to fix my functions according to this...

apdavison:
@alperyeg @mdenker ok to merge this?

elephant/sta.py
```diff
@@ -0,0 +1,146 @@
...
+                unused_spikes[i] += 1
+
+        # normalization
+        result_sta[:, i] = result_sta[:, i] / used_spikes[i]
+
+        total_used_spikes += used_spikes[i]
+
+    if total_used_spikes == 0:
+        raise ValueError(
+            "No spike at all was either found or used for averaging")
```
Junji110: I personally don't think that this is the case where an error should be raised.
What's the argument against just returning zeros as STA in this case (as the function does in the case where one of given multiple spike trains happens to have no spike to be used for the averaging)?

apdavison: Why zeros? Suppose the signal lies entirely between -67 mV and -58 mV. Returning 0 mV is surely not the right approach - it could really screw up any further analysis based on averaging.

You are correct though, that we should be consistent, so perhaps we should raise an Exception in the multiple spike train situation as well. Alternatively, we could return an array containing NaN.

Junji110: I agree with returning NaN consistently.
I'm concerned about a case where a user applies this function to a series of datasets one after another, and the whole analysis is aborted just because one of the datasets happens to have a unit which hardly have any spikes.

alperyeg: `NaN` is a little bit strange, since the problem is that there are no spikes found and not a calculation problem, like dividing through zero. One could also think about an array of `None`.
But in general I can live with that, but maybe you should include a warning message, indicating that no spikes were found.

Junji110: Here is an argument in support of NaN: we can consider this case as zero-divided-by-zero, which is defined as NaN in IEEE standard for floating-point arithmetric (note that Python and Numpy don't follow this standard; they raise an error).

Junji110: ...and you cannot create an AnalogSignalArray from an array of None

alperyeg: >...and you cannot create an AnalogSignalArray from an array of None
True, didn’t think about that. Then I also support `NaN`.

apdavison: Python raises a ZeroDivisionError, but Numpy gives an array of NaN:

```python
>>> np.zeros(5,)/0.0
array([ nan, nan, nan, nan, nan])
```

mdenker:
The documentation does not mention the annotations used_spikes and unused_spikes placed on the resulting AnalogSignalArray object. They should be explained with the "Result".

From the doc it is not clear, if in the case of no spikes, nan is returned or an AnalogSignalArray containing a nan value.

mdenker:
There are still a few lines >79 chars.

mdenker:
There are error messages that mention window_starttime and window_stoptime, however these are internal variables that are not telling to the user. Rather refer to window[0] and window[1].

mdenker:
In general, we need to define in elephant a definition of how we interpret an analog signal (does the time stamp indicate the beginning of the sample bin, or its center?) and how we define windows, i.e., should we define both borders as inclusive (t>=start and t<=stop) or the right border as exclusive (t>=start and t<stop) to prevent spikes being counted double, for example.

In this code, I believe the choice is that analog signal time stamps are the "center" of where the signal is recorded. I infer this from line 126, where `np.round` is used instead of `np.floor`. In this cases consider a spike at spiketime=6.3, a window_stop=2.1, and t_stop of the analog signal at 8.0. 

This means that the analog signal is well-defined up to 8.49. However, the test line 124
`spiketime + window_stoptime <= signal.t_stop`
yields False and the spike is ignored (6.3+2.1=8.4 is not <= 8), although it should be considered.

So, in summary, independent of whether the above issue is indeed a problem, I think we need to first clearly define a rule on how these issues are interpreted in elephant, before we can fix this (and possibly other) functions. I will try to include such a treatment in the documentation to put up for discussion

mdenker:
The doc string of `window` is a bit confusion. I would like to suggest
`window : pair (2-tuple) of Time Quantity objects.`
--> 
`window : tuple of 2 Quantity objects with dimensions of time`
or maybe we can even come up with an elephant-wide convention
`window : tuple of 2 Quantity objects (dimensions: time)`
since this will often surface.

mdenker:
In line 137, the STA is normalized (i.e., averaged) by dividing by the number of spikes
`result_sta[:, i] = result_sta[:, i] / used_spikes[i]`
Since the future division module is not loaded in this module, I am afraid this line could go wrong (in terms of integer division) of the signal is of dtype int. Could you please check?

apdavison:
> In general, we need to define in elephant a definition of how we interpret an analog signal (does the time stamp indicate the beginning of the sample bin, or its center?) 

The time stamp represents the precise time at which the sample was taken. Neo takes this as the beginning of the sample bin (see the calculation of t_stop in AnalogSignal).

> and how we define windows, i.e., should we define both borders as inclusive (t>=start and t<=stop) or the right border as exclusive (t>=start and t<stop) 

I suggest exclusive, as this the usual case for Python.

> In this code, I believe the choice is that analog signal time stamps are the "center" of where the signal is recorded. I infer this from line 126, where np.round is used instead of np.floor. In this case consider a spike at spiketime=6.3, a window_stop=2.1, and t_stop of the analog signal at 8.0.

No, the time values are the beginning of the bin. Possibly we should use np.floor in line 126.

> This means that the analog signal is well-defined up to 8.49. 

It is defined only up to t_stop (8.0)

> However, the test line 124
> spiketime + window_stoptime <= signal.t_stop
> yields False and the spike is ignored (6.3+2.1=8.4 is not <= 8), although it should be considered.

No, it should not be considered, as the window extends outside the signal.

> So, in summary, independent of whether the above issue is indeed a problem, I think we need to first clearly define a rule on how these issues are interpreted in elephant, before we can fix this (and possibly other) functions. I will try to include such a treatment in the documentation to put up for discussion.

I think this is already determined by Neo, although it wouldn't hurt to make it clear in the documentation.

apdavison:
> In line 137, the STA is normalized (i.e., averaged) by dividing by the number of spikes
> result_sta[:, i] = result_sta[:, i] / used_spikes[i]
> Since the future division module is not loaded in this module, I am afraid this line could go wrong (in terms of integer division) of the signal is of dtype int. Could you please check?

This is a good point - we should have `from __future__ import division` in all Elephant modules. However, should we really allow `AnalogSignal` with dtype int, since `int` is not really analog...

mdenker:
Ok, great. I thought it was the center, since also the very first binarize function had returned bin centers. 
I very much more like the definition of it being the beginning of the sample bin, but then it should be `np.floor`, and exclusive right border is almost a must.

I think we need to have this in the documentation for sure, these are super important issues for users to understand -- also beyond neo in terms of how windows are defined in general in elephant (i.e., that a tuple `(start,stop)` really means [start,stop)).

mdenker:
Well, float is also not analog for that matter -- in general I saw that the name AnalogSignal causes some confusion with people and I think TimeSeries would be more fitting -- but that's a different topic.

But coming back to the question, yes, we must allow int, almost any recording device outputs integers, and often you want to keep that exact precision for as long as possible.

apdavison:
> in general I saw that the name AnalogSignal causes some confusion with people and I think TimeSeries would be more fitting

Yes, the old problem of naming things again. We had a long discussion of this when Neo was first being designed. The problem with "time series" is that in a large part of the Python data analysis world, this is assumed to mean data with calendar dates. Simply "Signal" might be the best choice, in the end.

mdenker:
Nonono, I didn't want to start a discussion :-) But I do think that integer data type should be valid for AnalogSignalArrays (at least unless we have something else to represent integer time series)...

dholstein:
> There are error messages that mention window_starttime and window_stoptime, however these are internal variables that are not telling to the user. Rather refer to window[0] and window[1].
> 
If I understand right, it is asked for letting appear in the error messages text like 'window[0]' or 'window[1]'.
I think this is less clear to the user than the text until now given in the error messages. 
In order to replace the internal variable names in the error messages  I suggest to replace 
e.g. 'window_starttime'  by text like 'starttime of the window' or 'starttime of the averaging interval' 
and keep variable names as they are.

mdenker:
Or another alternative: "start time of the window (window[0])" and "stop time of the window (window[1])"? I am fine with either, as long as it is understandable without looking inside the function.

dholstein:
Discussion of 25.03.15 on the files fully considered. 
Remark: As a consequence of the change from round to floor in line 152 
a further change from round to ceil in line 132 was introduced.

PR #24, opened 2015-03-09
Surrogates

carloscanova:
This PR adds the module `surrogates.py` and its respective unit tests. This module can be used to generate surrogate spike trains from a given `neo.SpikeTrain`. It includes the following surrogate types: 
* train shifting: randomly shifts the whole spike train by a given amount
* spike dithering: uniformly dithers spike times around their original position
* spike jittering: jitters the spike times within their bins
* spike time randomization: randomly redistributes the spike times
* isi shuffling: shuffles the inter spike intervals

### DIFF
--- a/elephant/surrogates.py
+++ b/elephant/surrogates.py
@@ -1,0 +1,490 @@
+# -*- coding: utf-8 -*-
+"""
+docstring goes here.
+
+:copyright: Copyright 2014 by the Elephant team, see AUTHORS.txt.
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+
+import numpy as np
+import quantities as pq
+import neo
+import elephant.statistics as es
+
+
+def spike_dithering(spiketrain, dither, n=1, decimals=None, edges=True):
+    """
+    Generates surrogates of a spike train by spike dithering.
+
+    The surrogates are obtained by uniformly dithering times around the
+    original position. The dithering is performed independently for each
+    surrogate.
+
+    The surrogates retain the :attr:`t_start` and :attr:`t_stop` of the
+    original `SpikeTrain` object. Spikes moved beyond this range are lost or
+    moved to the range's ends, depending on the parameter edge.
+
+
+    Parameters
+    ----------
+    spiketrain :  neo.SpikeTrain
+        The spike train from which to generate the surrogates
+    dither : quantities.Quantity
+        Amount of dithering. A spike at time t is placed randomly within
+        ]t-dither, t+dither[.
+    n : int (optional)
+        Number of surrogates to be generated.
+        Default: 1
+    decimals : int or None (optional)
+        Number of decimal points for every spike time in the surrogates
+        If None, machine precision is used.
+        Default: None
+    edges : bool (optional)
+        For surrogate spikes falling outside the range
+        `[spiketrain.t_start, spiketrain.t_stop)`, whether to drop them out
+        (for edges = True) or set that to the range's closest end
+         (for edges = False).
+        Default: True
+
+    Returns
+    -------
+    list of neo.SpikeTrain
+      A list of `neo.SpikeTrain`, each obtained from :attr:`spiketrain` by
+      randomly dithering its spikes. The range of the surrogate spike trains
+      is the same as :attr:`spiketrain`.
+
+    Example
+    -------
+    >>> import quantities as pq
+    >>> import neo
+    >>>
+    >>> st = neo.SpikeTrain([100, 250, 600, 800]*pq.ms, t_stop=1*pq.s)
+    >>> print spike_dithering(st, dither = 20*pq.ms)
+    [<SpikeTrain(array([  96.53801903,  248.57047376,  601.48865767,
+     815.67209811]) * ms, [0.0 ms, 1000.0 ms])>]
+    >>> print spike_dithering(st, dither = 20*pq.ms, n=2)
+    [<SpikeTrain(array([ 104.24942044,  246.0317873 ,  584.55938657,
+        818.84446913]) * ms, [0.0 ms, 1000.0 ms])>,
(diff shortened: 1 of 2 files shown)


elephant/surrogates.py
```diff
@@ -0,0 +1,490 @@
+# -*- coding: utf-8 -*-
+"""
+docstring goes here.
```
apdavison: Please write a docstring summarizing what is contained in the module

elephant/surrogates.py
```diff
@@ -0,0 +1,490 @@
...
+
+:copyright: Copyright 2014 by the Elephant team, see AUTHORS.txt.
```
apdavison: 2015

apdavison:
I think function names would be better as active verbs. e.g. `dither_spikes`, `shuffle_isis`. Given that the context is clear from the module, I also think shorter names like `dither`, `shuffle`, `jitter` are worth considering. What do others think?

elephant/surrogates.py
```diff
@@ -0,0 +1,490 @@
...
+        Amount of dithering. A spike at time t is placed randomly within
+        ]t-dither, t+dither[.
+    n : int (optional)
+        Number of surrogates to be generated.
+        Default: 1
+    decimals : int or None (optional)
+        Number of decimal points for every spike time in the surrogates
+        If None, machine precision is used.
+        Default: None
+    edges : bool (optional)
```
apdavison: I find "edges" to be a very unintuitive name for this argument.

elephant/surrogates.py
```diff
@@ -0,0 +1,490 @@
...
+    # Main: generate the surrogates
+    surr = data.reshape((1, len(data))) + 2 * dither * np.random.random_sample(
+        (n, len(data))) - dither
+
+    # Round the surrogate data to decimal position, if requested
+    if decimals is not None:
+        surr = surr.round(decimals)
+
+    if edges is False:
+        # Move all spikes outside [spiketrain.t_start, spiketrain.t_stop] to
```
apdavison: Is this ever a good idea? What about the option of allowing the surrogates to have different `t_start` and `t_stop` to the original spike train?

elephant/surrogates.py
```diff
@@ -0,0 +1,490 @@
...
+    ----------
+    spiketrain :  neo.SpikeTrain
+        The spike train from which to generate the surrogates
+    dither : quantities.Quantity
+        Amount of dithering. A spike at time t is placed randomly within
+        ]t-dither, t+dither[.
+    n : int (optional)
+        Number of surrogates to be generated.
+        Default: 1
+    decimals : int or None (optional)
```
apdavison: What is the use case for limiting the number of decimal places?

elephant/test/test_surrogates.py
```diff
@@ -0,0 +1,308 @@
...
+
+        nr_surr = 2
+        dither = 10 * pq.ms
+        surrs = surr.spike_dithering(st, dither=dither, n=nr_surr)
+
+        self.assertEqual(type(surrs), list)
+        self.assertEqual(len(surrs), nr_surr)
+
+        for surrog in surrs:
+            self.assertEqual(type(surrs[0]), neo.SpikeTrain)
```
apdavison: can use `self.assertIsInstance(surrs[0], neo.SpikeTrain)` to perform the same test but with a more informative error message if it fails.

elephant/surrogates.py
apdavison: In particular, I think it would be helpful to explain in what circumstances one might want to use surrogates, preferably with a reference to a review paper, book chapter or encyclopaedia article which gives more depth.

elephant/surrogates.py
etorre: When the input spiketrain was created at limited sampling precision (e.g. 1 ms with NEST, or 1/30 - 1/10 ms with most recording systems, one might need to preserve such precision.

elephant/surrogates.py
etorre: It might be done if one then needs to bin the surrogate spiketrain into a 0-1 sequence, or wants to preserve the total spike count. I am personally fine with dropping the option though - never used it.
different t_start and t_stop are not fine instead, because of the meaning of surrogate data - stochastic spike times inside a well-defined and fixed time window [t_start,t_stop].

elephant/surrogates.py
etorre: A question here: how do we handle authorship and acknowledgments? How to know who contributed what?

elephant/surrogates.py
etorre: You are right that it's confusing, especially if one has the NEO (or numpy.hostogram) vocabulary in mind. Any suggestions? Would "borders" or "push_to_borders" fit?

elephant/surrogates.py
apdavison: This is still to be decided. What I suggest is:

1) Firstly, everyone who contributes (above some threshold of contributions, so that just fixing a single typo doesn't count) puts their name and affiliation (if any) in AUTHOR.txt.
2) Secondly, when you contribute a module, class or function, you can put your name in the docstring if you wish. If someone subsequently modifies the code for that function, they can add their name, e.g. "Original implementation by X, performance improvements by Y."

I think it is simplest to keep the phrase "The Elephant team" in the copyright statement however; the actual copyright is probably held jointly by the institutions to which the contributors belong (FZJ, CNRS, etc.) although this depends on your contract (e.g. students in some countries get to keep the copyright).

apdavison:
Replaced by #27

etorre:
Hey all. Since I worked further on the comments but I wasn't the creator of this pull request, I made a new pull request "Surrogate spike trains". Please let's continue our discussion there. 
I considered all comments above and either applied the suggested changes or answered here why not.
Can any of the admin or Carlos close this pull request? Thanks!

PR #26, opened 2015-03-16
Time Histogram

alperyeg:
This PR adds the function `time_histogram` to the `statistics` module.

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -527,11 +527,10 @@ def instantaneous_rate(spiketrain, sampling_period, form, sigma, m_idx=None,
     return rate
 
 
-def psth(spiketrains, binsize, t_start=None, t_stop=None, output='counts',
-         clip=False):
+def time_histogram(spiketrains, binsize, t_start=None, t_stop=None,
+                   output='counts', binary=False):
     """
-    Peristimulus Time Histogram (PSTH) of a list of :attr:`neo.SpikeTrain`
-    objects.
+    Time Histogram of a list of :attr:`neo.SpikeTrain` objects.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


elephant/statistics.py
```diff
@@ -524,3 +525,101 @@ def instantaneous_rate(spiketrain, sampling_period, form, sigma, m_idx=None,
                                  units=pq.Hz, t_start=t_start)
 
     return rate
+
+
+def psth(spiketrains, binsize, t_start=None, t_stop=None, output='counts',
+         clip=False):
```
apdavison: In the discussion of the binned spike train class, we agreed to drop the name "clip", as it is confusing/ambiguous for some people, and use a name involving "bool" or "binary" instead.

elephant/statistics.py
```diff
@@ -524,3 +525,101 @@ def instantaneous_rate(spiketrain, sampling_period, form, sigma, m_idx=None,
...
+    """
+    Peristimulus Time Histogram (PSTH) of a list of :attr:`neo.SpikeTrain`
```
apdavison: The word "stimulus" appears only in the title, and nowhere else in the docstring or the list of arguments. For greater clarity, I wonder if we shouldn't rename "t_start" to something like "onset", "stimulus_onset", "event_time", etc., and replace "t_stop" with "duration". I also think there is a case for making "t_start/onset/whatever" a required argument.

elephant/statistics.py
alperyeg: I think this is a different context here. In binned spike train class the parameter `as_bool` was used to define the output. Here it is used internally, to clip a list of spiketrains to a binary representation but the output is not binary.  In this case the output is a sum/number of spiking neurons represented as a Time Histogram.
Maybe `clipping` (as a process) is more clear?  Or some other term?

elephant/statistics.py
alperyeg: Or we change the function name to `time_histogram` and have an alias `psth`. 
Why do you think `t_start` should be required? In the worst case it will be extracted from the maximal `t_start` of the spike train objects. But I am also fine with making it a requirement.

elephant/statistics.py
apdavison: Another term. To me, "clipping" suggests you are changing the time limits (i.e. t_start and t_stop).

elephant/statistics.py
apdavison: No, I think we should stick with "psth" because it is well known. "Peristimulus" implies that there is a stimulus, and that the histogram is relative to the onset time of that stimulus. This is why I think we should use a name different from `t_start` for the stimulus onset time (to make it clear this is different from `t_start` of the spike trains), and why the argument should be required.

paulchorley: I think the issue here is that 'PSTH' is a specific use case for this function.  In many instances users (at least we here in Juelich) may not have a stimulus, but would still want to compute the time histogram over a set of SpikeTrain objects.  

For example, when there are, say, a hundred of so parallel recordings from an elecrode array and we wish to compute a so-called 'Population Time Histogram'.  We can use the 'psth' function for this, but then the name is not appropriate.  

My feeling is that the function should be made to be generic.

Therefore, I would like the function to be called time_histogram and to take t_start and t_stop as parameters, to allow the result to be computed only over simultaneous subsections of each input.

elephant/statistics.py
paulchorley: I think that we should move towards recognising that this is really a delta function representation, either Dirac or Kronecker depending on ... well you know this.  So maybe we could call the option 'delta', or 'delta_clip'?  Perhaps we could also use the same terminology in the binned spike train class?

PR #28, opened 2015-03-26
Correlation coefficient

mdenker:
This pull request provides a new module on spike train correlations. It provides a function to calculate the matrix of correlation coefficients for a list of BinnedSpikeTrain objects (using its internal sparse matrix representation, thus avoiding the need to construct large binned matrices).

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -51,6 +51,21 @@ def corrcoef(binned_sts, binary=False):
         binned_sts[i] and binned_sts[j]. If binned_sts contains only one
         SpikeTrain, C=1.0.
 
+    Examples
+    --------
+    Generate two Poisson spike trains
+    >>> st1=elephant.spike_train_generation.homogeneous_poisson_process(
+            rate=10.*Hz, t_start=0.*s, t_stop=10.*s)
+    >>> st2=elephant.spike_train_generation.homogeneous_poisson_process(
+            rate=10.*Hz, t_start=0.*s, t_stop=10.*s)
+
+    Calculate the correlation matrix.
+    >>> cc_matrix=elephant.spike_train_correlation.corrcoef(
+        elephant.conversion.BinnedSpikeTrain([st1,st2], binsize=5*ms))
+
+    The correlation coefficient between the spike trains is stored in
+    cc_matrix[0,1] (or cc_matrix[1,0])
+
     Notes
     -----
     * The spike trains in the binned structure are assumed to all cover the
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -4,7 +4,7 @@ Spike train correlation
 
 This modules provides functions to calculate correlations between spike trains.
 
-:copyright: Copyright 2014 by the Elephant team, see AUTHORS.txt.
+:copyright: Copyright 2015 by the Elephant team, see AUTHORS.txt.
 :license: Modified BSD, see LICENSE.txt for details.
 """
 from __future__ import division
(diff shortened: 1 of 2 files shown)


elephant/spike_train_correlation.py
```diff
@@ -0,0 +1,131 @@
...
+    C = np.zeros((num_neurons, num_neurons))
+
+    # Retrieve unclipped matrix
+    spmat = binned_sts.to_sparse_array()
+
+    # For each row, extract the nonzero column indices and the corresponding
+    # data in the matrix (for performance reasons)
+    bin_idx_unique = []
+    bin_counts_unique = []
+    if binary:
```
alperyeg: I was wondering if it would be more cleaner in terms of code complexity to split this function internally into 2 private functions which are called depending on the parameter `binary`.

elephant/spike_train_correlation.py
```diff
@@ -0,0 +1,131 @@
...
+    else:
+        for s in spmat:
+            bin_counts_unique.append(s.data)
+
+    # All combinations of spike trains
+    for i in range(num_neurons):
+        for j in range(i, num_neurons):
+            # Number of spikes in i and j
+            if binary:
+                n_i = len(bin_idx_unique[i])
```
alperyeg: Wouldn't it more efficient to convert the list into a numpy array before and use the array(s) for indexing?

elephant/spike_train_correlation.py
mdenker: Not sure.. I see your point, but I thought it's still in limits to not split, and maybe its easier to maintain to have the equivalent steps for binary=True/False always side-by-side.

elephant/spike_train_correlation.py
mdenker: Good point, I will investigate.

elephant/spike_train_correlation.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+"""
+Spike train correlation
+
+This modules provides functions to calculate correlations between spike trains.
+
+:copyright: Copyright 2014 by the Elephant team, see AUTHORS.txt.
```
apdavison: 2015

elephant/spike_train_correlation.py
```diff
@@ -0,0 +1,131 @@
...
+    Each entry in the matrix is a real number ranging between -1 (perfectly
+    anti-correlated spike trains) and +1 (perfectly correlated spike trains).
+
+    If binary is True, the binned spike trains are clipped before computing the
+    correlation coefficients, so that the binned vectors b_i, b_j are binary.
+
+    Parameters
+    ----------
+    binned_sts : elephant.conversion.BinnedSpikeTrain
+        A binned spike train class containing the spike trains to be evaluated.
```
apdavison: "instance" or "object" or simply "binned spike train", not "class"

elephant/spike_train_correlation.py
```diff
@@ -0,0 +1,131 @@
...
+    binary : bool, optional
+        If True, two spikes of a particular spike train falling in the same
+        bin are counted as 1, resulting in binary binned vectors b_i. If False,
+        the binned vectors $b_i$ contain the actually spike counts per bin.
```
apdavison: remove "actually"

elephant/spike_train_correlation.py
```diff
@@ -0,0 +1,131 @@
...
+This modules provides functions to calculate correlations between spike trains.
+
+:copyright: Copyright 2014 by the Elephant team, see AUTHORS.txt.
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+from __future__ import division
+import numpy as np
+
+
+def corrcoef(binned_sts, binary=False):
```
apdavison: What is the reason for providing a binned spike train object, rather than a list of `neo.SpikeTrain`s?

mdenker:
The reason for passing the BinnedSpikeTrain object was that this gives full control over the binning parameters (binsize, tstart, tstop, duration) without having to duplicate these parameters for the corrcoeff function, which seemed additional clutter. This is debatable, but I think that this is a cleaner and simpler solution, that makes it transparent to the user what is happening.

Another advantage is that users may already have constructed their binned representations, and thus it does not have to be recreated.

I have added an example to the documentation to show how the function is to be used with the BinnedSpikeTrain object.

PR #34, opened 2015-06-10
complexity_pdf and unit_test

pietroquaglio:
This pull request adds to the statistics module the complexity_pdf() function, in which is computed the density function of the number of active neurons per bin, fixed a certain binsize. The relative unit_test is added to the tests.

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -648,6 +648,63 @@ def time_histogram(spiketrains, binsize, t_start=None, t_stop=None,
                                  t_start=t_start)
 
 
+def complexity_pdf(spiketrains, binsize):
+    """
+    Complexity Distribution [1] of a list of :attr:`neo.SpikeTrain` objects.
+
+    Probability density computed from the complexity histogram which is the
+    histogram of the entries of the population histogram of clipped (binary)
+    spike trains computed with a bin width of binsize.
+    It provides for each complexity (== number of active neurons per bin) the
+    number of occurrences. The normalization of that histogram to 1 is the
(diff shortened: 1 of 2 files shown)


elephant/statistics.py
```diff
@@ -648,6 +648,63 @@ def time_histogram(spiketrains, binsize, t_start=None, t_stop=None,
...
+    # Computing the population histogram with parameter binary=True to clip the
+    # spike trains before summing
+    pophist = time_histogram(spiketrains, binsize, binary=True)
+
+    # Computing the histogram of the entries of pophist (=Complexity histogram)
+    complexity_hist = np.histogram(
+        pophist.magnitude, bins=range(0, len(spiketrains)+2))[0]
+
+    # Normalization of the Complexity Histogram to 1 (probabilty distribution)
+    sum_complexity = float(sum(complexity_hist))
```
apdavison: Minor issue, but `complexity_hist.sum()` is likely to be faster. In addition, `float` should not be necessary provided we have `from __future__ import division` at the top of the file.

elephant/statistics.py
```diff
@@ -648,6 +648,63 @@ def time_histogram(spiketrains, binsize, t_start=None, t_stop=None,
...
+    complexity_hist = [s/sum_complexity for s in complexity_hist]
```
apdavison: `complexity_hist` is an array, so this could just be `complexity_hist /= sum_complexity`

pietroquaglio:
This commit includes the @apdavison  comments

PR #36, opened 2015-06-15
Implementation of covariance

mdenker:
This is an implementation of the `covariance()` function. It is closely related to the `corrcoef()` function, and therefore overlapping code between correlation coefficient and covariance was moved to a new private function.

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -11,15 +11,80 @@ from __future__ import division
 import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


elephant/spike_train_correlation.py
```diff
@@ -11,15 +11,80 @@
 import numpy as np
 
 
+def covariance(binned_sts, binary=False):
```
apdavison: what is an `sts`? I think `binned_spiketrains` would be much clearer.

PR #39, opened 2015-06-16
draft implementation of crosscorrelogram function

btel:
This is my implementation of cross-correlograms. It is not final yet - there are still some work to be done, but hopefully it will initiate some discussion on how to implement a general crosscorrelogram function.

I will appreciate your feedback, especially on:

* the API - what arguments and in what format the function should have,
* which algorithm and normalisation to use in the implementation.


alperyeg:
@btel can you have a look here https://github.com/NeuralEnsemble/elephant/issues/63 and close this PR?

apdavison:
Replaced by #63

Issue #33, opened 2015-06-09
CI Tests fail 


alperyeg:
When doing the PR in https://github.com/NeuralEnsemble/elephant/pull/32 I noticed the nosetests performed by Travis CI fail, with following error:
```python
AssertionError: attr is not equal [names]: FrozenList([None]) != FrozenList([u'0'])
```

I checked and saw that this issue is related to the `test_pandas_bridge.py` unittest. It seems that the `panda` version specified in the install script has problems with the newest `numpy` library or vice versa. 
However after trying locally with all the newest versions the tests still fail. 
I assume the unittest regarding the panda test has to be adapted according to the newest versions.

alperyeg:
@toddrjen any clues on this?

apdavison:
fixed by #46

PR #42, opened 2015-06-17
Fix pandas_bridge and Travis CI 

alperyeg:
This PR changes the `install.sh` and `travis.yml` filles to use the newest libraries (numpy, scipy, pandas) to avoid hard-coding the version numbers and problems with interoperability between the different library versions. 
Additionally I fixed errors in `test_pandas_bridge`.  In the newest(?) panda version `0.16.2` the parameter `names` is also checked, when using `assert_index_equal ` in the unittests. This throws an error (as described here https://github.com/NeuralEnsemble/elephant/issues/33), because in our case e.g., the strings  `t_start` and `t_stop` are checked for equality.  So I set the parameter `names` to `False` not to check for equality.

### DIFF
--- a/elephant/test/test_pandas_bridge.py
+++ b/elephant/test/test_pandas_bridge.py
@@ -235,23 +235,23 @@ class SpiketrainToDataframeTestCase(unittest.TestCase):
         self.assertEqual(keys, res8.columns.names)
 
         for value, level in zip(values, res0.columns.levels):
-            assert_index_equal(value, level)
+            assert_index_equal(value, level, False, False)
         for value, level in zip(values, res1.columns.levels):
-            assert_index_equal(value, level)
+            assert_index_equal(value, level, False, False)
         for value, level in zip(values, res2.columns.levels):
-            assert_index_equal(value, level)
+            assert_index_equal(value, level, False, False)
         for value, level in zip(values, res3.columns.levels):
-            assert_index_equal(value, level)
+            assert_index_equal(value, level, False, False)
         for value, level in zip(values, res4.columns.levels):
-            assert_index_equal(value, level)
+            assert_index_equal(value, level, False, False)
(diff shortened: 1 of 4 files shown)


elephant/test/test_pandas_bridge.py
```diff
@@ -235,23 +235,23 @@ def test__spiketrain_to_dataframe__parents_empty(self):
         self.assertEqual(keys, res8.columns.names)
 
         for value, level in zip(values, res0.columns.levels):
-            assert_index_equal(value, level)
+            assert_index_equal(value, level, False, False)
```
apdavison: (1) please can you add the keywords i.e. `exact=False, check_names=False` to make it clearer what is going on.
(2) is it necessary to have `exact=False`?

apdavison:
One of the things I like about the test matrix is that it tests both recent and old versions of the main dependencies. It's fine to update the most recent version to 1.9, and maybe we can update the "old" version to whatever is used in the most recent Ubuntu LTS release, but removing the hard-coded version numbers altogether makes Elephant more fragile.

elephant/test/test_pandas_bridge.py
alperyeg: I will refactor this. `exact=False` may be set by default already. I have to check.

alperyeg:
Travis uses Ubuntu 12.04, so we have to use the versions supported by 12.04. 
The problem I had was when installing the `pandas` library. If installed per pip it will automatically install the newest numpy. We can try to install an older version of pandas, at least for the Ubuntu distribution. 

I would suggest then not to use pip in the test matrix with the ubuntu distribution but its own package manager, because it seems that this is the most fragile test case.

alperyeg:
I am closing this and will make a new clean PR.

Issue #43, opened 2015-06-23
Extract SpikeTrain from AnalogSignal


rgerkin:
@apdavison Is there a function for this yet?  I didn't see it in a brief search.  If not let me know if I should just contribute the old AnalogSignal.threshold_detection code from NeuroTools, and if so to what module (maybe a new module for AnalogSignal manipulations, which could subsume sta.py since it currently only has one function).

apdavison:
No, there isn't a function for this yet; I'd be happy for you to adapt the code from NeuroTools. Not sure about the module. Maybe `spike_train_generation`?

In general, I'm not too worried at the moment about module naming, and where certain functions should go; I think this is a discussion to have once we have a larger number of functions implemented. At that point we can move things around.

samuelgarcia:
You can also take code from 
https://github.com/OpenElectrophy/OpenElectrophy/tree/master/OpenElectrophy/spikesorting/detection
to port it to elephant.
There are several detection methods there.

apdavison:
Excellent idea @samuelgarcia

apdavison:
On looking at this a second time, are your modules for extracting spikes from extracellular recordings, @samuelgarcia ? I think @rgerkin is talking about the easier task of extracting spikes from intracellular recordings; perhaps we should provide both, although I'm not sure Elephant should tackle issues related to spike sorting, as there are already Python packages for that.

samuelgarcia:
Intracellular spike detection is the most simple case of extra cellular spike detection : one electrode + absolut threshold.
The are two algo implemented about that in OpenElectrophy : one with threshold detection and one with peak detection above a threshold. Both are implemented in numpy style or numepxr.
See
https://github.com/OpenElectrophy/OpenElectrophy/blob/master/OpenElectrophy/spikesorting/detection/tools.py
This very simple codes that can be copy/paste/adapted.

toddrjen:
@apdavison Yes, it seems like spike sorting is a big enough issue on its own to be outside the scope of elephant.  Having it work well with spike sorters makes sense, but I don't think spike sorting itself belongs in elephant.

mdenker:
Yes, I agree, at least for now spike sorting is too big a topic for Elephant, in particular since spike sorting is also something that profits a lot from visualization.

rgerkin:
[...]

PR #45, opened 2015-06-25
Spike Train Extraction from Analog Signals via Threshold Detection

rgerkin:
Addresses #43.  Also added little Izhikevich neuron simulator for generation of membrane potential waveforms for testing the new threshold_detection function.  This simulator could potentially be moved elsewhere.  I thought it was better to simulate real-looking waveforms though, since it will be easier to create a larger collection of robust spike detection tests e.g. on different spike neuron classes.  I only added one test based on a tonic firing neuron.


apdavison:
I think having the Izhikevich generator is not a bad idea, but it needs its own unit tests. Can I suggest you generate some test voltage traces using a reference simulator such as NEST, and use these for unit testing?

rgerkin:
@apdavison Good idea but would it be possible to merge everything except bbf88f2 for the time being?  Then I can get NeuralEnsemble:master into my projects.  Or would that not be OK since bbf88f2 contains, for now, the only test of the spike extraction code?

rgerkin:
@apdavison Another question is whether you want any unit tests to depend upon a NEST installation (this will add 10+ minutes to every Travis run).  Maybe a lighter-weight reference simulator?

rgerkin:
@apdavison Brian2, maybe?  Avoids too many dependencies, Python3 compatible, active development.  Not part of PyNN yet, though, right.

apdavison:
I was suggesting that you use NEST to generate some test data, and add those to the repository. The test would then compare the results of the Izhikevich mini-simulator against the pre-recorded data, not against NEST directly.

rgerkin:
@apdavison OK, got it.

rgerkin:
@apdavison OK that's all taken care of.  I used Brian2 to generate the test data though since NEST was not compiling on my machine.  Documented [here](https://github.com/neuroelectro/elephant/blob/master/elephant/test/make_spike_extraction_test_data.py) which is part of 456ebac

PR #49, opened 2015-08-12
Fix/travis

alperyeg:
This PR is Travis related: 
* Fixed import error in `elephant.spectral`: Travis can't find the `six` library when imported from `scipy.lib`
* Changed to new container system: Our old script is obsolete, but fortunately only small adaptations needed to be done


apdavison:
Many thanks for taking care of that Alper !

PR #51, opened 2015-09-21
updated documentation for 0.2 release

alperyeg:
Please have a look. 
In general a thorough clean up is needed. 
With respect to the short time for the upcoming summit I just updated the docs and added release notes. 
If merged we can give it the 0.2 tag and upload it to pypi.


apdavison:
I don't see the interest in changing the logo. In any case, the new files are far too big; they need to be the same size as will be displayed in the documentation, to avoid resizing and minimize bandwidth.

alperyeg:
>I don't see the interest in changing the logo. In any case, the new files are far too big; they need to be the same size as will be displayed in the documentation, to avoid resizing and minimize bandwidth.

The **E**lephant text one the logo is now corrected. I can make the logo smaller.

apdavison:
I actually prefer "elephant" for the logo (although "Elephant" to refer to the project).

doc/release_notes.rst
```diff
@@ -0,0 +1,29 @@
+*************
+Release Notes
+*************
+
+Elephant 0.2 release notes
+==========================
+September 21th 2015
```
apdavison: "21st" - although probably best to make it "22nd" now!

doc/release_notes.rst
```diff
@@ -0,0 +1,29 @@
...
+* Added **Cumulant Based Inference for higher-order of Correlation (CuBIC)** in the `cubic` module for correlation analysis of parallel recorded spike trains
+
+API changes
+===========
+* **Optimized kernel bandwidth** in `rate_estimation` function: Calculates the optimized kernel width when the paramter kernel width is specified as `auto`
+
+Other changes
+=============
+* **Optimized creation of sparse matrices**: The creation speed of the sparse matrix inside the `BinnedSpikeTrain` class is optimized
+* Added **Izhikevich neuron simulator** for generation of membrane potential waveforms the in `make_spike_extraction_test_data` module
```
apdavison: I think this and the next item are too specific. I would put something like: "minor improvements to the test and continuous integration infrastructure"

doc/release_notes.rst
```diff
@@ -0,0 +1,29 @@
+*************
+Release Notes
+*************
+
+Elephant 0.2 release notes
```
apdavison: "0.2.0" - since there may well be a "0.2.1", etc.

alperyeg:
Ironically (see release notes) the travis build breaks atm.
The travis team is aware of it and investigating: 
https://github.com/travis-ci/travis-ci/issues/4816

alperyeg:
@apdavison are the changes alright now? Can you merge?

mdenker:
Ok, I am fine with the small elephant, I actually also think it looks a bit better, but just thought it might be more consistent with the spelling. I notice also with Neo that people are tempted to write neo instead of Neo. But I agree, let's leave it.

The other subtle change was to make the lower text in gray in order to make it more legible (small orange on white is difficult). I will include this change in a future doc update if that's ok...

doc/install.rst
```diff
@@ -74,8 +74,8 @@ To download and install manually, download the latest package from http://pypi.p
 
 Then::
 
-    $ tar xzf elephant-0.1.tar.gz
-    $ cd elephant-0.1
+    $ tar xzf elephant-0.2.tar.gz
```
apdavison: just realized, this should be "0.2.0" not "0.2"

PR #55, opened 2015-09-29
Inverse methods for current source density (CSD) analysis

espenhgn:
These methods are applicable with 1D laminar electrode recordings. See Pettersen et al. (2006), J Neurosci Meth. http://dx.doi.org/10.1016/j.jneumeth.2005.12.005 for details. 

All files from its original repository/branch https://github.com/espenhgn/iCSD/tree/dev are included.


alperyeg:
@espenhgn can  you check the errors? One is because of plotting, which is not supported in elephant, yet. 
Also, can you rename the `testing` file to `test_csd` or similar? 
The second error comes from the `init` file (`elephant/elephant/icsd/__init__.py`), where the `icsd` cannot be found. Is the path right?

espenhgn:
Seems like Travis is using Python v2.7.3, where I'm getting "TypeError: 'exceptions.AssertionError' object is not callable" and similar, however, "raise Exception("error message") is fine in python 2.7.10. I would choose to ignore it. Furthermore, module quantities appear to be outdated, it does not know about the unit Siemens (SI unit S) for example.

espenhgn:
@alperyeg Failing tests are now in other submodules than the icsd code.

elephant/test/test_icsd.py
```diff
@@ -0,0 +1,1243 @@
...
+    C_i : float*pq.A/pq.m**3
+        current source density on circular disk in units of charge per area
+    R_i : float*pq.m
+        radius of disk source
+    sigma : float*pq.S/pq.m
+        conductivity of medium in units of S/m
+
+    Notes
+    -----
+    Sympy can't deal with eq. 11 in Pettersen et al 2006, J neurosci Meth,
```
alperyeg: Do you use sympy any where in the code? Why this reference?

alperyeg:
One possible approach would be to wait until my fixes are accepted and you update your code after that, so we can check if everything works. 
Otherwise it looks fine for me. Btw. can you squash your commits?

espenhgn:
@alperyeg Squashed commits

apdavison:
Elephant is BSD-licenced, so we can't include GPL-licenced code. It would be great to have this in Elephant, but unfortunately we can't merge this unless the licence is changed.

espenhgn:
@apdavison I will change the license and remove references to GPL. 
@alperyeg said he would look into the build issue with Python 3.4.4, as I could not reproduce it locally. Most likely due to quantities being outdated, not supporting the SI unit Siemens (S).

alperyeg:
Sorry for the long wait. Here is a short update. 
I could reproduce the same assertions errors on our cluster with the travis build 2 (pyhton 3.5, numpy 1.8.2) and the latest quantities version 0.11.0. Furthermore I tested build 5 (ubuntu, python 2.7, numpy 1.8.2) which also produced the same errors. 
I am not sure how to proceed from here on. Either we try to make a fix in your code so it works (if possible) or try to circumvent this problem by not testing for the specific versions. In this case I would like to hear more input from @apdavison  and @mdenker.

espenhgn:
thanks @alperyeg. Will look into testing/debugging on the cluster if I can use the same Python environments as soon as possible

alperyeg:
Thank you Espen for the changes. Can you squash your commit history? After that we can merge!

espenhgn:
Hi @alperyeg. It looks as though you can squash a pull request when merging: https://help.github.com/articles/about-pull-request-merge-squashing/
Can you try that? I'm getting some errors trying to rebase.

PR #56, opened 2015-10-20
added downwards compatibility for numpy versions smaller than 1.9

alperyeg:
I added downwards compatibility for numpy in the `conversion` module. Before I was using a function which was introduced in numpy v1.9. Since we want to support also smaller versions I thought this may be a good idea to include. This code fragment will be called once the `BinnedSpikeTrain` is to going to be created.


alperyeg:
I am closing this in favour of https://github.com/NeuralEnsemble/elephant/pull/58

PR #57, opened 2015-10-23
updated travis and install script

alperyeg:
This is the first part of pull requests to fix the latests bugs. 
I modified the travis.yml and the install script, so that the missing libgfortran error disappears. Furthermore, I restricted the installation of pandas to v.0.16.0 this is the last configuration we know is working with our code. 

Of course all  tests will not run through. 
But I have working branch with everything included [here](https://github.com/alperyeg/elephant/tree/fix/conversion).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
       - liblapack-dev
       - gfortran
       - python-scipy
-      
+
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


alperyeg:
@apdavison @mdenker @Junji110 @toddrjen can s.o. of you have a look at my fixes?

PR #60, opened 2015-10-23
Fixes for spike train threshold detection

rgerkin:
Described in #54.

PR #62, opened 2015-10-28
Fix unsized spike train and np.allclose issues in testing

rgerkin:
@alperyeg 
Replaces #59 and #60.  Must be used with #57 to work in Python 3.


rgerkin:
@alperyeg 
OK, I added the warning.

PR #63, opened 2015-11-24
Added the cross_correlation_histogram() function and relative test

pietroquaglio:
This pull request add the function cross_correlation_histogram() to compute the cross correlation between two binned spike trains. It merge the two pull requests https://github.com/NeuralEnsemble/elephant/pull/40 and https://github.com/NeuralEnsemble/elephant/pull/39 and relative comments.
I would like in particular have some feedback from @mdenker  and @btel, that create the original pull requests. There are still some open points in respect to the previous comments:
* input binned spike trains vs. spike trains in continuous time
* automatic section of the parameter that now is called `method` to  switch between speed version or low memory consuming
* whether to trim the spike trains when the `window` parameter is specified or just cut off the cross-correlation histogram

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -9,6 +9,9 @@ This modules provides functions to calculate correlations between spike trains.
 """
 from __future__ import division
 import numpy as np
+import neo
+import warnings
+import quantities as pq
 
 
 def covariance(binned_sts, binary=False):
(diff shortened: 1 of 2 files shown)


btel:
hi Pietro. Thanks for the work. I will take a look at the PR in the coming days.

elephant/spike_train_correlation.py
```diff
@@ -246,3 +249,408 @@ def __calculate_correlation_or_covariance(binned_sts, binary, corrcoef_norm):
...
+                    'length %d of the resulting CCH.' % (
+                        np.abs(l) + np.abs(r) + 1))
+            kern = np.array(kern, dtype=float)
+            kern = 1. * kern / sum(kern)
+        # Check kern parameter
+        elif kern is not None:
+            raise ValueError('Invalid smoothing kernel.')
+
+        # Smooth the cross-correlation histogram with the kern
+        if kern is not None:
```
btel: some of the code is duplicated in both methods. Could you refactor it into separate functions?

elephant/spike_train_correlation.py
```diff
@@ -246,3 +249,408 @@ def __calculate_correlation_or_covariance(binned_sts, binary, corrcoef_norm):
...
+        cch_result = neo.AnalogSignalArray(
+            signal=counts.reshape(counts.size, 1),
+            units=pq.dimensionless,
+            t_start=(bin_ids[0] - 0.5) * st_1.binsize,
+            sampling_period=st_1.binsize)
+        # Return only the hist_bins bins and counts before and after the
+        # central one
+        return cch_result, bin_ids
+
+    if method is "memory":
```
btel: == is for equality testing

elephant/spike_train_correlation.py
```diff
@@ -246,3 +249,408 @@ def __calculate_correlation_or_covariance(binned_sts, binary, corrcoef_norm):
...
+        if win is not None:
+            # Window parameter given in number of bins (integer)
+            if isinstance(win[0], int) and isinstance(win[1], int):
+                # Check the window parameter values
+                if win[0] >= win[1] or win[0] <= -max_num_bins \
+                        or win[1] >= max_num_bins:
+                    raise ValueError(
+                        "The window exceed the length of the spike trains")
+                # Assign left and right edges of the cch
+                l, r = win[0], win[1]
```
btel: why not `l, r = win`?

elephant/spike_train_correlation.py
```diff
@@ -246,3 +249,408 @@ def __calculate_correlation_or_covariance(binned_sts, binary, corrcoef_norm):
...
+            st1_arr = np.array(st1_arr > 0, dtype=int)
+            st2_arr = np.array(st2_arr > 0, dtype=int)
+        max_num_bins = max(len(st1_arr), len(st2_arr))
+
+        # Cross correlate the spiketrains
+
+        # Case explicit temporal window
+        if win is not None:
+            # Window parameter given in number of bins (integer)
+            if isinstance(win[0], int) and isinstance(win[1], int):
```
btel: do we need this exception? Can't we simply pad with zeros?

elephant/spike_train_correlation.py
```diff
@@ -246,3 +249,408 @@ def __calculate_correlation_or_covariance(binned_sts, binary, corrcoef_norm):
...
+                        or win[1] >= max_num_bins * binsize:
+                    raise ValueError("The window exceed the length of the"
+                                     " spike trains")
+                # Assign left and right edges of the cch
+                l, r = int(win[0].rescale(binsize.units) / binsize), int(
+                    win[1].rescale(binsize.units) / binsize)
+
+            # Cross correlate the spike trains
+            corr = np.correlate(st2_arr, st1_arr, mode='full')
+            counts = corr[len(st1_arr)+l+1:len(st1_arr)+1+r+1]
```
btel: It is not exactly in the spirit of PR #39. Here you calculate the cross-correlation between the whole spike trains, which for long spike trains can take considerable amount of time. Most of the time the correlation window of interest (`win`) is much shorter than the full length of both spike trains,  so that the `correlate` function returns arrays much longer than necessary.  In #39 we trim the spike trains such that the `convolve` (and so `correlate`) returns an array only with the bins of interest. In most cases this is a huge performance boost.

btel:
We should do some benchmarks on this code. Apart from the issue I added as a line comment, the `_cch_speed` method will need to convert the sparse binned spikes array to dense object (when implementing #39 I was not aware that `BinnedSpikeTrain` is implemented as a sparse array), which is a big performance drain. We need to make sure that `_cch_speed` is really faster than `_cch_memory` in real-like use cases.

alperyeg:
@btel Pietro and I will have a look next week. Thank you for the comments. At the code jam I discussed with Michael this PR and we had similar remarks.

pietroquaglio:
@btel :
 I implemented some of your suggestion (refactored code + small changes). 
* Regarding the performances we applied both methods on two spike trains with rate 100Hz and for 100 seconds and we figured out that the current version using `_cch_memory` is 6 times faster than `_cch_speed`.  In order to improve the performances for high firing rate we would suggest to change `numpy.correlate` with `scipy.signal.fftconvolve`. What is your opinion about it?
* The reason for which we didn't trim the spike train is that we thought to make changes to the data implicitly in the function. The user could decide on is own to preprocessing the data on his own in order to exploit the options of the function (e.g. using together the option `mode=valid` and  `win` ).

btel:
@pietro thanks for the changes. 

I am not surprised about the differences in the performance. Replacing `correlate` with `fftconvolve` might improve the performance, but in my opinion the bottleneck is that in either case we calculate the correlograms for all possible lags, while we need only a very short window. The trimming was a way to alleviate the problem; it used the `"valid"` mode of correlate to calculate and return only short segment of the correlogram. I agree it's a bit of hack and does not use full data, but for longer spike trains it makes little difference.

If you send me your benchmarking script, I can do more test and see whether `_cch_speed` can be improved. If not, than I would suggest to drop `_cch_speed` implementation altogether.

alperyeg:
@btel: Here are the [benchmark scripts](https://gist.github.com/alperyeg/b6d796cb69c620433441)
There are 3 files. The first result is obtained by [line_profiler](https://github.com/rkern/line_profiler). That's why you will see `@profile` decorator commented out. The second script (profile_cch.py) is the one I used to run the benchmarks. And the last one ( spike_train_correlation.py) includes the `cch` using `fftconvolve`. 
You can run the second script alone to have a general idea how fast the methods are. But as expected for low firing rates the `memory` method is to favour, for higher rates it is best to use the `speed` method.

btel:
Hi Pietro, sorry for the delay.

I did some further optimizations in both methods - some of the improvements in performance are substantial. Look at the [results](https://gist.github.com/btel/9d5c98ca27d24780a4a4#file-profile_stats-txt). You can check the code out from my [branch](https://github.com/btel/elephant/tree/combined_cch). To ease the comparison, I made a PR on your branch: https://github.com/pietroquaglio/elephant/pull/1

I also added an extra test to make sure that the xcorrs in all windows are correctly calculated (it was relevant for some of the optimizations that I made).

I adapted your profiling scripts to run benchmarks for more sets of parameters. The conclusions are:

* the memory method is much faster (~1000 times) for small rates (probability of a spike per bin ~0.01) and large CCH windows (at the order of the spike train length)
* the speed method is ~10 times faster for large rates (the prob of a spike per bin ~1) and large CCH windows
* the speed method is ~100 times faster for large rates and small CCH windows (~100 bins)
* for small windows and small rates the methods are comparable.

I think we should do such test more systematic and switch the methods automatically (when `method='auto'`) depending on the data passed to the function. Of course, we can always keep the option for manual method selection

PR #65, opened 2015-12-04
Kernel class file and adaptation of instantaneous_rate function

dholstein:


elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+import scipy.special
+
+default_kernel_area_fraction = 0.99999
+
+
+class Kernel(object):
+    """
+    Base class for kernels.
+
+    :param sigma: Standard deviation of the kernel.
```
alperyeg: Please adapt to numpy doc. 
```
Parameters
---------------
Sigma: ....
```

elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+    having any effect in the case of symmetric kernels.
+    """
+    min_cutoff = np.sqrt(5.0)
+
+    def _evaluate(self, t):
+        return (3.0 / (4.0 * np.sqrt(5.0) * self._sigma_scaled)) * np.maximum(
+            0.0,
+            1 - (t / (np.sqrt(5.0) * self._sigma_scaled)).magnitude ** 2)
+
+    def boundary_enclosing_at_least(self,
```
alperyeg: Needs documentation.

elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+    parameter `direction` needed for asymmetric kernels also exists without
+    having any effect in the case of symmetric kernels.
+    """
+    min_cutoff = 3.0
+
+    def _evaluate(self, t):
+        return (1 / (np.sqrt(2.0) * self._sigma_scaled)) * np.exp(
+            -(np.absolute(t) * np.sqrt(2.0) / self._sigma_scaled).magnitude)
+
+    def boundary_enclosing_at_least(self,
```
alperyeg: Here also documentation needed.

elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+    def _evaluate(self, t):
+        """ Evaluates the kernel.
+
+        :param t: Interval on which the kernel is evaluated, not necessarily
+            a time interval.
+        :type t: Quantity 1D
+        :returns: The result of the kernel evaluations.
+        :rtype: Quantity 1D
+        """
+        raise NotImplementedError()
```
alperyeg: Do you intend to provide an implementation later, or why do you raise a  `NotImplementedError`?

elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+        The formula in this method using retrieval of the sampling interval
+        from t only works for t with equidistant intervals!
+        The formula calculates the Median slightly wrong by the potentially
+        ignored probability in the distribution corresponding to lower values
+        than the minimum in the array t.
+        """
+        return np.nonzero(self(t).cumsum() *
+                          (t[len(t)-1] - t[0]) / (len(t) - 1) >= 0.5)[0].min()
+
+    def is_symmetric(self):
```
alperyeg: Shouldn't this return `True`?

elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+        """ Should return `True` if the kernel is symmetric. """
+        return False
```
alperyeg: Shouldn't this return `True`?

elephant/statistics.py
```diff
@@ -422,7 +433,8 @@ def make_kernel(form, sigma, sampling_period, direction=1):
     return kernel, norm, m_idx
 
 
-def instantaneous_rate(spiketrain, sampling_period, form,
+# to finally be taken out of Elephant
+def oldfct_instantaneous_rate(spiketrain, sampling_period, form,
```
alperyeg: Maybe a warning indicating that this function is deprecated  would be good.

elephant/kernels.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+class SymmetricKernel(Kernel):
+    """ Base class for symmetric kernels. """
+
+    # def __init__(self, sigma, direction):
```
alperyeg: Comments here should be removed.

alperyeg:
Are the unittests for the kernel class the same as the old one? Was just wondering because I could not see any test file.

dholstein:
test_statistics.py now adapted to the new version of the function instantaneous_rate and test_kernels.py added

elephant/kernels.py
dholstein: Changed towards style of numpy doc.

elephant/kernels.py
dholstein: The NotImplementedError message is now verbose. It is not intended to provide an implementation later, however, it is preferred to show the full list of existing methods of derived classes in the parent class for giving the user directly a list of methods to care for, if intending to supply further kernels in the future.

elephant/kernels.py
dholstein: I think returning 'False' is correct here. In the case of symmetric kernels, this method is overwritten in the class SymmetricKernel and then returns 'True', remaining with 'False' for the asymmetric ones.

elephant/statistics.py
dholstein: Several deprecation warnings introduced in statistics.py.

elephant/kernels.py
alperyeg: Ok can you add your comment to the documentation? Then it should be clear.

elephant/kernels.py
dholstein: Decorator inherit_docstring introduced, from which the methods boundary_enclosing_area_fraction and _evaluate now have documentation inherited from the parent class and optionally further kernel-specific documentation.

elephant/kernels.py
dholstein: True, the documenting line wasn't very suitable. Changed now.

elephant/kernels.py
dholstein: Comments of this type are removed now.

dholstein:
Pullrequests for kernels.py and rate_estimation in statistics.py will be separated.

PR #69, opened 2016-02-15
Updated documentation/requirements files to use a specific Neo snapshot.

apdavison:
(also added the `six` package, which is used by the `spectral` module).

This is needed so that development can continue in the Neo `apibreak` branch.


apdavison:
This PR seems to have opened a can of worms with respect to the conda_min test environment for Python 2.7, so I'm going to close it until I get Travis sorted out.

PR #72, opened 2016-02-16
Updated documentation/requirements files to use a specific Neo snapshot.

apdavison:
(also added the six package, which is used by the spectral module).

This is needed so that development can continue in the Neo apibreak branch.


apdavison:
I can't get the hang of this rebase stuff. I don't understand why there are two commits, there should be only one. I'm going to close this again and start again.

PR #74, opened 2016-02-17
Bug-fix release, 0.2.1

apdavison:
I'd like to make a release mainly to update the installation instructions concerning Neo, but there have been quite a few fixes in the master branch, so it's probably a good time anyway.


mdenker:
I agree, good point. Changes are all good.

PR #76, opened 2016-03-03
Added ASSET code + unittests

etorre:
This pull request adds an analysis known as ASSET (analysis of sequences of synchronous events) which searches sequences of synchronous spikes in parallel spike trains ("synfire-like activity"), tests their statistical significance, and reconstructs the chain of neuron groups that participated in the activity.

### DIFF
--- a/elephant/asset.py
+++ b/elephant/asset.py
@@ -1,0 +1,1712 @@
+"""
+ASSET is a statistical method [1] for the detection of repeating sequences
+of synchronous spiking events in parallel spike trains.
+Given a list `sts` of spike trains, the analysis comprises the following
+steps:
+1) Build the intersection matrix `imat` (optional) and the associated
+   probability matrix `pmat` with the desired bin size:
+   `>>> binsize = 5 * pq.ms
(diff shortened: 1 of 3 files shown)


alperyeg:
Ah yes, I totally forgot to say that the build is going to fail since we do not have scitkit learn in the dependicies list. We need to make scikit learn optional as we do for pandas.

etorre:
You're right. I just added the optional dependency to requirements.txt:
#sklearn>=0.15.1
and rebased the log history. Have a look and let me know

alperyeg:
No I meant something like  [this](https://github.com/NeuralEnsemble/elephant/blob/master/elephant/test/test_pandas_bridge.py) and for the install script we need to add an optional settings [here](https://github.com/NeuralEnsemble/elephant/blob/master/continuous_integration/install.sh) and [here](https://github.com/NeuralEnsemble/elephant/blob/master/.travis.yml). I can take of that later.

alperyeg:
@apdavison Do you know why Travis is not running on this pull request any more? Very strange...

PR #79, opened 2016-03-17
added package scikit-learn

alperyeg:
This PR addresses the missing packages in PR https://github.com/NeuralEnsemble/elephant/pull/76. Simultaneously it fixes a travis issue, where libgfortran is messed up again. I included `conda install libgfortran` directly when conda is invoked. 
For the moment I don't give a specific `scikit-learn` version, and travis grabs the latest one. It could later be fixed to a static version.


apdavison:
I thought I fixed the libgfortran issue in a commit yesterday. At any rate the build is passing on master.
I also don't understand how installing the package in the base conda, but not in the testenv, helps. 
Finally, I think we should keep the pin to version 1 of libgfortran, as I found that the build broke when allowing conda to install libgfortran 3.0.

alperyeg:
Well I didn't see you fixed and merged that already, and since the error appeared while intalling numpy I thought it would be better to have it install even before the testenv is created and all packages installed. 
I am going to close this PR then and reopen only with the new additional package.

PR #82, opened 2016-03-23
spike field coherence (sfc) function and unit tests

JuliaSprenger:
This pull request contains the spike field coherence function and its unit tests. The function is contained in the sta.py file, because we did not want to add additional files which only contain a single function and the sta and sfc functions are related.

### DIFF
--- a/elephant/sta.py
+++ b/elephant/sta.py
@@ -181,10 +181,10 @@ def spike_field_coherence(signal, spiketrain, **kwargs):
     Calculates the spike-field coherence between a analog signal(s) and a
     (binned) spike train.
 
-    The current implementation makes use of scipy.signal.cohere(). Additional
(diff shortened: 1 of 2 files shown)


alperyeg:
@JuliaSprenger thanks for the corrections. In the unittest it is better to use decorators, e.g. as it was done in the `asset unittest` [here](https://github.com/etorre/elephant/blob/feature/asset_code/elephant/test/test_asset.py) (L14-25). Or in the same manner you can use the following: 
```python
@unittest.skipIf(not hasattr(scipy.signal, 'coherence'),
                 "Please update scipy to version ...")
```

mdenker:
I could verify that this works to correctly catch missing cohere on both 0.14 and 0.15 scipy -- thanks for fixing this.

alperyeg:
:+1:

PR #83, opened 2016-03-29
This pull request contains the kernel functions and its unit tests.

dholstein:
The kernel module provides a class to calculate a variety of kernels that can be employed in a number of analysis routines, e.g., for smoothing.

### DIFF
--- a/elephant/kernels.py
+++ b/elephant/kernels.py
@@ -1,0 +1,518 @@
+# -*- coding: utf-8 -*-
+"""
+Definition of a hierarchy of classes for kernel functions to be used
+in convolution, e.g., for data smoothing (low pass filtering) or
+firing rate estimation.
+
+Examples of usage:
+kernel1 = kernels.GaussianKernel(sigma=100*ms)
+kernel2 = kernels.ExponentialKernel(sigma=8*mm, invert=True)
+
+:copyright: Copyright 2015 by the Elephant team, see AUTHORS.txt.
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+
+import quantities as pq
+import numpy as np
+import scipy.special
+
+def inherit_docstring(fromfunc, sep=""):
+    """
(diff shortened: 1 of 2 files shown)


mdenker:
We have a theory of what goes wrong.

The evaluate functions of the kernels partly use `np.piecewise` to evaluate the kernel at given time points.

In older numpy versions, `np.piecewise` will not accept inputs that are not array-like. Thus, these `_evaluate` functions will not work, if the kernel is only to be evaluated at one particular input value.

For AlphaKernel, there is no dedicated analytic version of the `boundary_enclosing_fraction` function, so the generic version of the parent class is called. This function, however, calls evaluate for a single input value. Hence, the tests fail only for `AlphaKernel` with the conda_min build (which has an old numpy).

In either case, evaluate should be able to deal with single input values, not only because of `boundary_enclosing_fraction`. Thus, the `evaluate` functions of all kernels should be fixed accordingly. 

I see two options: 

1) in `_evaluate`, if a single value t is passed, it is converted to an array or a list. This is potentially a bit slow?

2) `np.piecewise` could be avoided entirely, e.g., using something like (here, for the example of `_evaluate` of the AlphaKernel):

```python
    @inherit_docstring(Kernel._evaluate)
    def _evaluate(self, t):
        if not self.invert:
            kernel = (t > 0) * 2. * (t / self._sigma_scaled**2).magnitude *\
                np.exp((
                    -t * np.sqrt(2.) / self._sigma_scaled).magnitude) / t.units
        elif self.invert:
            kernel = (t < 0) * -2. * (t / self._sigma_scaled**2).magnitude *\
                np.exp((
                    t * np.sqrt(2.) / self._sigma_scaled).magnitude) / t.units
        return kernel
```

dholstein:
np.piecewise replaced by suggested indicator function construction (zero included) for both asymmetric kernels.

PR #84, opened 2016-03-29
This pull request contains changes to the function instantaneous_rate in statistics.py and corresponding unit tests.

dholstein:


elephant/statistics.py
```diff
@@ -13,8 +13,11 @@
 import scipy.stats
 import scipy.signal
 import neo
-import warnings
+from neo.core import SpikeTrain
 import elephant.conversion as conv
+import kernels
```
alperyeg: Please write `import elephant.kernels as kernels`. This then should work also with the python3 builds of travis.

dholstein: Done

mdenker:
Verified that old make_kernel() implementation matches in principle new kernel class implementation. 

Minor deviations between the previous instantaneous_rate() and the current implementation seem to originate from the fact that for defining the normalization factor the new kernel class will not clip the kernel width the the prescribed time grid, but sticks to the full kernel representation on the original time scale. If required, this clipping is a feature that could be added to the kernels module in a later step.

PR #85, opened 2016-03-30
This pull request contains single unit Victor-Purpura and Van Rossum spike train metrics.

dholstein:


mdenker:
Personally, I find the module name a bit long. I understand the argument of not taking spike_train_metrics, but it seems 

 - spike_train_synchrony
 - spike_train_similarity
 - spike_train_dissimilarity

could be alternatives that are a bit shorter?

dholstein:
Among the 3 suggestions for the filename I have a preference for the 3rd: spike_train_dissimilarity. Changes made accordingly.

elephant/spike_train_dissimilarity.py
```diff
@@ -0,0 +1,406 @@
...
+
+:copyright: Copyright 2015 by the Elephant team, see AUTHORS.txt.
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+
+import quantities as pq
+import numpy as np
+import scipy as sp
+import elephant.kernels as kernels
+import cmath
```
mdenker: Is cmath required?

elephant/spike_train_dissimilarity.py
```diff
@@ -0,0 +1,406 @@
...
+        Default: None
+    sort: bool 
+        Spike trains with sorted spike times will be needed for the
+        calculation. You can set `sort` to `False` if you know that your
+        spike trains are already sorted to decrease calculation time.
+        Default: True
+    algor: string
+        Allowed values are 'fast_algor' or 'intuitive_algor', each selecting an
+        algorithm with which to calculate the pairwise Victor-Purpura distance.
+        Default: 'fast_algor'
```
mdenker: I think the user should have a hint here, on what to choose. Although you argue down below why both algorithms are kept, this doc string should probably contain a small hint that for all practical purposes, fast_algor should be used, and that there is now problem in using the efficient version (i.e., identical results).

Also, I would suggest to use 'fast" and 'intutitive" instead of 'fast_algor' and 'intutitive_algor', since otherwise code will look double

...algor="fast_algor"

I would also suggest to use 'algorithm' as variable name, 'algor' seems a bit strange as an abbreviation (algor, algoris: coldness, frost).

elephant/spike_train_dissimilarity.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Given the average number of spikes :math:`n` in a spike train and
+    :math:`N` spike trains the run-time complexity of this function is
+    :math:`O(N^2 n^2)` and :math:`O(N^2 + n^2)` memory will be needed.
+
+    Parameters
+    ----------
+    trains : Sequence of :class:`neo.core.SpikeTrain` objects of
+        which the distance will be calculated pairwise.
+    q: Quantity scalar
```
mdenker: I see that below you explicitly allow for q=0 and q=np.inf. Please indicate these two special cases and their meaning here (otherwise users will not know that they can actually specify inifinite kernels).

elephant/spike_train_dissimilarity.py
```diff
@@ -0,0 +1,406 @@
...
+        spike_counts = [st.size for st in trains]
+        return np.sqrt(spike_counts + np.atleast_2d(spike_counts).T)
+    elif tau == np.inf:
+        spike_counts = [st.size for st in trains]
+        return np.absolute(spike_counts - np.atleast_2d(spike_counts).T)
+
+    k_dist = _summed_dist_matrix(
+        [st.view(type=pq.Quantity) for st in trains], tau, not sort)
+    vr_dist = np.empty_like(k_dist)
+    # for i, j in np.ndindex(*k_dist.shape):
```
mdenker: This commented line could go out?

dholstein:
All 4 correction suggestions of 04.04.16 included

PR #92, opened 2016-07-10
Detect peaks in SpikeTrain and extract SpikeTrain with waveforms

danielmuellernai:
Added two functions: peak_detection detects events by threshold similar to threshold_detection but returns a SpikeTrain where the time stamps represent the peak of the event.
spike_extraction returns a SpikeTrain where the .waveform attribute contains the waveforms of the spikes.
Both functions are essential in my analysis of action potential and EPSC properties from intracellular recordings. I've been working with them on real in-vitro data.
I wrote test cases for peak_detection on the sample data. Test cases for spike_extraction will follow.

### DIFF
--- a/elephant/spike_train_generation.py
+++ b/elephant/spike_train_generation.py
@@ -210,13 +210,15 @@ def peak_detection(signal, threshold=0.0 * mV, sign='above', format=None):
     else:
         # Select thr crossings lasting at least 2 dtps, np.diff(cutout) > 2
         # This avoids empty slices
-        border_start = np.where(np.diff(cutout) > 2)[0]
-        border_end = np.where(np.diff(cutout) > 2)[0] + 1
+        border_start = np.where(np.diff(cutout) > 1)[0]
+        border_end = np.where(np.diff(cutout) > 1)[0] + 1
         borders = np.concatenate((border_start, border_end))
         borders = np.append(0, borders)
         borders = np.append(borders, len(cutout)-1)
         borders = np.sort(borders)
         true_borders = cutout[borders]
+        right_borders = true_borders[1::2] + 1
+        true_borders = np.sort(np.append(true_borders[0::2], right_borders))
 
         # Workaround for bug that occurs when signal goes below thr for 1 dtp,
         # Workaround eliminates empy slices from np. split
--- a/elephant/spike_train_generation.py
+++ b/elephant/spike_train_generation.py
@@ -17,12 +17,13 @@ from neo import SpikeTrain
 import random
 from elephant.spike_train_surrogates import dither_spike_train
 
-def spike_extraction(signal, threshold = 0.0 * mV, sign = 'above',
-                     time_stamps = None, extr_interval = (-2 * ms, 4 * ms)):
+
+def spike_extraction(signal, threshold=0.0 * mV, sign='above',
+                     time_stamps=None, extr_interval=(-2 * ms, 4 * ms)):
     """
-    Return the peak times for all events that cross threshold and the waveforms.
-    Usually used for extracting spikes from a membrane potential to calculate
-    waveform properties
+    Return the peak times for all events that cross threshold and the
+    waveforms. Usually used for extracting spikes from a membrane
+    potential to calculate waveform properties.
     Similar to spike_train_generation.peak_detection.
 
     Parameters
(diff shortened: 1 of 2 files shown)


elephant/spike_train_generation.py
```diff
@@ -17,6 +17,94 @@
 import random
 from elephant.spike_train_surrogates import dither_spike_train
 
+def spike_extraction(signal, threshold = 0.0 * mV, sign = 'above',
```
apdavison: From PEP8: "Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value."

elephant/spike_train_generation.py
```diff
@@ -17,6 +17,94 @@
...
+    ----------
+    signal : neo AnalogSignal object
+        'signal' is an analog signal.
+    threshold : A quantity, e.g. in mV
+        'threshold' contains a value that must be reached for an event 
+        to be detected.
+    sign : 'above' or 'below'
+        'sign' determines whether to count thresholding crossings
+        that cross above or below the threshold.
+    spike_train: None, quantity array or Object with .times interface
```
apdavison: Above, the argument is named "time_stamps", not "spike_train"

apdavison:
Many thanks for this. I noted a couple of PEP8 formatting issues in the code. The tools `pep8` and `autopep8` can be useful for catching these (although you should always manually review the output of `autopep8` since it can produce some ugly formatting).

elephant/test/test_spike_train_generation.py
```diff
@@ -16,7 +16,7 @@
 from numpy.testing.utils import assert_array_almost_equal
 from scipy.stats import kstest, expon
 from quantities import ms, second, Hz, kHz, mV, dimensionless
-
+import matplotlib.pyplot as plt
```
alperyeg: Matplotlib is not used within Elephant atm. If you remove this import than the unittests should work again.

danielmuellernai:
I fixed the pep8 formatting problems manually. The pep8 tool was very helpful with that, thank you for the tip. I also added a test case for spike_extraction. If you find any other problems or necessary improvements, I would be happy to work on them.

elephant/spike_train_generation.py
```diff
@@ -18,6 +18,104 @@
...
+        to be detected.
+    sign : 'above' or 'below'
+        'sign' determines whether to count thresholding crossings
+        that cross above or below the threshold.
+    time_stamps: None, quantity array or Object with .times interface
+        if 'spike_train' is a quantity array or exposes a quantity array
+        exposes the .times interface, it provides the time_stamps
+        around which the waveform is extracted. If it is None, the
+        function peak_detection is used to calculate the time_stamps
+        from signal.
```
alperyeg: Can you add `Default: None` to indicate the default state (below, too).

elephant/spike_train_generation.py
```diff
@@ -18,6 +18,104 @@
...
+    extr_interval: unpackable time quantities, len == 2
+        'extr_interval' specifies the time interval around the
+        time_stamps where the waveform is extracted. The default is an
+        interval of 6 * ms.
```
alperyeg: Here I would put `` around the 6 * ms and omit the *.

alperyeg: I would also mention that you set the default from -2 ms to 4 ms and that the type is a tuple of quantities.

elephant/spike_train_generation.py
```diff
@@ -72,6 +170,86 @@ def threshold_detection(signal, threshold=0.0 * mV, sign='above'):
     return result_st
 
 
+def peak_detection(signal, threshold=0.0 * mV, sign='above', format=None):
+    """
+    Return the peak times for all events that cross threshold.
+    Usually used for extracting spike times from a membrane potential.
+    Similar to spike_train_generation.threshold_detection. Returns
```
alperyeg: Sentence is missing `... Returns`?

elephant/spike_train_generation.py
```diff
@@ -72,6 +170,86 @@ def threshold_detection(signal, threshold=0.0 * mV, sign='above'):
...
+        cutout = np.where(signal < threshold)[0]
+        peak_func = np.argmin
+
+    if len(cutout) <= 0:
+        events_base = np.zeros(0)
+    else:
+        # Select thr crossings lasting at least 2 dtps, np.diff(cutout) > 2
+        # This avoids empty slices
+        border_start = np.where(np.diff(cutout) > 1)[0]
+        border_end = np.where(np.diff(cutout) > 1)[0] + 1
```
alperyeg: Isn't `border_end = border_start + 1` ?

elephant/spike_train_generation.py
```diff
@@ -18,6 +18,104 @@
...
+    # len(np.shape(waveforms)) == 1 if waveforms do not have the same width.
+    # this can occur when extr_interval indexes beyond the signal.
+    # Workaround: delete spikes shorter than the maximum length with
+    if len(np.shape(waveforms)) == 1:
+        max_len = (np.array([len(x) for x in waveforms])).max()
+        to_delete = np.array([idx for idx, x in enumerate(waveforms)
+                             if len(x) < max_len])
+        waveforms = np.delete(waveforms, to_delete, axis=0)
+        waveforms = np.array([x for x in waveforms])
+        print("WARNING: Waveforms " +
```
alperyeg: Why don't you use the warning module?

elephant/test/test_spike_train_generation.py
```diff
@@ -69,6 +68,66 @@ def test_threshold_detection(self):
...
+        self.true_time_stamps = [0.0124,  0.0354,  0.0713,  0.1192,  0.1695,
+                                 0.2201,  0.2711] * second
+
+    def test_peak_detection_time_stamps(self):
+        # Test with default arguments
+        result = stgen.peak_detection(self.vm)
+        self.assertEqual(len(self.true_time_stamps), len(result))
+        self.assertIsInstance(result, neo.core.SpikeTrain)
+
+        try:
```
alperyeg: What is this test doing? Can you give a short explanation? More precise, why is it that the array is almost equal and sometimes not?

elephant/spike_train_generation.py
danielmuellernai: I did not use the warning module because it was not imported in spike_train_generation.py when I started working and I did not want to introduce additional imports. But I can use the warning module if it is ok to import.

elephant/test/test_spike_train_generation.py
danielmuellernai: assert_array_almost_equal asserts the equality of arrays up to a decimal place (6 decimals by default). This is not necessary for integers, which is why self.assertEqual is used to compare the length of arrays. To be honest I took the try: except: construct from the already implemented AnalogSignalThresholdDetectionTestCase and I am not entirely sure what it does and why it catches specifically the AttributeError but I will look into it.

elephant/spike_train_generation.py
alperyeg: Yes it is ok.

elephant/test/test_spike_train_generation.py
alperyeg: If the integers are same then `assert_array_almost_equal` is anyway not giving an error. So, I think you could just leave only the `assert_array_almost_equal` without the try except block. Also In the  `AnalogSignalThresholdDetectionTestCase` I am wondering what it does. Furthermore, it excepts an `AttributeError` whereas  I would expect an `AssertionError`.

danielmuellernai: In `AnalogSignalThresholdDetectionTestCase` it is commented that the try: except AttributeError: is supposed to catch the case `#If` numpy version too old to have `allclose` but from looking at the source code, `assert_array_almost_equal` does not use `allclose` in numpy v1.12.0. Even if it did, it would be a major bug if assert_array_almost_equal would use a not implemented function in any numpy version.

alperyeg: Ah, ok I see now. Thanks for clearing this up! :+1:

alperyeg:
If there are no other comments, remarks, suggestions etc., I would like to merge this PR.

PR #94, opened 2016-08-02
Fixes import elephant after install, default spatial filter applied as convolution, cleanup

espenhgn:


espenhgn:
Fixes/additions:
- elephant/icsd/*.py files not copied by distutils setup.py script
- spatial filter applied using scipy.signal.convolution by default, not using scipy.signal.filtfilt
- the StandardCSD.calc_csd() method can be called in succession. 
- cleaned up some trailing whitespaces

PR #98, opened 2016-12-01
included library six into required packages

alperyeg:
[SIX](https://pypi.python.org/pypi/six) is a library to ensure compatibility between Python 2 and 3. When installing `Elephant` it is not installed but the package is used already, e.g., in `spectral.py`. It could be useful in other modules, too. For instance in `asset.py` I would like to use the `xrange` function provided by `SIX`to be compatible with Python 2 and 3 and instead remove the [patch](https://github.com/NeuralEnsemble/elephant/blob/master/elephant/asset.py#L60) we have there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,8 @@ long_description = open("README.rst").read()
 install_requires = ['neo>0.3.3',
                     'numpy>=1.8.2',
                     'quantities>=0.10.1',
-                    'scipy>=0.14.0']
+                    'scipy>=0.14.0',
+                    'six']
 extras_require = {'pandas': ['pandas>=0.14.1'],
                   'docs': ['numpydoc>=0.5',
                            'sphinx>=1.2.2'],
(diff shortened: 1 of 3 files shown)


mdenker:
Fine with me...

setup.py
```diff
@@ -7,7 +7,8 @@
 install_requires = ['neo>0.3.3',
                     'numpy>=1.8.2',
                     'quantities>=0.10.1',
-                    'scipy>=0.14.0']
+                    'scipy>=0.14.0',
+                    'six']
```
mdenker: Why not `six>=1.10.0` here?

alperyeg: No real reason. `Six` is not updated since 2015. Anyway pip would grab version 1.10.0 which is the latest one.

setup.py
```diff
@@ -8,7 +8,7 @@
                     'numpy>=1.8.2',
                     'quantities>=0.10.1',
                     'scipy>=0.14.0',
-                    'six']
+                    'six>=1.10.0']
```
alperyeg: @mdenker I added the version number.

PR #99, opened 2017-01-31
Use Neo 0.5 API

apdavison:


alperyeg:
@apdavison we have quite a few problems in the csd module. How should we handle this?

alperyeg:
@ccluri and @espenhgn  can you also have a look? The problems are mostly due to the new changes of `neo 0.5` (`RecordingChannel` is removed and replaced by `ChannelIndex`). I could mostly solve the problems just by replacing the correct objects and variables. Usually it explodes in the module `iscd` line 75 :) I will also investigate further.

apdavison:
This is because I did the work prior to the CSD branch being merged, so the `csd` module still needs to be updated to the 0.5 API. I would be happy if someone is able to take care of this; if not, I will tackle it when I return from the HBP Malaga meeting, i.e. in a week or so.

ccluri:
I can look at this only after 23rd February. I have my thesis defense. thanks.

alperyeg:
I am having a look with Julia (@JuliaSprenger). We could already fix some issues (some unrelated to csd). If there are any problems I will comment again. @apdavison should we PR against your fork then or what is the best strategy?

apdavison:
Yes, a PR against my fork (channelindex branch, I think) sounds like the best approach

alperyeg:
Julia and I have fixed the problems which occurred so far with the unittests. You can have a look at my [fork](https://github.com/alperyeg/elephant/tree/apdavison-neo-channelindex) There is still a strange problem, which I couldn't solve and need help with (that's why I didn't PR yet). Travis fails on one test which is no problem on every other install/ version configuration I tested so far (local and on our cluster). 
[Here](https://travis-ci.org/alperyeg/elephant/jobs/202649085) you can see in the log file that the dither function in the spike train surrogate method procduces the same result, which shouldn't be the case unless numpy random is not working correctly on travis. I have atm. no clue how to approach the problem.

alperyeg:
Alright, the fixes are out and PRed to your branch @apdavison.

alperyeg:
I would like to merge this PR as soon as possible, if there are no objections. 
After the merge we should also publish a new version of elephant.

apdavison:
ok with me

PR #101, opened 2017-03-22
bug fixed in surrogate_TrialByTrial method

Vahidrostami:


alperyeg:
Thanks for the fix! I will merge as soon as the tests are done.

PR #104, opened 2017-05-05
Bug fixing for isi function

pietroquaglio:
Two small fixes to the isi function:
* If the input is neo.SpikeTrain wchich t_start is larger than the smallest isi, the function was returning an error, because np.diff(neo.SpikeTrain) returns a neo,SpikeTrain + changed the test to cover such corner case
* Sorted the times of the spikes in input (only if sorted it makes sense to take the diff to compute the isis)

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -51,7 +51,6 @@ def isi(spiketrain, axis=-1):
             np.sort(spiketrain.times.view(pq.Quantity)), axis=axis)
     else:
         intervals = np.diff(np.sort(spiketrain), axis=axis)
-    if hasattr(spiketrain, 'waveforms'):
         intervals = pq.Quantity(intervals.magnitude, units=spiketrain.units)
     return intervals
 
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -46,7 +46,11 @@ def isi(spiketrain, axis=-1):
     """
     if axis is None:
         axis = -1
-    intervals = np.diff(spiketrain, axis=axis)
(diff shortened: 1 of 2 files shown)


alperyeg:
I think the sorting issue should also be addressed in the `neo.SpikeTrain` module. What's your thought on this @apdavison ?

mdenker:
I think the multiple if's are redundant, since
`if hasattr(spiketrain, 'waveforms'):`
essentially asks whether this is a spiketrain object. Thus, this if statement should be merged with the preceding
`if isinstance(...`

pietroquaglio:
@mdenker: I agree (done it!)

alperyeg:
I think Michael meant  the other way around. Merge it to the hasattribute condition.

elephant/statistics.py
```diff
@@ -46,8 +46,11 @@ def isi(spiketrain, axis=-1):
...
     if axis is None:
         axis = -1
-    intervals = np.diff(spiketrain, axis=axis)
-    if hasattr(spiketrain, 'waveforms'):
+    if isinstance(spiketrain, neo.SpikeTrain):
+        intervals = np.diff(
+            np.sort(spiketrain.times.view(pq.Quantity)), axis=axis)
+    else:
+        intervals = np.diff(np.sort(spiketrain), axis=axis)
         intervals = pq.Quantity(intervals.magnitude, units=spiketrain.units)
```
alperyeg: This causes an error, since you do not have neo spiketrains anymore but plain numpy arrays.

pietroquaglio: True sorry that was very lazy of me! Going to fix it!

PR #106, opened 2017-05-09
Fix waveform extraction for protruding time windows

JuliaSprenger:
Previously, waveforms of spikes which were too close to the beginning or end of the analogsignal were just deleted, resulting in inconsistent number of spiketimes in comparison to the extracted waveforms. This lead to an Error raised by the SpikeTrain.
With this PR, incomplete waveforms are replaced by arrays of Nones, therefore restoring the consistency of the spike times array and the waveform array.


JuliaSprenger:
I am not sure why the test is failing. I don't see the connection between my changes and the error.

JuliaSprenger:
I did multiple other changes in the spiketrain_generation.py file and the corresponding tests, also in relation to multidimensionality of analogsignals and waveforms. As this would probably be a mess if I add the changes to this PR, I will close this one and open a new one, with the complete set of changes I made.

PR #108, opened 2017-07-13
Added function to compute the CV2 (coefficient of variation for non-stationary time series)

pietroquaglio:
Added CV2 (coefficient of variation for non-stationary time series Reference: http://jn.physiology.org/content/jn/75/5/1806.full.pdf) to the statistics module


elephant/statistics.py
```diff
@@ -250,6 +250,67 @@ def lv(v):
...
+    # convert to array, cast to float
+    v = np.asarray(v)
+
+    # ensure the input ia a vector
+    if len(v.shape) > 1:
+        raise AttributeError("Input shape is larger than 1. Please provide "
+                             "a vector in input.")
+
+    # ensure we have enough entries
+    if v.size < 2:
```
mdenker: For this function, and for LV, one may ask whether it's better to return np.NaN instead of throwing an error. An error will be fool-proof, but returning a NaN (or None) will often be extremely practical.

pietroquaglio: I agree that it is a possibility to consider. For the moment I choose to return  an error for sake of consistency with other functions in elephant (e.g. statistics.lv()). Probably what to return in case of invalid input needs to be discussed in another more general pool request involving multiple function.

mdenker: Fair enough, then this will be discussed in #115

PR #121, opened 2017-10-19
Calculation of the spike-triggered phase and amplitude

mdenker:
This is a first contribution towards tools to work with measures based on phase synchronization analysis. This function extracts the phase and amplitude of a signal at the times of spikes.

The function could use more unit tests, and there is certainly a lot of room for optimization, but before continuing I would like to know if this function would cover the required basic functionality?

 @JuliaSprenger I think you may have similar code -- could you check in how far this is similar in functionality?


JuliaSprenger:
I this the code covers the basic requirements and the tests work on my side.

The code I used for calculating was partially overlapping with the current STA implementation at that time since collection the individual phases at the spike time is similar to collecting the LFP values at the spike times.

I suggest to adjust the behaviour of the  spike_triggered_phase function to make it more similar to the sta in terms of reordering the arguments. Do you think it makes sense to include a 'window' option in the future? In that case the code would be again partially similar to the STA implementation (lines 150-165).

mdenker:
I think adding a window option is something to discuss -- however, maybe this should be modularized, such that there is a common way in which all such functions handle the cutting. Raised an issue #151.

elephant/phase_analysis.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Calculate spike-triggered phases and amplitudes of the oscillation:
+    >>> phases, amps, times = elephant.phase_analysis.spike_triggered_phase(
+            elephant.signal_processing.hilbert(analogsignal),
+            spiketrain,
+            interpolate=True)
+    """
+
+    # Convert inputs to lists
+    if type(spiketrains) is not list:
```
alperyeg: For type comparision please use `instanceof`

elephant/phase_analysis.py
```diff
@@ -0,0 +1,166 @@
...
+    Example
+    -------
+    Create a 20 Hz oscillatory signal sampled at 1 kHz and a random Poisson
+    spike train:
+
+    >>> f_osc = 20. * pq.Hz
+    >>> f_sampling = 1 * pq.ms
+    >>> tlen = 100 * pq.s
+    >>> time_axis = np.arange(
+            0, tlen.rescale(pq.s).magnitude,
```
alperyeg: In the example you do not need to resacle.

elephant/phase_analysis.py
```diff
@@ -0,0 +1,166 @@
...
+            num_spiketrains != num_phase:
+        raise ValueError(
+            "Number of spike trains and number of phase signals"
+            "must match, or either of the two must be a single signal.")
+
+    # For each trial, select the first input
+    start = [elem.t_start for elem in hilbert_transform]
+    stop = [elem.t_stop for elem in hilbert_transform]
+
+    result_phases = [[] for _ in range(num_spiketrains)]
```
alperyeg: Do you need to create empty lists? Can't you use later on extend or just numpy arrays while intializing?

elephant/phase_analysis.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # Convert outputs to arrays
+    for i, entry in enumerate(result_phases):
+        result_phases[i] = np.array(entry).flatten()
+    for i, entry in enumerate(result_amps):
+        result_amps[i] = pq.Quantity(entry, units=entry[0].units).flatten()
+    for i, entry in enumerate(result_times):
+        result_times[i] = pq.Quantity(entry, units=entry[0].units).flatten()
+
+    return (result_phases, result_amps, result_times)
```
alperyeg: You can remove redundant brackets.

Issue #114, opened 2017-10-07
Unittests do not run through


alperyeg:
Somehow when running the tests of the [ubuntu environment](https://travis-ci.org/NeuralEnsemble/elephant/jobs/284302015#L1179) via travis, the unittests get stuck and do not run to the end. Travis stops at some point the build due to no response. On my laptop I could investigate that the CSD tests take a lot of time but not more than 10 minutes. On my own branch I even put a `travis_wait` command for the tests but that didn't help, too.  
I would appreciate any further information and help on this issue.

alperyeg:
Issue fixed with https://github.com/NeuralEnsemble/elephant/pull/123

PR #123, opened 2017-11-03
hotfixing travis ci build

alperyeg:
This PR provides a hotfix to solve the problems we have with travis ci and is related to https://github.com/NeuralEnsemble/elephant/issues/114

Following points were adjusted in `travis.yml` and `install.sh`: 
* fall-back to `Ubuntu precise` instead of `Ubuntu trusty`, this is the main hotfix
* added explicit python version                          
* removed usage of system site packages               
* removed creation of virtual envs in the Ubuntu build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
+dist: precise
 language: python
 sudo: false
-virtualenv:
-  system_site_packages: true
+
 addons:
    apt:
       packages:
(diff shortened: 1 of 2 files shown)


alperyeg:
In case someone is worried about why I removed the virtual env, travis has an inbuild env. See [here](https://travis-ci.org/NeuralEnsemble/elephant/jobs/296794387#L221).

PR #125, opened 2017-11-15
matrix env for mpi testing

alperyeg:
This PR adds an additional environment to the matrix in `travis.yml` and corresponding changes in `install.sh`. With this env we should be able to test for mpi via `mpi4py`.

### DIFF
--- a/continuous_integration/test_script.sh
+++ b/continuous_integration/test_script.sh
@@ -14,7 +14,7 @@ python -c "import scipy; print('scipy %s' % scipy.__version__)"
 
 if [[ "$COVERAGE" == "true" ]]; then
     if [[ "$MPI" == "true" ]]; then
-	mpiexec -n 2 nosetests --with-coverage --cover-package=elephant
+	mpiexec -n 1 nosetests --with-coverage --cover-package=elephant
     else
 	nosetests --with-coverage --cover-package=elephant
     fi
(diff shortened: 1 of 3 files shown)


continuous_integration/test_script.sh
```diff
@@ -13,7 +13,11 @@ python -c "import numpy; print('numpy %s' % numpy.__version__)"
 python -c "import scipy; print('scipy %s' % scipy.__version__)"
 
 if [[ "$COVERAGE" == "true" ]]; then
-    nosetests --with-coverage --cover-package=elephant
+    if [["$MPI == true" ]]; then
+	mpiexec -n 2 nosetests --with-coverage --cover-package=elephant
```
mdenker: I guess this should be 
`"$MPI" == "true"`
to work...

continuous_integration/test_script.sh
```diff
@@ -13,7 +13,7 @@ python -c "import numpy; print('numpy %s' % numpy.__version__)"
 python -c "import scipy; print('scipy %s' % scipy.__version__)"
 
 if [[ "$COVERAGE" == "true" ]]; then
-    if [["$MPI == "true" ]]; then
+    if [["$MPI" == "true" ]]; then
```
mdenker: Still throwing an error, I think it needs a space after [[:
`[[ "$MPI" ...`

alperyeg: Are you testing locally?

mdenker: No, but travis had errors on that bash line such that mpiexec was not executed. Now it seems it is, and creates "real" errors...

mdenker:
This PR sets up the environment for parallel tests, but individual parallel codes must still be adapted to make use of `mpirun n>1`.

PR #126, opened 2017-12-08
Small fixes to spade 

pietroquaglio:
This pull request is for minor fixes to the spade.py module:

- unnecessary check for the number of surrogate in input when 'patterns' output format is requested
- in pattern set reduction now all possible window position are compared and not only the first occurrence of the pattern


mdenker:
Seems ok to me and @alperyeg

PR #93, opened 2016-07-22
CSD implentation - icsd and kcsd

ccluri:
Here is kernel-CSD implementation for 1D, 2D, and 3D cases of electrode configurations. Also included is MoIKCSD (See Ness et.al, 2015).

The kCSD used here is a mirror of https://github.com/Neuroinflab/kCSD-python/tree/elephant
Further demos on how to use kCSD is illustrated here https://github.com/Neuroinflab/kCSD-python (see tests folder).

PR #127, opened 2018-01-23
Fix *CSD methods/tests

espenhgn:
Fixes some issues related to `neo.AnalogSignal` assumes signal shape `(# timesteps, # nchannels)`

### DIFF
--- a/elephant/current_source_density.py
+++ b/elephant/current_source_density.py
@@ -95,7 +95,7 @@ def estimate_csd(lfp, coords=None, method=None,
     Returns
     -------
     Estimated CSD
-       neo.AnalogSignal Object
+       neo.AnalogSignal object
        annotated with the spatial coordinates
 
     Raises
@@ -235,7 +235,7 @@ def generate_lfp(csd_profile, ele_xx, ele_yy=None, ele_zz=None,
 
         Returns
         -------
-        LFP : list(neo.AnalogSignal type objects)
+        LFP : neo.AnalogSignal object
            The potentials created by the csd profile at the electrode positions
            The electrode postions are attached as RecordingChannel's coordinate
     """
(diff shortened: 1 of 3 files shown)


espenhgn:
Suggest @ccluri take a look.

alperyeg:
@espenhgn the fixes look fine. I can merge if you want. 
@ccluri further any input on this?

ccluri:
@espenhgn : https://github.com/espenhgn/elephant/blob/a7d78489bee4d1e2a4443ade9bb211d43f232f94/elephant/current_source_density.py#L238 minor fix to the documentation.

Otherwise okay with the PR.

Does this break backward compatibility with neo 0.4? 
For future reference, the reason for the inputs to be list(neo.analogsignal) was mentioned here - point number 6 here - https://github.com/NeuralEnsemble/elephant/pull/93#issuecomment-249233213

alperyeg:
> Does this break backward compatibility with neo 0.4?

The `AnalogSignal` was replaced in `v0.5` with the `AnalogSignalArray`. So it depends what you do with the returned output, if it was formatted correctly then it should not make a problem. Regarding the coordinates (your point 6) there is not (probably will not in the future) an ultimate solution to have a consistent mapping. Maybe one could think to have a mapping in the annotations?

espenhgn:
Hi,
I fixed the docstring @ccluri pointed out. 
@alperyeg, your comment confuses me. I'm using neo 0.5.2, and there is no AnalogSignalArray class, only AnalogSignal. 

I'm fine with merging.

alperyeg:
Yes it is confusing. What I meant is the internal representation of the AnalogSignal.

espenhgn:
@alperyeg Ok, I understand.

PR #130, opened 2018-02-15
Fixed bug in spade in calculating max_spikes and max_occ

pietroquaglio:
max_occ and max_spikes were not calculated correctly in case of input parameters equal to None.
Added a test to check this analyzing a compound poisson process.

### DIFF
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -441,13 +441,15 @@ def concepts_mining(data, binsize, winlen, min_spikes=2, min_occ=2,
     # objects (window positions) and attributes (spikes,
     # indexed with a number equal to  neuron idx*winlen+bin idx)
     context, transactions, rel_matrix = _build_context(binary_matrix, winlen)
-    # By default, set the maximum pattern size to the number of spiketrains
+    # By default, set the maximum pattern size to the maximum number of
+    # spikes in a window
     if max_spikes is None:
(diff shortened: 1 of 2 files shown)


elephant/spade.py
```diff
@@ -441,13 +441,15 @@ def concepts_mining(data, binsize, winlen, min_spikes=2, min_occ=2,
...
     if max_spikes is None:
-        max_spikes = np.max(np.sum(binary_matrix, axis=-1)) + 1
-    # By default set maximum number of data to number of bins
+        max_spikes = int(np.max(np.sum(rel_matrix, axis=1)))
+    # By default, set maximum number of occurrences to number of non-empty
+    # windows
     if max_occ is None:
-        max_occ = np.max(np.sum(binary_matrix, axis=1)) + 1
-    # check if fim.so available and use it
+        max_occ = (np.sum(np.sum(rel_matrix, axis=1)>0))
```
alperyeg: I think you forgot an `int(...)` here or `dtype=int`.

pietroquaglio: It should work nevertheless (also because the test works), but you are right, thanks. Should I fix it or you can do it?

alperyeg: I edited it now. Yes, it should work nevertheless. But like this it is more consistent to lines 446-447: 
```python
 max_spikes = int(np.max(np.sum(rel_matrix, axis=1)))
```

PR #131, opened 2018-02-15
Added inhomogeneous_poisson function and relative tests

pietroquaglio:
Added function to generate no-stationary poisson spike trains


pietroquaglio:
Sorry did not realize I added a file which does not belong here (test_non_stat_poiss_nonstat.py)

PR #140, opened 2018-03-02
Fixes to spade for the integration of updated version of fim 

pietroquaglio:
The updated version of fim.so do not accept anymore list of numpy arrays as input and has a different output. This pull request contains the changes to spade necessary to use the updated version of fim.


pietroquaglio:
@alperyeg @mdenker Please take a look to the paragraph added to the module documentation!

PR #143, opened 2018-03-14
Excluded some old code/non-relevant code from coverage.

mdenker:
Some functions in `statistics,py` are deprecated and should not contribute to coverage (will be removed soon).

The file `make_spiketrain_extraction_test_data` is only contained for reference to indicate how the fixed test data for the spike_train_extraction_module was generated originally, and should also not be covered.


alperyeg:
@mdenker I will merge this PR, but can you make an issue to remind us to remove the old parts later on, e.g. for the next release (v0.5).

PR #147, opened 2018-03-19
Fix missing import in example

jens-k:
Example does not run without.


mdenker:
Thanks for the fast fix!

jens-k:
#ANDA2018 ;)

PR #149, opened 2018-03-21
Fixed bug for dimension of the pattern spectrum and set default parameter pvalue_spectrum to None in concept_output_to_patterns

pietroquaglio:
Fixed bug for dimension of the pattern spectrum and set default parameter pvalue_spectrum to None in concept_output_to_patterns corrected spectrum but wrong inizialization of the matrix) and set default parameter pvalue_spectrum to None concept_output_to_patterns  and added related unit test

### DIFF
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -359,8 +359,9 @@ def spade(data, binsize, winlen, min_spikes=2, min_occ=2, min_neu=1,
             if n_surr == 0:
                 pv_spec = []
             # Transfroming concepts to dictionary containing pattern infos
-            output['patterns'] = concept_output_to_patterns(concepts, pv_spec,
+            output['patterns'] = concept_output_to_patterns(concepts,
                                                             winlen, binsize,
+                                                            pv_spec,
                                                             data[0].t_start)
         else:
             output['patterns'] = concepts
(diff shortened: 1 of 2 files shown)


elephant/spade.py
```diff
@@ -1568,12 +1569,13 @@ def concept_output_to_patterns(concepts, pvalue_spectrum, winlen, binsize,
...
             ((spikes in the pattern), (occurrences of the patterns))
-    pvalue_spectrum: tuple
-        Contains a tuple of signatures and the corresponding p-value
     winlen: int
         Length (in bins) of the sliding window used for the analysis
-    t_start: int
-    t_start (in bins) of the analyzed spike trains
+    pvalue_spectrum: None or tuple
+        Contains a tuple of signatures and the corresponding p-value. If equal
+        to None all the pvalues are set to -1
```
alperyeg: In the comment below you write: 
> If an empty list is given in input to the pval spectrum the pvalue is set to -1 (pvalue spectrum not available)

but here you mention that all pvalues are set to -1. It doesn't match with the documentation here. I also see that the dictionary will have afterwards only one entry, `-1`
```python
output_dict['pvalue'] = -1
```

PR #152, opened 2018-04-04
Update test_unitary_event_analysis.py

Vahidrostami:
test added to reproduce the results of Riehle et al. 1997 science using unitary event method (implemented in Elephant) and check the consistency with the original publication (see Rostami et al. 1997 [Re] science for a detailed explanation)

### DIFF
--- a/elephant/test/test_unitary_event_analysis.py
+++ b/elephant/test/test_unitary_event_analysis.py
@@ -338,6 +338,86 @@ class UETestCase(unittest.TestCase):
             UE_dic['indices']['trial26'],expected_indecis_tril26))
         self.assertTrue(np.allclose(
             UE_dic['indices']['trial4'],expected_indecis_tril4))
+        
+    # test if the result of newly implemented Unitary Events in
+    # Elephant is consistent with the result of
+    # Riehle et al 1997 Science
+    # (see Rostami et al (2016) [Re] Science, 3(1):1-17)
+    def test_Riehle_et_al_97_UE(self):
+        import sys
+        import os
+        from neo.rawio.tests.tools import (download_test_file,
+                                           create_local_temp_dir,
+                                           make_all_directories)
+        from neo.test.iotest.tools import (cleanup_test_file)
+        url = [
+            "https://raw.githubusercontent.com/ReScience-Archives/" +
+            "Rostami-Ito-Denker-Gruen-2017/master/data",
+            "https://raw.githubusercontent.com/ReScience-Archives/" +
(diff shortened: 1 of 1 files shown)


Vahidrostami:
Hi Michael,
matplotlib is imported in utils.py which is one of the file that we download from Rescience repo. I can not change anything in the utils.py file.  Could you think of any workaround? 
One solution would be to write the function that we import from utils.py in the UE test file and remove importing utils.py.

mdenker:
Hmm, interesting problem. Given we cannot fix this in the published code, I'd suggest to do a mock import of matplotlib according to somehing I found here:
https://stackoverflow.com/questions/8658043/how-to-mock-an-import

In our case, before importing `utils.py`, maybe try the following:
```
import sys
import mock
sys.modules['matplotlib']=mock.Mock()
sys.modules['matplotlib.pyplot']=mock.Mock()
...
import utils
```
Not really beautiful, but maybe this works?

Vahidrostami:
@mdenker Please check if the problem is solved

mdenker:
I forgot, the mock library is required for this fix for python 2.0. I'll enter it in the requirements and see if that works...

mdenker:
@Vahidrostami Ah, actually I cannot. Could you please enter
`mock==2.0.0`
as last line in the requirements.txt file in the main directory and make a new commit?

Vahidrostami:
@mdenker done! #154

mdenker:
Ah, sorry, a bit more difficult, I forgot there are other parts where this would go, however...

... I just started thinking about his again. In principle I like the idea of using the original code to load the data, in terms of reproducibility and so on. However, it's not unlikely that this code in `utils.py` wil break at some point, e.g., with python 4.0. Thus, maybe the better option is actually to stop this mock business and go with your other suggestion and just copy the required code to the unit test. This would allow future maintainers to adapt the loading routine if it breaks.

mdenker:
Maybe the problem is that the tests throw a UnicodeError instead of a UnicodeDecodeError for numpy 1.10 (failed tests 1 and 5),
and a nose.proxy.UnicodeDecodeError for ascii for failing test 3. For the latter, maybe we just have to live with the fact, that this test will not execute in numpy 1.8 and lower, unless there is a way to load the data with pickle instead of npload..?

alperyeg:
Probably copy paste error why I wrote `UnicodeEncodeError`, should be `UnicodeError`. Corrected now.

alperyeg:
This problems seems to be caused by the underlying pickle process. I assume the files are saved before using Python 2 and numpy. So in the environment with Python 3 and numpy v1.8 this is bound to fail.

PR #154, opened 2018-04-05
Update requirements.txt

Vahidrostami:
new requirement: mock library which is used in unitary event analysis unit test


mdenker:
Temporarily closed, see #152

PR #157, opened 2018-05-04
added support for binary input to BinnedSpikeTrain 

alperyeg:
This PR adds the functionality to accept a binary input (i.e. array or list with 0's and 1's) and converts it to a `BinnedSpikeTrain` object. The functionality of the `BinnedSpikeTrain` class does not change.

### DIFF
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -349,8 +349,11 @@ class BinnedSpikeTrain(object):
 
     Parameters
     ----------
-    spiketrains : List of `neo.SpikeTrain` or a `neo.SpikeTrain` object
-        Spiketrain(s)to be binned.
+    spiketrains : List of `neo.SpikeTrain`, a `neo.SpikeTrain` object
+        or a binary array or list.
+        Spiketrain(s)to be binned. 
+        Accepts also a binary representation and converts it to a 
+        `BinnedSpikeTrain` object.
     binsize : quantities.Quantity
         Width of each time bin.
         Default is `None`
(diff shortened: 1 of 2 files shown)


alperyeg:
TODO: 
* Add functionality to accept also non-binary input.  
* Define `self.is_binary` as class property 
* Add to the documentation that `self.is_binary` is not indicating a "clipped" matrix

alperyeg:
@mdenker now `BinnedSpikeTrain` class can accept a matrix (not necessarily binary) and convert it to a `BinnedSpikeTrain` object. Now it is possible to input to use `to_array()` on a `BinnedSpikeTrain` object, do some operations on it and e.g. feed it back to a `BinnedSpikeTrain` object. 
Additionally, there is now a property to check for binary input (return True or False), which could be useful in certain situations.

elephant/conversion.py
```diff
@@ -812,3 +844,43 @@ def _convert_to_binned(self, spiketrains):
...
+
+def _check_not_neo_spiketrain(matrix):
+    """
+    Checks if given matrix does not contain neo spiketrain objects
+
+    """
+    # Check for lists
+    if isinstance(matrix, list):
+        # Check for proper dimension
+        # arr = np.array(matrix)
```
mdenker: Can this code vanish?

alperyeg:
@mdenker I accidentally submitted my fixes for the unitary event analysis, which cause the errors for this PR. I wanted to suggest those fixes in a separate pull request but now they are here. I know it is not a nice style. 
If you think I should make a separate pull request I will do so and remove the two lines from this PR, breaking the tests again.

alperyeg:
In the context of PR https://github.com/NeuralEnsemble/elephant/pull/171 we were thinking that it makes sense to output a warning in the checks of the `BinnedSpikeTrain` if empty spiketrains were detected. Would that be a useful addition?

alperyeg:
@mdenker I fixed the conflicts, if possible please merge.

Issue #158, opened 2018-05-08
Deduplicate requirement information


healther:
Currently requirements are specified both directly in `setup.py` as well as in `requirements.txt`.

Would it be possible to parse `requirements.txt` with something like (would require removal of commented out lines in `requirements.txt`)
```
with open('requirements.txt') as fp:
    install_requires = fp.read()

setup(install_requires=install_requires, ...)
```
rather than have two different locations for the same information?

healther:
cc @muffgaga @schmitts

muffgaga:
BTW. the software specified in the travis job seems to be slightly different (optional dependency on `pandas` version) from `requirements.txt`.

alperyeg:
@healther  @muffgaga, I completely agree. I can do it, otherwise feel free to contribute. Of course, we are happy for any other suggestions for improving the setup and ci routines (see e.g. #129).

rgerkin:
@healther @muffgaga @alperyeg To make sure this works in all cases, consider looking at my implementation in SciUnit [here](https://github.com/scidash/sciunit/blob/master/setup.py).  This handles pip <=9 vs pip 10, and parses some edge cases where a valid line in `requirements.txt` is not a valid `install_requires` list item.  It also supports `--process-dependency-links` if we ever need that.  Lastly, don't forget to add `requirements.txt` to a new `MANIFEST.in` file, otherwise `requirements.txt` will not get included in the pypi release and setup will fail because it won't be able to find it.

healther:
I'll try to get you a PR sometime next week

alperyeg:
> Would it be possible to package the `fim.so`s directly with the elephant sources
I would not like to change that. The `fim.so` is directly downloaded from Christian Borgelts HP (the developer and maintainer of the frequent itemset algorithm we use) and we do not want to ship all the files. Besides, there is a fall-back version in  pure but slow python if the files cannot be accessed.

healther:
The problem is that right now the following will happen if you don't have an internet connection when doing `python setup.py install`

```
$ python setup.py install
Traceback (most recent call last):
  File "/Users/healther/.pyenv/versions/3.5.2/lib/python3.5/urllib/request.py", line 1254, in do_open
    h.request(req.get_method(), req.selector, req.data, headers)
  File "/Users/healther/.pyenv/versions/3.5.2/lib/python3.5/http/client.py", line 1106, in request
    self._send_request(method, url, body, headers)
[...]

PR #159, opened 2018-05-14
deduplicate requirements

healther:
PR for #158

tested on a python 3.6.3 install on OSX.

I didn't use @rgerkin 's implementation as it would add a `pip` requirement. All this PR does is: 
- remove the hardcoded `requires` lists from `setup.py` 
- parse `requirements{-*}.txt` and populate the `*_requires` from it


setup.py
```diff
@@ -9,15 +9,12 @@
...
-                    'six>=1.10.0']
-extras_require = {'pandas': ['pandas>=0.14.1'],
-                  'docs': ['numpydoc>=0.5',
-                           'sphinx>=1.2.2'],
-                  'tests': ['nose>=1.3.3']}
+with open('requirements.txt', 'r') as fp:
+    install_requires = fp.read()
+extras_require = {}
+for extra in ['pandas', 'docs', 'tests']:
+    with open('requirements-{0}.txt'.format(extra), 'r') as fp:
```
alperyeg: To understand, wouldn't this require 3 additional txt files?

healther: yes and they are there. In case there would be multiple extras that would depend on the same package this would again lead to a duplication.

alperyeg: Ah, I see it now. 
Actually, I am against this style to have each time a  `requirements-*.txt` for a new optional package. I would keep the extra requirements as they are atm. They are not essential to have extra requirements (in the sense they are not required :)) .

healther: The problem with that approach (and the reason why I stumbled onto this) is that we use [spack](github.com/spack/spack) as our package manager and I needed to declare the dependencies in order to prevent `setuptools` from manually downloading the dependencies. Since the dependencies that we declared where from `requirements.txt` but the actually used ones where from `setup.py` the build crashed (the build node is isolated from the internet).

Is there a reason why you are against using multiple files? The two alternatives which I see  would be to either ONLY put dependencies in `setup.py` and remove the `requirements.txt` or use pip's parsing (adding a dependency on pip) of a fully fledged `requirements.txt`.
I find annoying the former annoying, because `setup.py` is executed as a python script meaning, that the lists `install_requires` etc. can be modified at multiple places and that makes finding out which dependencies a package actually has error prone.

alperyeg: Regarding spack I only know it from Eric, that's why I am asking. When building the package with spack are the `extra_requirements installed? As far as I know the extras are only installed if specified. 

Ok, I think as far as we restrict it to 3 additional requirement files it shouldn't be too cluttered. Can you rename `requirements-pandas.txt` to `requirements-extras.txt` and accordingly in the `setup.py`, because we may add more modules to it in future.

alperyeg: Sounds like a lot of effort and pain. Remembering Eric's face it is! 

We do not require `pandas` that's why it is a loose dependency. For those packages we have the `extra_require`assignment. Some of the more advanced methods such as `Asset` or `Spade` require those packages, but for the overall library they are uninteresting, so to speak. 

Why do you think `requirements-extras.txt` is not intuitive? It is like "extra packages" (the variable in the setup file is even called like that; `extras_require`), for a Python developer not so unknown I would assume. The idea is to put there other packages, which are not necessarily needed. I would like to keep it not too cluttered, meaning not to have for each additional package a requirements file. 

Besides, from the user point of view this is rather unimportant, since `pip install elephant` does it's job for the most people (never saw or know someone who did `pip install elephant[tests]`). A developer or package maintainer will anyway clone the repository and have a look at the code. 
So if I get you, when packing for spack it is easier to have thousands of requirement files with single package names instead thousand of lines with package names in a single file (this is slightly exaggerated :)).

healther: > Why do you think requirements-extras.txt is not intuitive? 

Because I don't know how to use it :D I guess that to get the dependencies you would have to do `pip install elephant[extras]`? Which should in practise be the same as not declaring the dependency at all (i.e. users would do a `pip install elephant && pip install pandas` anyways)

> So if I get you, when packing for spack it is easier to have thousands of requirement files with single package names instead thousand of lines with package names in a single file (this is slightly exaggerated :)).

No for spack it's simpler to have one `requirements.txt` (one place to look at), but parsing that in `setup.py` in a sensible way requires `pip` or no case selectivity. But the most easy packages to work with (that I have found so far) were those that had multiple ones (for ease of parsing). Having only one `requirements.txt` typically means a) the authors didn't distinguish dependencies at all, b) `requirements.txt` is not actually used or c) it is a pip-installed package. So for `spack` it really doesn't matter either way, but my experience tells me that most packages are of type b), and finding dependencies of these packages is a pain (for me^^ and because people are doing ... interesting... stuff in their `setup.py`'s)

From my side it doesn't really matter anymore (as the `package.py` now exists and I know the dependencies), this is really more a quality of life improvement for the next guy (or furture-me). But realistically the only people running into this will be people contributing to package managers

alperyeg: I agree, thanks for the explanations.  
To shorten things, can you make the extra requirements and include in there pandas and scikit-learn? 
In sum, there will be `requirements.txt`, `requirements-docs.txt` `requirements-test.txt` and `requirements-extra`. 
If this happens to not go well we can still change it and go the other way or around.

healther: no problem, done

alperyeg:
Shall I merge, or do you still want to add something? E.g. in the corresponding issue we were discussing about the `fim.so` file. How did you solve this for `spack`?

healther:
From my side this is good to go.
> How did you solve this for spack?

I didn't solve it at all, I only noticed it when looking at the `setup.py` (and didn't have the time to actually look up what happens on an air gapped cluster)

PR #162, opened 2018-05-18
changed ubuntu env build

alperyeg:
Since the `setup.py` now is incorporating the `requirements.txt` directly, I changed one of our CI builds, so that the packages are directly installed following the `requirements.txt`.


mdenker:
Looks good to me, much clearer. Minimal numpy and scipy versions are still covered by the codamin environments.

PR #164, opened 2018-07-05
Vectorize max

halvarsu:
The Python function max is slow for neo and quantities objects, see [here](https://gist.github.com/halvarsu/87502edf8842b129e38cfc21d0016f76)


alperyeg:
Hi @halvarsu thank you for the contribution and the insightful comparison! Will merge.

PR #170, opened 2018-08-09
Cross-correlation function for analog signals

mvonpapen:
Added a function "cross_correlation_function" in signal_processing.py that calculates the pairwise cross-correlation of selected channels within a neo.AnalogSignal. Optionally calculates the envelope of the cross-correlation function, which is helpful to determine the correlation length of an oscillatory signal.

The two signals, e.g. LFP, are zero padded to minimize boundary effects, transformed into Fourier domain, multiplied and transformed back into time domain. The signals are afterwards corrected (unbiased estimator) for zero-padding.

This function should not be confused with cross_corr_coeff option in cross_correlation_histogram function within spiketrain_correlation.py. The latter only works for CCH obtained from spiketrains and assumes Poissonian statistics (Equation (5.10) in "Analysis of parallel spike trains", 2010, Gruen & Rotter, Vol 7).

The input is 
- signal: neo.AnalogSignal
- channel1/2: int, list, tuple, array; this determines the indices of the channels; cross-correlation will be calculated for pairs channel1[0] vs. channel2[0], channel1[1] vs channel2[1], and so forth
- dt: sample resolution
- env: bool; envelope yes or no
- nlags: int; how many lags the output rho should have

The output is:
- rho: array, the cross-correation function (ranging between -1 and 1) OR env, the envelope of rho
- tau: array, vector of lags

### DIFF
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -221,7 +221,7 @@ def cross_correlation_function(signal, ch_pairs, dt=1., env=False, nlags=None):
     # multiply in Fourier space, and transform back. Correct for bias due 
     # to zero-padding
     xcorr = np.zeros((Nt, Nch))
-    for i in xrange(Nch):
+    for i in range(Nch):
         xcorr[:,i] = scipy.signal.fftconvolve(x[:,i], y[::-1,i], mode='same')
     bias = npm.repmat((Nt-abs(tau/dt)), Nch, 1).T
     xcorr = xcorr / bias
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -142,59 +142,68 @@ def cross_correlation_function(signal, ch_pairs, env=False, nlags=None):
     """
     Computes unbiased estimator of the cross-correlation function.
 
-    Calculates the unbiased estimator of the cross-correlation function
-    R(tau) = 1/(N-|k|) R'(tau), where R'(tau) = E[x(t)*y(t+tau)] in a
-    pairwise manner, i.e. signal[ch_pairs[0,0]] vs signal2[ch_pairs[0,1]],
-    signal[ch_pairs[1,0]] vs signal2[ch_pairs[1,1]], and so on (see also
-    Hall & River (2009) Spectral Analysis of Signals, Spectral Element Method
-    in Structural Dynamics, Eq. 2.2.3). The cross-correlation function is
-    obtained by scipy's fftconvolve. Time series in signal are zscored
-    beforehand. Alternatively returns the Hilbert envelope of R(tau), which is
-    useful to determine the correlation length of oscillatory signals.
+    Calculates the unbiased estimator of the cross-correlation function [1]_
+    
+    .. math::
+             R(\\tau) = \\frac{1}{N-|k|} R'(\\tau) \\ ,
+    
+    where :math:`R'(\\tau) = \\left<x(t)y(t+\\tau)\\right>` in a pairwise 
+    manner, i.e. `signal[ch_pairs[0,0]]` vs `signal2[ch_pairs[0,1]]`,
+    `signal[ch_pairs[1,0]]` vs `signal2[ch_pairs[1,1]]`, and so on. The
+    cross-correlation function is obtained by `scipy.signal.fftconvolve`.
+    Time series in signal are zscored beforehand. Alternatively returns the
+    Hilbert envelope of :math:`R(\\tau)`, which is useful to determine the 
+    correlation length of oscillatory signals.
 
     Parameters
     -----------
-    signal : neo.AnalogSignal (nt x nch)
+    signal : neo.AnalogSignal (`nt` x `nch`)
         Signal with nt number of samples that contains nch LFP channels
-    ch_pairs : list (or array with shape (n,2))
+    ch_pairs : list (or array with shape `(n,2)`)
         list with n channel pairs for which to compute cross-correlation,
         each element of list must contain 2 channel indices
-    env: bool
+    env : bool
         Return Hilbert envelope of cross-correlation function
         Default: False
-    nlags: int
+    nlags : int
         Defines number of lags for cross-correlation function. Float will be
-        rounded to nearest integer. Number of samples of output is 2*nlags+1.
+        rounded to nearest integer. Number of samples of output is `2*nlags+1`.
         If None, number of samples of output is equal to number of samples of
-        input signal, namely nt
+        input signal, namely `nt`
         Default: None
 
     Returns
     -------
-    cross_corr : neo.AnalogSgnal (2*nlag+1 x n)
+    cross_corr : neo.AnalogSgnal (`2*nlag+1` x `n`)
         Pairwise cross-correlation functions for channel pairs given by
-        `ch_pairs`. If env=True, the output is the Hilbert envelope of the
+        `ch_pairs`. If `env=True`, the output is the Hilbert envelope of the
         pairwise cross-correlation function. This is helpful to compute the
         correlation length for oscillating cross-correlation functions
 
-    Example:
-        dt = 0.02
-        N = 2018
-        f = 0.5
-        t = np.arange(N)*dt
-        x = np.zeros((N,2))
-        x[:,0] = 0.2 * np.sin(2.*np.pi*f*t)
-        x[:,1] = 5.3 * np.cos(2.*np.pi*f*t)
-        # Generate neo.AnalogSignals from x
-        signal = neo.AnalogSignal(x, units='mV', t_start=0.*pq.ms,
-            sampling_rate=1/dt*pq.Hz, dtype=float)
-        rho = elephant.signal_processing.cross_correlation_function(
-            signal, [0,1], nlags=150)
-        env = elephant.signal_processing.cross_correlation_function(
-            signal, [0,1], nlags=150, env=True)
-        plt.plot(rho.times, rho)
-        plt.plot(env.times, env) # should be equal to one
-        plt.show()
+    Examples
+    --------
+        >>> dt = 0.02
+        >>> N = 2018
+        >>> f = 0.5
+        >>> t = np.arange(N)*dt
+        >>> x = np.zeros((N,2))
+        >>> x[:,0] = 0.2 * np.sin(2.*np.pi*f*t)
+        >>> x[:,1] = 5.3 * np.cos(2.*np.pi*f*t)
+        >>> # Generate neo.AnalogSignals from x
+        >>> signal = neo.AnalogSignal(x, units='mV', t_start=0.*pq.ms,
+        >>>     sampling_rate=1/dt*pq.Hz, dtype=float)
+        >>> rho = elephant.signal_processing.cross_correlation_function(
+        >>>     signal, [0,1], nlags=150)
+        >>> env = elephant.signal_processing.cross_correlation_function(
+        >>>     signal, [0,1], nlags=150, env=True)
+        >>> plt.plot(rho.times, rho)
+        >>> plt.plot(env.times, env) # should be equal to one
+        >>> plt.show()
+
+    References
+    ----------
+    .. [1] Hall & River (2009) "Spectral Analysis of Signals, Spectral Element
+       Method in Structural Dynamics", Eq. 2.2.3
     """
 
     # Make ch_pairs a 2D array
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -192,7 +192,7 @@ def cross_correlation_function(signal, ch_pairs, env=False, nlags=None):
                 signal, [0,1], nlags=150)
         env = elephant.signal_processing.cross_correlation_function(
                 signal, [0,1], nlags=150, env=True)
-        plt.plot(rho.tmes, rho)
+        plt.plot(rho.times, rho)
         plt.plot(env.times, env) # should be equal to one
         plt.show()
     """
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -12,7 +12,6 @@ import numpy as np
 import scipy.signal
 import quantities as pq
 import neo
-import numpy.matlib as npm
 
 
 def zscore(signal, inplace=True):
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -180,6 +180,17 @@ def cross_correlation_function(signal, ch_pairs, env=False, nlags=None):
         pairwise cross-correlation function. This is helpful to compute the
         correlation length for oscillating cross-correlation functions
 
+    Raises
+    ------
+    ValueError
+        If the input signal is not a neo.AnalogSignal.
+    ValueError
+        If `ch_pairs` is not a list of channel pair indices with shape `(n,2)`.
+    KeyError
+        If keyword `env` is not a boolean.
+    KeyError
+        If `nlags` is not an integer or float larger than 0.
+
     Examples
     --------
         >>> dt = 0.02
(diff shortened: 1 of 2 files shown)


elephant/signal_processing.py
```diff
@@ -134,6 +135,111 @@ def zscore(signal, inplace=True):
...
+    dt : float
+        Defines sampling period
+    env: bool
+        Return Hilbert envelope of cross-correlation function
+    nlags: int
+        Defines number of lags for cross-corelation function. Floats will be 
+        rounded to nearest integer. Output is of length 2*nlags+1. If None, 
+        len(tau) = Nt
+
+    Returns
```
JuliaSprenger: The description of the return values seems to be incomplete. How about going more into details about the envelope / rho calculation. Also mention tau here.

elephant/signal_processing.py
```diff
@@ -134,6 +135,111 @@ def zscore(signal, inplace=True):
...
+    determine the correlation length of oscillatory signals.
+
+    Parameters
+    -----------
+    signal : neo.AnalogSignal
+        Signal that contains the LFP channels
+    ch_pairs : list (or array with shape (N,2))
+        list with N channel pairs for which to compute cross-correlation, 
+        each element of list must contain 2 channel indices
+    dt : float
```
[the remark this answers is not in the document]
JuliaSprenger: And if you use ```dt``` as a sampling time, then it should be a quantity and not a float.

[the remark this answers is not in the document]
JuliaSprenger: ```dt``` should be always larger or equal to the sampling period of the signals.

elephant/signal_processing.py
```diff
@@ -134,6 +135,111 @@ def zscore(signal, inplace=True):
...
+def cross_correlation_function(signal, ch_pairs, dt=1., env=False, nlags=None):
+
+    """
+    Computes unbiased estimator of the cross-correlation function.
+    
+    Calculates the unbiased estimator of the cross-correlation function 
+    R(tau) = 1/(N-|k|) R'(tau), where R'(tau) = E[x(t)*y(t+tau)] for signal1 
+    and signal2 in a pairwise manner, i.e. signal1[0] vs signal2[0], signal1[1] 
+    vs signal2[1] and so on. The cross-correlation function is obtained 
+    by np.correlate. signal1 and signal2 are zscored beforehand.
```
JuliaSprenger: You seem not to use ```np.correlate```

elephant/signal_processing.py
```diff
@@ -134,6 +135,111 @@ def zscore(signal, inplace=True):
...
+        Signal that contains the LFP channels
+    ch_pairs : list (or array with shape (N,2))
+        list with N channel pairs for which to compute cross-correlation, 
+        each element of list must contain 2 channel indices
+    dt : float
+        Defines sampling period
+    env: bool
+        Return Hilbert envelope of cross-correlation function
+    nlags: int
+        Defines number of lags for cross-corelation function. Floats will be
```
JuliaSprenger: ```Floats``` should read ```Float```

elephant/signal_processing.py
```diff
@@ -134,6 +135,111 @@ def zscore(signal, inplace=True):
...
+        list with N channel pairs for which to compute cross-correlation, 
+        each element of list must contain 2 channel indices
+    dt : float
+        Defines sampling period
+    env: bool
+        Return Hilbert envelope of cross-correlation function
+    nlags: int
+        Defines number of lags for cross-corelation function. Floats will be 
+        rounded to nearest integer. Output is of length 2*nlags+1. If None, 
+        len(tau) = Nt
```
JuliaSprenger: What is ```Nt```? The number of samples in the signals?

elephant/test/test_signal_processing.py
```diff
@@ -15,12 +15,48 @@
...
+        for f in freq:
+            x[:,0] = 0.2 * np.sin(2.*np.pi*f*t+phi)
+            x[:,1] = 10.7 * np.sin(2.*np.pi*f*t+phi*2.)
+            x[:,2] = 5.3 * np.cos(2.*np.pi*f*t)
+            x[:,3] = 1875.35 * np.cos(2.*np.pi*f*t)
+            # Generate neo.AnalogSignals from x and y
+            signal = neo.AnalogSignal(x, units='mV', t_start=0.*pq.ms, 
+                                      sampling_rate=1/dt*pq.Hz, dtype=float)
+            rho, tau = elephant.signal_processing.cross_correlation_function(
+                    signal, [[0, 2], [1, 3]], dt=dt, nlags=N/4.6)
```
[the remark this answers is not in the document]
mvonpapen: No, it doesnt. I just wanted to use some float smaller than the number of input samples. Do you think it would be better to use a fixed float, e.g. 31.3? Or what would you suggest?

elephant/signal_processing.py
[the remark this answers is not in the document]
mvonpapen: Good point. I will extract dt from the signal then.

elephant/signal_processing.py
```diff
@@ -134,6 +135,111 @@ def zscore(signal, inplace=True):
...
         return result
+    
+    
+def cross_correlation_function(signal, ch_pairs, dt=1., env=False, nlags=None):
+
+    """
+    Computes unbiased estimator of the cross-correlation function.
+    
+    Calculates the unbiased estimator of the cross-correlation function 
+    R(tau) = 1/(N-|k|) R'(tau), where R'(tau) = E[x(t)*y(t+tau)] for signal1
```
JuliaSprenger: Can you add a reference for this normalization? Also please add a description of the variables you are using.

JuliaSprenger:
This function is accompanied by a single, quite complex test. We suggest to add multiple, more simple and stupid tests, eg like testing that ```tau``` is independent of the ```env``` keyword.

elephant/test/test_signal_processing.py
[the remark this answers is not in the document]
JuliaSprenger: It just looks like it has some deeper meaning if you write it like this. So you should either add a comment, telling that 4.6 is just a random choice, or use a fixed float, which you can also define together with all the other parameters above.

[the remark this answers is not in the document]
mvonpapen: I will define nlag = 30 in the beginning. I only use nteger in order not to complicate things

elephant/signal_processing.py
mvonpapen: description has been changed accordingly

elephant/signal_processing.py
mvonpapen: Nt is now clearly defined

elephant/signal_processing.py
mvonpapen: Will add: Hall & River (2009) Spectral Analysis of Signals, Spectral Element Method in Structural Dynamics, Eq. 2.2.3

mvonpapen:
> This function is accompanied by a single, quite complex test. We suggest to add multiple, more simple and stupid tests, eg like testing that `tau` is independent of the `env` keyword.

I added separate tests for env, nlags, different frequencies and phase shift

alperyeg:
@JuliaSprenger let me know when to merge. Thank you for the review.

elephant/signal_processing.py
```diff
@@ -134,6 +135,124 @@ def zscore(signal, inplace=True):
...
+    signal : neo.AnalogSignal (Nt x Nch)
+        Signal with Nt samples that contains Nch LFP channels
+    ch_pairs : list (or array with shape (N,2))
+        list with N channel pairs for which to compute cross-correlation, 
+        each element of list must contain 2 channel indices
+    env: bool
+        Return Hilbert envelope of cross-correlation function
+        Default: False
+    nlags: int
+        Defines number of lags for cross-corelation function. Float will be
```
JuliaSprenger: 'corelation' -> 'correlation'

elephant/signal_processing.py
```diff
@@ -134,6 +135,124 @@ def zscore(signal, inplace=True):
...
+        Signal with Nt samples that contains Nch LFP channels
+    ch_pairs : list (or array with shape (N,2))
+        list with N channel pairs for which to compute cross-correlation, 
+        each element of list must contain 2 channel indices
+    env: bool
+        Return Hilbert envelope of cross-correlation function
+        Default: False
+    nlags: int
+        Defines number of lags for cross-corelation function. Float will be 
+        rounded to nearest integer. Sample length of output is 2*nlags+1.
```
JuliaSprenger: 'Sample length'? maybe better 'Number of samples' or similar?

elephant/signal_processing.py
```diff
@@ -134,6 +135,124 @@ def zscore(signal, inplace=True):
...
+        list with N channel pairs for which to compute cross-correlation, 
+        each element of list must contain 2 channel indices
+    env: bool
+        Return Hilbert envelope of cross-correlation function
+        Default: False
+    nlags: int
+        Defines number of lags for cross-corelation function. Float will be 
+        rounded to nearest integer. Sample length of output is 2*nlags+1. 
+        If None, sample length of output is equal to number of samples in 
+        input signal, namely Nt
```
JuliaSprenger: Nt.

elephant/signal_processing.py
```diff
@@ -134,6 +135,124 @@ def zscore(signal, inplace=True):
...
+        rho = elephant.signal_processing.cross_correlation_function(
+                signal, [0,1], nlags=150)
+        env = elephant.signal_processing.cross_correlation_function(
+                signal, [0,1], nlags=150, env=True)
+        plt.plot(rho.times, rho)
+        plt.plot(env.times, env) # should be equal to one
+        plt.show()
+    """
+    
+    # Make ch_pairs an 2D array
```
JuliaSprenger: 'an' -> 'a'

JuliaSprenger:
Thanks for the improvements! The code is already more readable!
I would suggest to add some more checks on the input parameters. Right now the error messages for messed up inputs are not really telling. For example when providing a negative value for `nlags` or a string as `env` parameter.

mvonpapen:
Hi Julia,

Thanks very much for the thorouh review. I added the suggested checks for the input (keyword) arguments and corrected for pep8. I additionally changed some variable names to comply with snake_case naming style.

While doing pep8 checks, I noticed that the refractor test complained about too many local variables. I therefore also reduced the number of these where possible (e.g. variables a, b, and bias).

JuliaSprenger:
Ready to merge from my side! Thanks for fixing all the details!

elephant/signal_processing.py
```diff
@@ -136,6 +137,129 @@ def zscore(signal, inplace=True):
         return result
 
 
+def cross_correlation_function(signal, ch_pairs, env=False, nlags=None):
+
+    """
+    Computes unbiased estimator of the cross-correlation function.
+
+    Calculates the unbiased estimator of the cross-correlation function
+    R(tau) = 1/(N-|k|) R'(tau), where R'(tau) = E[x(t)*y(t+tau)] in a
```
alperyeg: You could use `.. math::` for the mathematical formulations to have them in a nice format on the docs page. 
```suggestion
    .. math:: R(tau) = frac{1}/{(N-|k|)} R^{\prime}(tau), \text{where} R^{\prime}(tau) = E[x(t)*y(t+tau)] 
```
Please have a look at section 12 of the [numpy docs](https://numpydoc.readthedocs.io/en/latest/format.html) for possibilities.

elephant/signal_processing.py
```diff
@@ -136,6 +137,129 @@ def zscore(signal, inplace=True):
...
+    pairwise manner, i.e. signal[ch_pairs[0,0]] vs signal2[ch_pairs[0,1]],
```
alperyeg: Encapsulate variables or code snippets etc with  ` `` `
```suggestion
`signal[ch_pairs[0,0]]` vs `signal2[ch_pairs[0,1]]`
```

elephant/signal_processing.py
```diff
@@ -136,6 +137,129 @@ def zscore(signal, inplace=True):
...
+    signal[ch_pairs[1,0]] vs signal2[ch_pairs[1,1]], and so on (see also
+    Hall & River (2009) Spectral Analysis of Signals, Spectral Element Method
```
alperyeg: I would put this on the references section (References) and only reference it here.

elephant/signal_processing.py
```diff
@@ -136,6 +137,129 @@ def zscore(signal, inplace=True):
...
+
+    Returns
+    -------
+    cross_corr : neo.AnalogSgnal (2*nlag+1 x n)
+        Pairwise cross-correlation functions for channel pairs given by
+        `ch_pairs`. If env=True, the output is the Hilbert envelope of the
+        pairwise cross-correlation function. This is helpful to compute the
+        correlation length for oscillating cross-correlation functions
+
+    Example:
```
alperyeg: This is not the correct format for the example section, please have a look e.g. [here](https://github.com/NeuralEnsemble/elephant/blob/master/elephant/cell_assembly_detection.py#L189)

elephant/signal_processing.py
```diff
@@ -136,6 +137,129 @@ def zscore(signal, inplace=True):
...
+        plt.show()
+    """
+
+    # Make ch_pairs a 2D array
+    pairs = np.array(ch_pairs)
+    if pairs.ndim == 1:
+        pairs = pairs[:, np.newaxis]
+
+    # Check input
+    if not isinstance(signal, neo.AnalogSignal):
```
alperyeg: Do you have a raises section (Raises) in the documentation, you should mention there the `ValueError` and `KeyError`.

elephant/signal_processing.py
```diff
@@ -136,6 +137,129 @@ def zscore(signal, inplace=True):
...
+
+    # Cut off lags outside desired range
+    if nlags is not None:
+        nlags = int(np.round(nlags))
+        tau0 = int(np.argwhere(tau == 0))
+        xcorr = xcorr[tau0-nlags:tau0+nlags+1, :]
+
+    # Return neo.AnalogSignal
+    cross_corr = neo.AnalogSignal(xcorr,
+                                  units='',
```
alperyeg: Why do you give an empty string for the `units`

elephant/signal_processing.py
mvonpapen: Good idea and I tried it. Unfortunately, the latest Spyder version 3.2.6 cannot render this and the output becomes shitty (see below). I would therefore suggest not to write it like this.

--- docstring output ---

Calculates the unbiased estimator of the cross-correlation function [1]

.. math:: R(        au) = rac{1}{N-|k|} R'(        au) \ ,

ight>` in a pairwise manner,eft<x(t)y(t+    au)
i.e. signal[ch_pairs[0,0]] vs signal2[ch_pairs[0,1]], signal[ch_pairs[1,0]]
vs signal2[ch_pairs[1,1]], and so on. The cross-correlation function is
obtained by scipy.signal.fftconvolve. Time series in signal are zscored
beforehand. Alternatively returns the Hilbert envelope of :math:`R( au)`,
which is useful to determine the correlation length of oscillatory signals.

[1] Hall & River (2009) "Spectral Analysis of Signals, Spectral Element
Method in Structural Dynamics", Eq. 2.2.3

elephant/signal_processing.py
mvonpapen: correlation is unitless

elephant/signal_processing.py
mvonpapen: added raises section

elephant/signal_processing.py
alperyeg: Right, didn't even now that this is possible in `neo`.

PR #168, opened 2018-08-01
small fix in asset for future numpy versions

alperyeg:
This PR is a fix for upcoming numpy versions, which does not support the `-` operation as used so far. The operation is replaced with numpy's`logical_xor` function.

PR #173, opened 2018-09-13
updating scripts to be compatible with latest numpy versions

alperyeg:
PR to fix breaking travis script:
* updated to latest numpy, scipy and pandas versions in the corresponding env build
* updated several scripts which broke because of the new libraries (incorporating PR #168)
* updated accompanying tests 
* pep8 in those scripts

### DIFF
--- a/elephant/asset.py
+++ b/elephant/asset.py
@@ -41,7 +41,6 @@ References:
 [1] Torre, Canova, Denker, Gerstein, Helias, Gruen (submitted)
 """
 
-
 import numpy as np
 import scipy.spatial
 import scipy.stats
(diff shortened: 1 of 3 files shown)


elephant/asset.py
```diff
@@ -78,7 +78,7 @@ def _signals_same_tstart(signals):
     signals : list
         a list of signals (e.g. AnalogSignals or SpikeTrains) having
         attribute `t_start`
-
+en
```
Junji110: What is this?

alperyeg: It's a typo. If you look at my second commit it should have been removed. Please squash when merging.

Junji110: Ah, upper commit is the later one, that I didn't notice.
OK, I have no more comments now.

Junji110:
Sorry, I pressed the revert button by mistake.
How can I cancel this?

alperyeg:
I just checked out the master branch the commits are applied. I think we do not need to take action.

Junji110:
The master branch is OK, but the revert created a new branch named "revert-173-fix/travis_numpy" in this repository.
It shouldn't exist.
Can you delete this branch?

Issue #174, opened 2018-09-14
BinnedSpikeTrain returns incorrect bin edges and bin centers when arguments have different units


pietromarchesi:
Running neo 0.6.1, quantities 0.12.2.

```python
import numpy as np
import neo
import elephant
import quantities as pq

train = neo.SpikeTrain(times=np.array([1.001, 1.002, 1.005])*pq.s,
                       t_start=1*pq.s, t_stop=1.01*pq.s)

bs = elephant.conversion.BinnedSpikeTrain(train,
                                     t_start=1 * pq.s, t_stop=1.01 * pq.s,
                                     binsize=1 * pq.ms)

print(bs.bin_edges)
print(bs.bin_centers)
```
Returns
```
[1.    1.001 1.002 1.003 1.004 1.005 1.006 1.007 1.008 1.009 1.01 ] ms
[1.5   1.501 1.502 1.503 1.504 1.505 1.506 1.507 1.508 1.509] ms
```
The correct units for `bin_edges` are seconds (instead of milliseconds), whereas the `bin_centers` seem pretty off. This bug has to do with the case in which `BinnedSpikeTrain` receives different units, because converting manually

```python
bs = elephant.conversion.BinnedSpikeTrain(train,
                                     t_start=1000 * pq.ms, t_stop=1010 * pq.ms,
                                     binsize=1 * pq.ms)

print(bs.bin_edges)
print(bs.bin_centers)
```
Gives the expected output
```
[1000. 1001. 1002. 1003. 1004. 1005. 1006. 1007. 1008. 1009. 1010.] ms
[1000.5 1001.5 1002.5 1003.5 1004.5 1005.5 1006.5 1007.5 1008.5 1009.5] ms
```

alperyeg:
@pietromarchesi thanks for the bug report I will have a look.

PR #175, opened 2018-09-14
Fixed correct return unit for bin edges and centers

alperyeg:
In response to #174. `BinnedSpikeTrain` didn't check if the unit of `binsize` is corresponding to e.g `t_start`. Also added regression test.


elephant/conversion.py
```diff
@@ -564,6 +564,8 @@ def _check_consistency(self, spiketrains, binsize, num_bins, t_start,
             raise TypeError(
                 "Number of bins (num_bins) is not an integer or < 0: " + str(
                     num_bins))
+        elif binsize.units != self.t_start.units:
```
rgutzen: These two lines only handle the inconsistency of t_start.units and binsize.units and disregards the case where t_start.units != t_stop.units.

To fix this, these two lines can be deleted and instead the calculation in the bin_edges() function should be changed. See below

rgutzen: Change lines 586-589 to:

       return pq.Quantity(np.linspace(self.t_start.rescale('s').magnitude,
                                       self.t_stop.rescale('s').magnitude,
                                       self.num_bins + 1, endpoint=True), units='s').rescale(self.binsize.units)

elephant/test/test_conversion.py
```diff
@@ -500,6 +500,22 @@ def test_binned_spiketrain_different_units(self):
...
+        target_edges = np.array([1., 1.001, 1.002, 1.003, 1.004, 1.005, 1.006,
+                                 1.007, 1.008, 1.009, 1.01], dtype=np.float)
+        target_centers = np.array(
+            [1.0005, 1.0015, 1.0025, 1.0035, 1.0045, 1.0055, 1.0065, 1.0075,
+             1.0085, 1.0095], dtype=np.float)
+        self.assertTrue(np.allclose(bst.bin_edges.magnitude, target_edges))
+        self.assertTrue(np.allclose(bst.bin_centers.magnitude, target_centers))
+        self.assertTrue(bst.bin_centers.units == pq.s)
+        self.assertTrue(bst.bin_edges.units == pq.s)
+
```
rgutzen: add second test with different t_start.unit and t_stop.unit

elephant/conversion.py
alperyeg: @rgutzen thanks for the comments. Wouldn't it make sense to have the same units for `start`, `stop` and `binsize`? Otherwise only the `bin_edges` property would be rescaled to seconds. I coulnd't find any side-effect so far, but they could appear with different units.

rgutzen: @alperyeg the idea was that the units for each variable remain the same as entered by the user. Rescaling them might lead to confusion (e.g. if only `.magnitude` is used).
The `bin_edges` then take on the units of `binsize`, only within the calculation in between the variables are all rescaled to seconds.

mdenker:
Fixes #174

Issue #115, opened 2017-10-09
lv() and cv2() raising an error for too short spike trains 


pietroquaglio:
Currently the functions lv() and cv2() in statistics raise an error if the spike trains in input has less three spikes. It might be preferable to replace this with a warning message and returning NaN or None instead (suggested in #108 ). For example this avoids to break for loops over list of spike trains.

mdenker:
I think this is a very general design decision to make: The safe solution is to error, the often more practical solution is to `nan`. Would it make sense to have functions such as the ones mentioned here error on default, but optionally only warn and return nan's?

apdavison:
I agree with Michael's suggestion - have error as default, but add an optional argument to return NaN, so that the user has to be explicitly aware of the choice. (This means the error message needs to include information about the NaN option)

alperyeg:
@EmanueleLucrezia can you fix this issue? Thanks!

EmanueleLucrezia:
I can try to adapt the code from here: https://github.com/NeuralEnsemble/elephant/pull/171/files, where the same issue arose for empty spike trains and an argument 'with_nan' was introduced

PR #179, opened 2018-10-09
fixed 115

EmanueleLucrezia:
This pull request fixes issue https://github.com/NeuralEnsemble/elephant/issues/115.
An option `withNaN` was added to the function lv and cv2: if set True, then the function will returns NaN in the case the spike train is too short.


elephant/statistics.py
```diff
@@ -225,7 +232,8 @@ def lv(v):
     AttributeError :
        If an empty list is specified, or if the sequence has less
        than two entries, an AttributeError will be raised.
-    ValueError :
+    ValueError : # should it be attribute?
```
alperyeg: Can you remove the remove the comment `# should it be attribute?`

elephant/statistics.py
```diff
@@ -239,18 +247,31 @@ def lv(v):
     """
     # convert to array, cast to float
     v = np.asarray(v)
+    
+    # ensure the input ia a vector
+    if len(v.shape) > 1: # adeed by me for symmetry with CV
```
alperyeg: Please remove the comment

elephant/statistics.py
```diff
@@ -239,18 +247,31 @@ def lv(v):
...
+                             "a vector as an input.")
 
     # ensure we have enough entries
     if v.size < 2:
-        raise AttributeError("Input size is too small. Please provide "
-                             "an input with more than 1 entry.")
+        if with_nan:
+           warnings.warn("Input size is too small. Please provide "
+                         "an input with more than 1 entry. lv returns 'NaN'"
+                         "since the argument `with_nan` is True")
```
alperyeg: Can you put around all the `True` and `False` keywords  " \` \`". This ensures a better readability.

alperyeg:
Can you also fix some of the pep8 issues?

elephant/statistics.py
```diff
@@ -215,6 +215,13 @@ def lv(v):
     v : quantity array, numpy array or list
         Vector of consecutive time intervals
 
+    with_nans : bool, optional
+        If `True`, correlations of empty spike trains are given NaN values. If
+        `False`, a boolean array indicating empty (`False`) and non-empty (`True`)
```
alperyeg: The boolean mask is not returned as far as I can see. Instead `np.NaN` is returned.

Issue #177, opened 2018-09-27
test_spike_train_generation.py uses the PyNNIO class to load test data 


pietroquaglio:
In `test_spike_train_generation.py` in order to test the spike_extraction function, some reference artificial data, generated with `make_spike_extraction_test_data.py` and stored in a .npz file as a neo block, are loaded using the PyNNIO class which does not exist any longer. This makes fail the test.

alperyeg:
@rgerkin did you provide the file and data? We have a working solution for it see [here](https://github.com/INM-6/elephant/commit/c037f6c163d3e4171969caa648e05e1d1c693713), using instead the `.npz` a `.txt` file and loading the analogsignal from there. Would you be fine with that solution? Because atm. we do not know any other way to load the data without the `PyNNIO`.

alperyeg:
@pietroquaglio I think you can submit the PR. Then we can continue the discussion over there.

alperyeg:
Fixed by #180

PR #180, opened 2018-10-11
Fixed bug in test_spiketrain_generation.py due to the removal of pynumpyio from neo

pietroquaglio:
Replaced .npz file containing test data with .txt (not storing an entire neo.block, but a=only the signal as a list). See issue  #177.


alperyeg:
This doesn't work because the pickle format is not supported for python 2.7. What about making an acii text file? Or alternatively using nix.

PR #184, opened 2018-10-19
Sorting of modules names

alperyeg:
Motivation: When I want to read the documentation of a specific function I look at the [module list](https://elephant.readthedocs.io/en/latest/modules.html) on the read the docs page to go from there to the function. Unfortunately the list is not ordered and this confuses me. 

I sorted the module list `modules.rst` but this does not lead to the desired outcome because we have [specific descriptions ](https://github.com/NeuralEnsemble/elephant/blob/master/doc/reference/conversion.rst)for the modules. 

This PR proposes a way how to deal with the sorting. By changing to a format `module name - description` it should be easier to automatically sort and find things faster with a growing list of modules. 
An other option would be categorizing the modules under specific topics.


alperyeg:
I changed the `conf.py` to suppress the wrong warnings when building the documentation. For more details read [this](https://github.com/phn/pytpm/issues/3#issuecomment-12133978).

mdenker:
Definitely an improvement, but let's keep categorization as an issue.

PR #187, opened 2018-11-06
Bugfixes in STTC Calculation

Kleinjohann:
Fixed potential unit error by enforcing simplification. This ensures that the quantity T in run_T() in spike_train_correlation.spike_time_tiling_coefficient() is dimensionless when we strip the units. PyQuantities does not automatically simplify combinations of ms and s even if the result is dimensionless as in this case. This problem emerged if the spike times of the input spike-trains and the synchronicity window dt did not have the same units.

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -690,7 +690,16 @@ def spike_time_tiling_coefficient(spiketrain_1, spiketrain_2, dt=0.005 * pq.s):
         PA = PA / N1
         PB = run_P(spiketrain_2, spiketrain_1, N2, N1, dt)
         PB = PB / N2
-        index = 0.5 * (PA - TB) / (1 - PA * TB) + 0.5 * (PB - TA) / (
+        # check if the P and T values are 1 to avoid division by zero
+        if PA * TB == 1:
+            if PB * TA == 1:
+                index = 1.
+            else :
+                index = 0.5 + 0.5 * (PB - TA) / (1 - PB * TA)
+        elif PB * TA == 1:
+            index = 0.5 + 0.5 * (PA - TB) / (1 - PA * TB)
+        else:
+            index = 0.5 * (PA - TB) / (1 - PA * TB) + 0.5 * (PB - TA) / (
             1 - PB * TA)
     return index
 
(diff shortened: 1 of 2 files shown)


elephant/test/test_spike_train_correlation.py
```diff
@@ -579,7 +579,7 @@ def setUp(self):
 
     def test_sttc(self):
         # test for result
-        target = 0.8748350567
+        target = 0.495860165593
```
alperyeg: Why did the target change?

elephant/spike_train_correlation.py
```diff
@@ -690,8 +690,17 @@ def run_T(spiketrain, N, dt):
         PA = PA / N1
         PB = run_P(spiketrain_2, spiketrain_1, N2, N1, dt)
         PB = PB / N2
-        index = 0.5 * (PA - TB) / (1 - PA * TB) + 0.5 * (PB - TA) / (
-            1 - PB * TA)
+        # check if the P and T values are 1 to avoid division by zero
+        if PA * TB == 1:
```
alperyeg: These cases should be tested, too.

elephant/spike_train_correlation.py
```diff
@@ -690,8 +690,17 @@ def run_T(spiketrain, N, dt):
...
+            if PB * TA == 1:
+                index = 1.
```
alperyeg: Setting `index == 1` should be documented. Also why 1?

elephant/test/test_spike_train_correlation.py
Kleinjohann: The previous target was wrong due to the unit error in the calculation. I ran the original C code which this function is based on; it yields 0.4958602 which is consistent with my result. [Here](https://onlinegdb.com/HkIK9PeTX) is the ready-to-run C code with the same input parameters in an online compiler.

elephant/spike_train_correlation.py
Kleinjohann: I added an explanation in the documentation and in the comment next to the if clauses. In those cases, every spike in the train with P = 1 is within dt of a spike in the other train, so I set the respective (partial) index to 1.

elephant/test/test_spike_train_correlation.py
alperyeg: I can confirm this after checking the units.

elephant/spike_train_correlation.py
alperyeg: Is it possible to add a test for the 0/0 case? I guess you had such a case?

alperyeg:
There are minor pep8 issues, can you take care of it?

Kleinjohann:
> There are minor pep8 issues, can you take care of it?

I fixed the issues, it seems like the remaining extra result is a false positive deprecation warning caused by "\sqrt{...}" in a docstring, but I'm not 100% sure.

PR #189, opened 2018-11-26
Change signal's `duplicate_with_new_array` to `duplicate_with_new_data`

jpgill86:
For compatibility with Neo 0.7.0 (soon to be released)

### DIFF
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -118,7 +118,7 @@ def zscore(signal, inplace=True):
         # Create new signal instance
         result = []
         for sig in signal:
-            sig_dimless = sig.duplicate_with_new_array(
+            sig_dimless = sig.duplicate_with_new_data(
                 (sig.magnitude - m.magnitude) / s.magnitude) / sig.units
             result.append(sig_dimless)
     else:
@@ -376,7 +376,7 @@ def butter(signal, highpass_freq=None, lowpass_freq=None, order=4,
         )
 
     if isinstance(signal, neo.AnalogSignal):
-        return signal.duplicate_with_new_array(np.rollaxis(filtered_data, -1, 0))
+        return signal.duplicate_with_new_data(np.rollaxis(filtered_data, -1, 0))
     elif isinstance(signal, pq.quantity.Quantity):
         return filtered_data * signal.units
     else:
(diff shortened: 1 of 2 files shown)


jpgill86:
Unless a more clever approach is taken in which the method called is determined dynamically depending on which version of Neo is installed, `requirements.txt` will need bumped too. I can try that after Neo 0.7.0 is officially released to see if the Travis tests pass.

jpgill86:
Actually, I think there will be many more changes needed if elephant is going to support Neo 0.7.0 fully, which changed a lot more than the name of this one method.

mdenker:
Cool, thanks for the quick response to this breaking issue. I can confirm your fix passes the corresponding tests on my machine with the upcoming Neo 0.7 code.

As you expect, there are a few other issues that arise, in particular an issue in the instantaneous rate, spike train correlation, as well as in the `neo_utils` module. It seems at first glance that at least some of these issues have to do with changes in the way `t_start` and `t_stop` are handled when rescaling a spike train. Previously, `t_start` and `t_stop` where scaled along with the spike times, in 0.7. this behavior seems changed and the start and stop are left alone (which may be rather the desired behavior). Requires more investigation, though...

mdenker:
The `t_start` and `t_stop` issue breaking some of Elephant's tests is now handled in 
https://github.com/NeuralEnsemble/python-neo/issues/613
https://github.com/NeuralEnsemble/python-neo/pull/614

jpgill86:
To handle the breaking change described here (Neo's `AnalogSignal.duplicate_with_new_array` was renamed to `AnalogSignal.duplicate_with_new_data` in 0.7.0), we could try to preserve backwards compatibility with older versions of Neo by dynamically using one name or the other depending on which version is installed. Is it worth it?

Since Neo 0.7.0 is live on PyPI, I'm going to, for now, try the simpler thing in this PR and just bump the Neo requirement to 0.7.0, and then we can see how the Travis tests fare.

mdenker:
Yes, the version number must be fixed. I think this is reasonable, since Neo and Elephant are coupled so strongly, that "hot fixes" will not be a good idea at this stage. 

However, we considered to make matching releases of the next major versions, i.e., fixing Elephant to use
```
neo>=0.7.0,<0.8.0
```
such that the future Neo releases up to Neo 1.0.0 (which are likely to have considerable enhancements) will not break Elephant.

jpgill86:
Alright, that sounds reasonable. I'm going to remove the change to `requirements.txt` from this PR since that seems to be part of a larger discussion and requires the coordination of multiple fixes.

mdenker:
Ok, yes, that makes sense. Will merge this fix as is.

mdenker:
This PR was merged and extended into cumulativ fix #203 to test CI pass.

mdenker:
Neo Version number of upcoming 0.6.1 will be fixed to >=0.7.0 and <0.8.0 to prevent similar breaks on 0.8.0 release.

Issue #198, opened 2019-01-31
change `.base` to `.magnitude`


JuliaSprenger:
This is related to #197.
`.base` is a method of `numpy` arrays and therefore is not aware of quantities. When using `.base` instead of `.magnitude` this might lead to unexpected behaviour if units can not be considered.

Issue #197, opened 2019-01-30
spike_dither() sometimes returns an empty array


PaulinaDabrowska:
If the units of SpikeTrain and its t_start and t_stop differ, spike_dither() may return an empty array (but sometimes seems to work correctly). An example to reproduce the error: 
```
import numpy as np
import quantities as pq
import elephant.spike_train_generation as elgen
import elephant.spike_train_surrogates as elsurr

dither = 15 * pq.ms
t_start_slice = 0 * pq.s
t_stop_slice = 1 * pq.s

# generate spike trains
st = elgen.homogeneous_poisson_process(36 * pq.Hz, t_start=0*pq.ms, t_stop=30000*pq.ms)
st_slice = st.time_slice(t_start_slice, t_stop_slice)
# generate surrogates
surr_st = elsurr.dither_spikes(st, dither)[0]
surr_slice = elsurr.dither_spikes(st_slice, dither)[0]
# check
print('Original spike trains: {} and {} spikes.'.format(len(st), len(st_slice)))
print('Surrogates: {} and {} spikes'.format(len(surr_st), len(surr_slice)))
```

PR #199, opened 2019-02-06
Fix/197 198

pietroquaglio:
This pull request should fix both issue #197 and #198. Additionally in the test now it is covered the case of t_stop with different unit than the spikes.


elephant/spike_train_surrogates.py
```diff
@@ -111,23 +111,23 @@ def dither_spikes(spiketrain, dither, n=1, decimals=None, edges=True):
     # Main: generate the surrogates
     surr = data.reshape((1, len(data))) + 2 * dither * np.random.random_sample(
         (n, len(data))) - dither
-
+    print(surr)
```
PaulinaDabrowska: There is a print left.

elephant/spike_train_surrogates.py
```diff
@@ -111,23 +111,23 @@ def dither_spikes(spiketrain, dither, n=1, decimals=None, edges=True):
...
 
     if edges is False:
         # Move all spikes outside [spiketrain.t_start, spiketrain.t_stop] to
         # the range's ends
-        surr = np.minimum(np.maximum(surr.base,
-            (spiketrain.t_start / spiketrain.units).base),
-            (spiketrain.t_stop / spiketrain.units).base) * spiketrain.units
+        surr = np.minimum(np.maximum(surr.simplified.magnitude,
+            spiketrain.t_start.simplified.magnitude),
+            spiketrain.t_stop.simplified.magnitude) * pq.s
```
PaulinaDabrowska: This solution looks clean and relatively compact, but it forces all numerical operations to be held in seconds. What about applying the unit of original spike train from the beginning: 
```
surr = np.minimum(np.maximum(surr.magnitude,
    (spiketrain.t_start.rescale(spiketrain.units).magnitude)),
    (spiketrain.t_stop.rescale(spiketrain.units).magnitude)) * spiketrain.units
```
Something has to be rescaled, if spiketrain and its t_start and t_stop have different units, but hard-coding any arbitrary units always seems a bit worrisome.

elephant/spike_train_surrogates.py
```diff
@@ -347,15 +347,15 @@ def dither_spike_train(spiketrain, shift, n=1, decimals=None, edges=True):
...
     else:
         # Leave out all spikes outside [spiketrain.t_start, spiketrain.t_stop]
-        tstart, tstop = (spiketrain.t_start / spiketrain.units).base,\
-                        (spiketrain.t_stop / spiketrain.units).base
-        surr = [s[np.all([s >= tstart, s < tstop], axis=0)] * spiketrain.units
-                for s in surr.base]
+        tstart, tstop = spiketrain.t_start.simplified.magnitude, \
+                        spiketrain.t_stop.simplified.magnitude
+        surr = [np.sort(s[np.all([s >= tstart, s < tstop], axis=0)]) * pq.s
```
PaulinaDabrowska: I would not add np.sort() here, since the method is only supposed to shift the original spike train, no matter how nicely constructed it was in the first place. So it should never change the order of spikes - this could affect some analyses, couldn't it?

elephant/spike_train_surrogates.py
pietroquaglio: Thanks for spotting it I will remove it

mdenker:
Hi @pietroquaglio 
the breaking unit tests in the master branch should now be fixed, so we can integrate this PR into Elephant. 

To do this, it would be great if the unit tests would pass on this PR.

Would you be able to base the PR off the current Elephant master branch, by
* pulling the current NeuralEnsemble Elephant master branch to your local master
* changing to you local copy of this feature branch, and making a `git rebase master`
* and then pushing the feature branch to update this PR?

Thanks!

elephant/spike_train_surrogates.py
pietroquaglio: The output spiketrains are in the end all re-scaled to the input for the units, honestly for the computations I prefer to have explicitely fixed units everywhere, which are not arbitrary but they match the simplified function

elephant/spike_train_surrogates.py
pietroquaglio: Sorting is necessary because the it is not automatically checked by the neo.Spiketrains class, but other functions (e.g. isi()) assumes the spike are in increasing order.

pietroquaglio:
Going through the process I created a small conflict but now it should be fine!

elephant/spike_train_surrogates.py
mdenker: Actually, a `sort()` for ISIs was already included in the meantime.

mdenker:
@PaulinaDabrowska and I discussed the outstanding issues again, and concluded to stick with the hard-coded seconds in the analysis until an agreement is reached on a more global level (e.g., by defining a global "standard time unit"). Also, the sorting is best left in for now. Therefore, we see no issues left in this PR.

PR #211, opened 2019-04-11
Corrected improper use of .base on quantities objects

mdenker:
This is a follow up of #198 in the `spike_train_generation` module.


mdenker:
Cross-checked by @PaulinaDabrowska

PR #213, opened 2019-04-11
Fix setup.py not installing the fast FIM module for Windows machines

### LINKED ISSUE #212: SPADE: setup.py should download fim.pyd for windows
[`spade.py`](https://github.com/NeuralEnsemble/elephant/blob/009e80f4945e2c12249c6f43f9f6b76fb7d0f520/elephant/spade.py#L76-L83) fails to import `fim` module on windows as [`setup.py`](https://github.com/NeuralEnsemble/elephant/blob/009e80f4945e2c12249c6f43f9f6b76fb7d0f520/setup.py#L23-L36) downloads only the `.so` version of the library, while windows requires the `.pyd` variation. 

The windows precompiled `pyd` files are available in the same  [location](http://www.borgelt.net/pyfim.html) as the `.so`

mdenker:
This fixes an issue where the setup script did not identify if it is running on Windows, so that it should download the pyd module for frequent itemset mining instead of the .so module. Fixes #212 .

### DIFF
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -509,8 +509,9 @@ def concepts_mining(data, binsize, winlen, min_spikes=2, min_occ=2,
     else:
         warnings.warn(
             'Optimized C implementation of FCA (fim.so/fim.pyd) not found ' +
-            'in elephant/spade_src folder,' +
-            'you are using the Python implementation of fast fca.')
+            'in elephant/spade_src folder, or not compatible with this ' +
+            'Python version. You are using the pure Python implementation ' +
+            'of fast fca.')
         # Return output
         mining_results = _fast_fca(
             context,
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -78,9 +78,6 @@ try:
     HAVE_FIM = True
 except ImportError:  # pragma: no cover
     HAVE_FIM = False
-    warnings.warn(
-        'fim.so not found in elephant/spade_src folder,' +
-        'you are using the python implementation of fast fca')
 
 
 def spade(data, binsize, winlen, min_spikes=2, min_occ=2, max_spikes=None,
@@ -510,6 +507,10 @@ def concepts_mining(data, binsize, winlen, min_spikes=2, min_occ=2,
         return mining_results, rel_matrix
     # Otherwise use fast_fca python implementation
     else:
+        warnings.warn(
+            'Optimized C implementation of FCA (fim.so/fim.pyd) not found ' +
+            'in elephant/spade_src folder,' +
+            'you are using the Python implementation of fast fca.')
         # Return output
         mining_results = _fast_fca(
             context,
--- a/setup.py
+++ b/setup.py
@@ -20,20 +20,28 @@ for extra in ['extras', 'docs', 'tests']:
 is_64bit = sys.maxsize > 2 ** 32
 is_python3 = float(sys.version[0:3]) > 2.7
 
-if is_python3:
-    if is_64bit:
-        urlretrieve('http://www.borgelt.net/bin64/py3/fim.so',
-                    'elephant/spade_src/fim.so')
-    else:
-        urlretrieve('http://www.borgelt.net/bin32/py3/fim.so',
-                    'elephant/spade_src/fim.so')
+if uname()[0] == "Windows":
+    oext = ".pyd"
+elif uname()[0] == "Linux":
+    oext = ".so"
 else:
-    if is_64bit:
-        urlretrieve('http://www.borgelt.net/bin64/py2/fim.so',
-                    'elephant/spade_src/fim.so')
+    oext = None
+
+if oext:
+    if is_python3:
+        if is_64bit:
+            urlretrieve('http://www.borgelt.net/bin64/py3/fim' + oext,
+                        'elephant/spade_src/fim' + oext)
+        else:
+            urlretrieve('http://www.borgelt.net/bin32/py3/fim' + oext,
+                        'elephant/spade_src/fim' + oext)
     else:
-        urlretrieve('http://www.borgelt.net/bin32/py2/fim.so',
-                    'elephant/spade_src/fim.so')
+        if is_64bit:
+            urlretrieve('http://www.borgelt.net/bin64/py2/fim' + oext,
+                        'elephant/spade_src/fim' + oext)
+        else:
+            urlretrieve('http://www.borgelt.net/bin32/py2/fim' + oext,
+                        'elephant/spade_src/fim' + oext)
 
 setup(
     name="elephant",
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 from setuptools import setup
 import os
 import sys
+import platform
 try:
     from urllib.request import urlretrieve
 except ImportError:
@@ -19,10 +20,9 @@ for extra in ['extras', 'docs', 'tests']:
 # spade specific
 is_64bit = sys.maxsize > 2 ** 32
 is_python3 = float(sys.version[0:3]) > 2.7
-
-if uname()[0] == "Windows":
+if platform.uname()[0] == "Windows":
     oext = ".pyd"
-elif uname()[0] == "Linux":
+elif platform.uname()[0] == "Linux":
     oext = ".so"
 else:
     oext = None
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,8 @@ setup(
         os.path.join('current_source_density_src', '*.py'),
         os.path.join('spade_src', '*.py'),
         os.path.join('spade_src', 'LICENSE'),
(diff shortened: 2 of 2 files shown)


mdenker:
@oraby This fix works on my windows virtual machine to download fim.pyd, and also loads the module when importing spade (i.e., when doing `import elephant.spade`, no warning is issued and `elephant.spade.HAVE_FIM == True`). Does it work for you when getting this branch and issuing a `python setup.py install` in an empty test environment?

jpgill86:
Your fix is not working for me on my Windows system (64-bit, Python 3, Anaconda). The `fim.pyd` file is downloaded and placed in `spade_src` during setup, but the warning is still issued, and trying to import it directly raises an `ImportError`:

```
In [1]: import elephant.spade
d:\users\jeffrey\documents\python-dev\elephant\git-repo\elephant\elephant\spade.py:83: UserWarning: fim.so not found in elephant/spade_src folder,you are using the python implementation of fast fca
  'you are using the python implementation of fast fca')

In [2]: from elephant.spade_src import fim
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-2-e9fd9d472c47> in <module>
----> 1 from elephant.spade_src import fim

ImportError: DLL load failed: The specified module could not be found.
```

I thought maybe that below [this line](https://github.com/NeuralEnsemble/elephant/blob/0c450705ae0e0e4144cf3c04b1997a198a55460a/setup.py#L57) the following might be needed:

`os.path.join('spade_src', '*.pyd'),`

However, that did not help.

jpgill86:
I noticed that on [this page](http://www.borgelt.net/pyfim.html), the Windows PYD file for Python 3 is listed as working with Python 3.5.x specifically. I'm running Python 3.6.7. I'm not sure if that makes a difference in this case.

mdenker:
Ups, yes, I forgot to include that line you mentioned in the PR when moving the setup.py between Windows and linux, thanks. Still don't understand why it's not working for you. But maybe you are right, I I tested it under a python 2.x environment under windows, I can test on a python 3 windows install later...

jpgill86:
I'm getting the same error in a fresh Python 3.5.5 environment.

mdenker:
I can confirm, it does not work in a Python 3.5 environment either (although the pyd file is copied to `spade_src`), but it works with Python 2.7.13.

mdenker:
...and also not in Python 3.6. It seems it may have to do with Python 3 simply not finding the module as such.

mdenker:
Contrary to that suspicion, confirmed that with Linux/Python3.5 (miniconda) it's still working, such that the `.so` file is found.

Junji110:
I've checked setup.py on Windows 32bit + miniconda Python 3.7.1, and it worked.

In [1]: import elephant.spade as spade
In [2]: dir(spade)
Out[2]:
['HAVE_FIM', 'HAVE_MPI', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_build_context', '_closure_probability_extensional', '_closure_probability_intensional', '_fast_fca', '_fca_filter', '_fdr', '_fpgrowth', '_fpgrowth_filter', '_give_random_idx', '_holm_bonferroni', '_pattern_spectrum_filter', '_stability_filter', 'approximate_stability', 'chain', 'combinations', 'concept_output_to_patterns', 'concepts_mining', 'conv', 'fast_fca', 'fim', 'neo', 'np', 'numpy', 'pattern_set_reduction', 'pq', 'pvalue_spectrum', 'spade', 'surr','test_signature_significance', 'time', 'warnings']
In [3]: from elephant.spade_src import fim
In [4]: dir(fim)
Out[4]:
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'accretion', 'apriacc', 'apriori', 'arules', 'carpenter', 'eclat', 'estpsp', 'fim', 'fpgrowth', 'genpsp', 'ista', 'patred', 'psp2bdr', 'relim', 'sam']

jpgill86:
I'm having the same success as @Junji110 in Windows 64-bit with Python 3.7.

This code creates a fresh conda environment, installs @mdenker's patches and succeeds in printing out the `fim` functions:

```
conda create -n fimtest -y python=3.7
conda activate fimtest
pip install git+https://github.com/INM-6/elephant.git@fix/fimwin
python -c "import elephant; from elephant.spade_src import fim; print(dir(fim))"
```

It also works with `python=2.7`. However, I get the ImportError with `python=3.6` or `python=3.5`. (I also tried 3.4 and encountered numpy/scipy build errors that I didn't want to deal with.)

The source site for downloading the `fim.pyd` file describes the Windows binary as as "Windows Python 3.5.x dynamic module". I wonder if it was actually built recently for Python 3.7 and isn't backwards compatible.

I discovered that conda-forge has builds of PyFim available here:

https://anaconda.org/conda-forge/pyfim/files

Among these are builds for Python 3.7, 3.6, and 2.7 on Windows. After running the code above with `python=3.6` and getting a failure, I manually downloaded conda-forge's build for Python 3.6, replaced the `fim.pyd` downloaded by elephant with it, and voilà, it works.

jpgill86:
If others are able to confirm what I described in my previous post, then it seems to me that changes proposed in this PR technically fix #212 and could be merged.

The remaining issue is that the pre-compiled library for Windows, `fim.pyd`, is apparently compatible with Python 3.7 and 2.7 but is known to be incompatible with Python 3.6 and 3.5. There are a several things that could be done about this, including contacting [the library's host site](http://www.borgelt.net/pyfim.html)'s owner to request a fixed build, or just washing our hands of it and moving on with our lives.

I don't use this feature of elephant and am new to the project, so I can't judge what's best. If fim were a typical optional dependency, I would expect it to be listed as an "extra" requirement, like pandas and scikit-learn in [`requirements-extras.txt`](https://github.com/NeuralEnsemble/elephant/blob/master/requirements-extras.txt). You could let users acquire copies of the library themselves via [conda-forge](https://anaconda.org/conda-forge/pyfim/files) or [PyPI](https://pypi.org/project/fim/#files) and fall back on the slower implementation included with elephant if they don't have it. Instead of downloading a copy during setup to the `spade_src` directory and then trying to import from there like [this](https://github.com/NeuralEnsemble/elephant/blob/ba3303b0f65cb2feac58c0d3420c0b5b31acfb11/elephant/spade.py#L77), a typical approach would be simply this:

```
try:
    import fim  # from anywhere the user has acquired it
    HAVE_FIM = True
except ImportError:
    HAVE_FIM = False
```

I'm not sure why it was decided for this project to go to the extra effort of downloading the binaries from a particular source during setup. Perhaps it's just that important, and building from the source available on PyPI is too difficult?

A major drawback of doing it the way it's done now is that a conda-forge user could install a working binary for Python 3.6 or 3.5 themselves, but elephant would never use it. It would instead only look for fim in `spade_src`.

Regardless of what approach is taken, I have one more recommendation: Move [this annoying warning](https://github.com/NeuralEnsemble/elephant/blob/ba3303b0f65cb2feac58c0d3420c0b5b31acfb11/elephant/spade.py#L81), which displays every time elephant is imported, to [here](https://github.com/NeuralEnsemble/elephant/blob/ba3303b0f65cb2feac58c0d3420c0b5b31acfb11/elephant/spade.py#L512) within `concepts_mining` so that it is only displayed when the fallback implementation is used.

oraby:
> @oraby This fix works on my windows virtual machine to download fim.pyd, and also loads the module when importing spade (i.e., when doing import elephant.spade, no warning is issued and elephant.spade.HAVE_FIM == True). Does it work for you when getting this branch and issuing a python setup.py install in an empty test environment?

It works for me, but again I'm using 3.7, so I have the same results as reported above

mdenker:
The main reason for having the fim library is that it provides a significant speed-up of the Frequent Itemset Mining algorithm over the fall-back solution implemented in pure Python. It is also the library provided by one of the co-authors of the original method. I think the choice for downloading the package from the author's webpage was to avoid having to compile the code during setup, which is difficult under windows (@alperyeg is this correct, or am I missing something?). However, I like the suggestion of having this library available via conda, that would be a much cleaner solution. I will discuss this with the author, and I openend an issue #214 on this topic.

Regarding, the second suggestion, I agree the warning can wait until the function is actually used.

mdenker:
Given this PR seems to fix the original issue as much as possible (i.e., with the python versions available), I will merge this for now, and leave the issue of having fim as real dependency for the  0.7 release.

alperyeg:
@mdenker  Yes, the compilation for fim using the `setup.py` didn't work out on Windows and Mac last time I tried it. Because we didn't experience any problems with the precompiled `.so` files for Linux we only included the the `.so` file in the install script.

mdenker:
@alperyeg Ok, thanks for the quick feedback.

Issue #216, opened 2019-04-17
BUILDSTDERR: urllib.error.URLError: <urlopen error [Errno -2] Name or service not known>


lbazan:
BUILDSTDERR: Traceback (most recent call last):
BUILDSTDERR:   File "/usr/lib64/python3.7/urllib/request.py", line 1317, in do_open
BUILDSTDERR:     encode_chunked=req.has_header('Transfer-encoding'))
BUILDSTDERR:   File "/usr/lib64/python3.7/http/client.py", line 1229, in request
BUILDSTDERR:     self._send_request(method, url, body, headers, encode_chunked)
BUILDSTDERR:   File "/usr/lib64/python3.7/http/client.py", line 1275, in _send_request
BUILDSTDERR:     self.endheaders(body, encode_chunked=encode_chunked)
BUILDSTDERR:   File "/usr/lib64/python3.7/http/client.py", line 1224, in endheaders
BUILDSTDERR:     self._send_output(message_body, encode_chunked=encode_chunked)
BUILDSTDERR:   File "/usr/lib64/python3.7/http/client.py", line 1016, in _send_output
BUILDSTDERR:     self.send(msg)
BUILDSTDERR:   File "/usr/lib64/python3.7/http/client.py", line 956, in send
BUILDSTDERR:     self.connect()
BUILDSTDERR:   File "/usr/lib64/python3.7/http/client.py", line 928, in connect
BUILDSTDERR:     (self.host,self.port), self.timeout, self.source_address)
BUILDSTDERR:   File "/usr/lib64/python3.7/socket.py", line 707, in create_connection
BUILDSTDERR:     for res in getaddrinfo(host, port, 0, SOCK_STREAM):
BUILDSTDERR:   File "/usr/lib64/python3.7/socket.py", line 748, in getaddrinfo
BUILDSTDERR:     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
BUILDSTDERR: socket.gaierror: [Errno -2] Name or service not known
BUILDSTDERR: During handling of the above exception, another exception occurred:
BUILDSTDERR: Traceback (most recent call last):
BUILDSTDERR:   File "setup.py", line 26, in <module>
BUILDSTDERR:     'elephant/spade_src/fim.so')
BUILDSTDERR:   File "/usr/lib64/python3.7/urllib/request.py", line 247, in urlretrieve
BUILDSTDERR:     with contextlib.closing(urlopen(url, data)) as fp:
[...]

PR #217, opened 2019-04-17
Changed setup.py to not crash when fim.so download is unavailable for some reason

mdenker:
This should fix assumed install problem in #216

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,13 +36,13 @@ if oext:
         arch = '64'
     else:
         arch = '32'
-    
+
     try:
-        urlretrieve('http://www.borgelt.net/bin' + arch + '/py' + py_ver + '/fim' + oext,
+        urlretrieve('http://www.borgelt.net/bin' +
+                    arch + '/py' + py_ver + '/fim' + oext,
                     'elephant/spade_src/fim' + oext)
     except:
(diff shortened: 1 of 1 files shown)


mdenker:
I am aware that it's not nice to have `except` without an explicit error in a `try` statement, but I am not sure what Error to put -- in doing some testing with things that could go wrong when internet is down or whatever, there were various types of errors that could be raised.

mdenker:
@alperyeg Hi Alper, would you mind having a look at this PR, intended to catch problems during install when, e.g., the network is not available or has a weird proxy or some other reason?

setup.py
```diff
@@ -29,19 +29,20 @@
...
+        arch = '64'
+    else:
+        arch = '32'
+    
+    try:
+        urlretrieve('http://www.borgelt.net/bin' + arch + '/py' + py_ver + '/fim' + oext,
+                    'elephant/spade_src/fim' + oext)
+    except:
+        print("Unable to download fim" + oext + " module.")
+        pass
```
alperyeg: You don't need `pass`

mdenker: Ah, no, of course not, thanks...

mdenker:
@alperyeg Thanks for your check.

PR #219, opened 2019-05-16
Added snr_hatsopoulos with unittests into statistics.py.

essink:
Implemented SNR calculation according to Hatsopoulos, N. G., Xu, Q. & Amit, Y.  Encoding of Movement Fragments in the Motor Cortex.  J. Neurosci. 27, 5105–5114 (2007).


mdenker:
Thanks for your contribution, that's very helpful. I think there are still a couple of small things to discuss.

1) I'd like to suggest to put this function into a new module that's concerned with features calculated on the basis of spike waveforms. This could then include other functions like calculating wave form width, wave form variability, waveform troughs, etc... A potential name for the module could be perhapts `waveform_features`?

2) Could you please add a paragraph in the doc string to indicate how the method works (in words and/or formula)?

3) Since this is a rather simple function, I'd suggest to rename it to something like `waveform_snr` only. If at some stage we add other alternatives, we could do this using a `method = 'hatsopoulos'` parameter.

4) In the unit tests, there is one function testing multiple things. I think it would be better to split this up in multiple functions to get better granularity.

5) you bumped up the numpy requirement to 1.16.3 -- is this really necessary?

6) Please don't forget to add yourself to `doc/authors.rst`

Issue #190, opened 2018-11-27
Implement VERSION file for version number


mdenker:
Similar to the Neo project, Elephant should hold the version number in a VERSION file that is parsed in relevant places.

dizcza:
Implemented in https://github.com/NeuralEnsemble/elephant/pull/231

PR #222, opened 2019-05-24
simplified travis build

dizcza:
Here are the build configs:

* conda 2.7
* pip 2.7
* pip 3.5
* conda 3.6
* pip 3.6
* pip 3.6 requirements-extras (tests `requirements-extras.txt`, including mpi)
* conda 3.7

Excluded tests:
* pip 3.6 requirements min version (replaces `>=` to `==` in `requirements.txt`). This test set is excluded because of unmet dependencies in neo, quantities, and scipy

Additional fixes:
* one `elephant/VERSION` file that controls the package version (https://github.com/NeuralEnsemble/elephant/issues/190)
* download `fim.so` at `pip install` and at `import elephant.spade`. This fixes travis issue so that `test_fpgrowth_fca` spade test is performed as well (it has been always skipped before).


dizcza:
Finally, can be **squashed** and merged.

.travis.yml
```diff
@@ -1,44 +1,84 @@
...
+      bash miniconda.sh -b -p ${MINICONDA_PATH};
+      conda config --set always_yes yes;
+      conda update conda;
+      conda create --name elephant_env python=${TRAVIS_PYTHON_VERSION};
+      source activate elephant_env;
+      conda install mkl;
+      conda config --append channels conda-forge;
+      sed '/^neo/d' requirements.txt > requirements-conda.txt;
+      conda install --file requirements-conda.txt;
+      conda install -c conda-forge python-neo;
```
mdenker: This would, to my understanding, install always the newest Neo version from forge for conda environments, however, requirements will typically indicate a fixed major version of neo, e.g.,
`neo>=0.7.1,<0.8.0`
I think it may be better to either install only the neo package via pip similar to the `sed` trick above, or to specify the neo version on the conda install line.

.travis.yml
```diff
@@ -1,44 +1,84 @@
...
+  - pip list
+  - python --version
+
+script:
+  nosetests --with-coverage --cover-package=elephant
+
 after_success:
-    - if [[ "$COVERAGE" == "true" ]]; then coveralls || echo "failed"; fi
-cache: apt
+  coveralls || echo "coveralls failed"
```
mdenker: I am not sure why coveralls stopped reporting it's current coverage on the github PR page. Do you think anything changed in the config could cause this? I didn't find anything though. 

In either case, looking into the coveralls site, it seems that it still calculates the correct values, and coverage luckily seems unchanged at 87%...

.travis.yml
dizcza: You are right - it'd be better to leave neo in `requirements-conda.txt`. However, python-neo conda version is outdated and not supported very well. If you run `conda install -c conda-forge neo>=0.7.1` it'll throw you an error indicating that it could not find a package with such a version because the latest one is 0.7.0.

.travis.yml
dizcza: maybe it's due to the fact that i did a force push... i'll make a dummy commit once more

.travis.yml
mdenker: Yes, in fact I am not even sure if the forge neo package is actually done by the project, or by someone outside. So better to stick to pip probably...

dizcza: I found it here https://github.com/NeuralEnsemble/python-neo/issues/233.
OK, I'll replace this line to `pip install neo>=...`

dizcza:
The builds seem to stabilize now but I want to track down test coverage loss. I'll check it the next week.

dizcza:
The difference in test coverage might come from the fact that sometimes `coveralls` report coverage, including tests files, which is useless, and sometimes it reports only the elephant-related coverage, without the tests. To avoid this mistake in the future one need to run ` coverage report --omit="*test*"` or create `.coveragerc` file with the rule to omit tests from coverage. I'll keep an eye on this.

dizcza:
Ready to merge.

setup.py
```diff
@@ -1,49 +1,21 @@
...
+    long_description = f.read()
+with open('requirements.txt') as fp:
     install_requires = fp.read()
 extras_require = {}
 for extra in ['extras', 'docs', 'tests']:
-    with open('requirements-{0}.txt'.format(extra), 'r') as fp:
+    with open('requirements-{0}.txt'.format(extra)) as fp:
         extras_require[extra] = fp.read()
 
-# spade specific
```
mdenker: If I understand correctly, the `fim.so` is now downloaded whenever the `spade` module first is imported. 

Nevertheless, when installing the elephant package from this branch (using `python setup.py install`), `fim.so` will already reside in the install directory automatically even before importing `spade`. This is in my opinion the desired behavior, since a user may not have write privileges to the install folder after installation, e.g. on a system wide install. 

However, I am not sure why the download is actually triggered during install, since the corresponding lines are removed from `setup.py`.

Is this potentially due to the massive importing of all modules in `__init.py__`, which triggers the download during install? If so, is #231 a problem, which removes these imports?

dizcza: The download now is not triggered during the install. I did it because travis does not download fim during the install (travis issue only). But I see the point - it's better to download the file immediately. And if we're about to get rid of pythonic fim implementation, we need to force travis to download this file to maintain the fim tests properly. I'll think how to do this

dizcza:
Now `fim.so` is downloaded at `pip install` and during the import (mostly to support travis tests).
To do so, I merged version_file branch because it includes removing unnecessary imports at `elephant/__init__.py`

Issue #227, opened 2019-06-14
ERROR: test_cubic (elephant.test.test_cubic.CubicTestCase) on 32 bit arches


sanjayankur31:
We're building elephant for multiple arches for [neurofedora](https://docs.fedoraproject.org/en-US/neurofedora/overview/). This one test fails on 32 bit architectures only. Here's the output from a failed i686 build:

```
BUILDSTDERR: ERROR: test_cubic (elephant.test.test_cubic.CubicTestCase)
BUILDSTDERR: ----------------------------------------------------------------------
BUILDSTDERR: Traceback (most recent call last):
BUILDSTDERR:   File "/builddir/build/BUILD/elephant-0.6.2/elephant/test/test_cubic.py", line 76, in test_cubic
BUILDSTDERR:     self.data_array, alpha=self.alpha)
BUILDSTDERR:   File "/builddir/build/BUILD/elephant-0.6.2/elephant/cubic.py", line 115, in cubic
BUILDSTDERR:     pval = _H03xi(kappa, xi, L)
BUILDSTDERR:   File "/builddir/build/BUILD/elephant-0.6.2/elephant/cubic.py", line 150, in _H03xi
BUILDSTDERR:     'kappa(2)= %f<%f=kappa(1)!!!' % (kappa[1], kappa[0]))
BUILDSTDERR: ValueError: H_0 can not be tested:kappa(2)= -0.130398<0.030000=kappa(1)!!!
BUILDSTDERR: ----------------------------------------------------------------------
```

- Failed i686 build: https://koji.fedoraproject.org/koji/taskinfo?taskID=35534431 
- The complete build log for this i686 build is here:  [elephant-i686-failed-build.txt](https://github.com/NeuralEnsemble/elephant/files/3289575/elephant-i686-failed-build.txt)
- This file includes what the "buildroot" was composed of---versions of packages and so on: [elephant-i686-failed-root.txt](https://github.com/NeuralEnsemble/elephant/files/3289581/elephant-i686-failed-root.txt)

Also fails on armv7hl:

- Failed armv7hl build: https://koji.fedoraproject.org/koji/taskinfo?taskID=35534721
- build log: [elephant-armv7hl-failed-build.txt](https://github.com/NeuralEnsemble/elephant/files/3289820/elephant-armv7hl-failed-build.txt)
[...]

PR #229, opened 2019-06-17
fixed 32-bit arch test_cubic.py

dizcza:
Fixes https://github.com/NeuralEnsemble/elephant/issues/227.

### DIFF
--- a/elephant/cubic.py
+++ b/elephant/cubic.py
@@ -143,11 +143,9 @@ def _H03xi(kappa, xi, L):
 
     # Check the order condition of the cumulants necessary to perform CuBIC
     if kappa[1] < kappa[0]:
-        # p = errorval
-        kstar = [0]
         raise ValueError(
             'H_0 can not be tested:'
-            'kappa(2)= %f<%f=kappa(1)!!!' % (kappa[1], kappa[0]))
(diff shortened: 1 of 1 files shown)


elephant/cubic.py
```diff
@@ -199,23 +197,16 @@ def _kstat(data):
...
+        The first three unbiased cumulants of the population count
     '''
-    L = len(data)
-    if L == 0:
+    if len(data) == 0:
         raise ValueError('The input data must be a non-empty array')
-    S = [(data ** r).sum() for r in range(1, 4)]
-    kappa = []
-    kappa.append(S[0] / float(L))
-    kappa.append((L * S[1] - S[0] ** 2) / (L * (L - 1)))
```
mdenker: Note: Double-checked using analytical formulas given in scipy help.

elephant/cubic.py
```diff
@@ -199,23 +197,16 @@ def _kstat(data):
...
-    kappa.append(
```
mdenker: Note: Verified numerically that both old and new implementation of moments yield `[0.03, 0.2991029910299103, 2.9731431937011825]` with the unit test data.

Issue #228, opened 2019-06-14
Documentation build failure


sanjayankur31:
With the 0.6.2 release, and the latest git snapshot, the package documentation does not build. To verify, I tested it out in a clean python 3 virtual environment where I installed all the required deps from the `requirements*.txt` files:

```
$ git log
commit 61b9fd0817063b4ed77c9dc406241ea73f0dcdc8 (HEAD -> master, origin/master, origin/HEAD)
..

$ pip freeze
alabaster==0.7.12
Babel==2.7.0
certifi==2019.3.9
chardet==3.0.4
docutils==0.14
elephant==0.6.2
idna==2.8
imagesize==1.1.0
Jinja2==2.10.1
joblib==0.13.2
MarkupSafe==1.1.1
neo==0.7.1
nose==1.3.7
numpy==1.16.4
numpydoc==0.9.1
packaging==19.0
pandas==0.24.2
Pygments==2.4.2
pyparsing==2.4.0
python-dateutil==2.8.0
pytz==2019.1
quantities==0.12.3
requests==2.22.0
scikit-learn==0.21.2
scipy==1.3.0
six==1.12.0
snowballstemmer==1.2.1
Sphinx==2.1.1
sphinxcontrib-applehelp==1.0.1
sphinxcontrib-devhelp==1.0.1
sphinxcontrib-htmlhelp==1.0.2
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.2
sphinxcontrib-serializinghtml==1.1.3
urllib3==1.25.3
```

```
$ make html
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v2.1.1
making output directory... done
loading intersphinx inventory from http://docs.python.org/objects.inv...
intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 26 source files that are out of date
updating environment: 26 added, 0 changed, 0 removed
/home/asinha/dump/py3-virts/elephant/lib64/python3.7/site-packages/numpydoc/docscrape.py:424: UserWarning: Unknown section Example in the docstring of <function joint_probability_matrix at 0x7fbfb0c3f0d0> in ../elephant/asset.py.
  warn(msg)

Exception occurred:
[...]

PR #230, opened 2019-06-18
docs fix

dizcza:
Fixes https://github.com/NeuralEnsemble/elephant/issues/228.

The main issue was `probability_matrix_analytical()` --> `probability_matrix_analytical : ` in the doc string.


readthedocs.yml
```diff
@@ -1,2 +1,7 @@
-conda:
-  file: doc/environment.yml
```
mdenker: We may need to reintroduce the `environments.yml` again at a later stage when redoing the documentation system with some advanced features, but for now I agree it is better to stick to the standard RTD build and standard dependencies.

PR #231, opened 2019-06-18
VERSION file; refactored MANIFEST.in

dizcza:
1. Made one file - VERSION - that keeps track of the current package version.
2. `include_package_data=True` tells an installer to include all files from `MANIFEST.in`.
3. Removed `from . import ...` in `elephant/__init__.py` since they are not used in `__init__.py` later on. Avoid the bad practice of using `from elephant import *` style.


elephant/__init__.py
```diff
@@ -6,29 +6,13 @@
 :license: Modified BSD, see LICENSE.txt for details.
 """
 
-from . import (statistics,
```
mdenker: I think the original reason was to assist users by not having to import each module individually, but to be able to do `import elephant` and have all submodules available. I do agree it's cleaner to insist on importing the individual modules, especially as the package grows. This change should be highlighted in the 0.7.0 release notes, as it may break existing code (in a trivial way).

dizcza: Yes, it'll be a breaking change. Since the next release will be 0.7.0 (and not 0.6.x), I guess it's OK to do it now... or don't do this at all.
When we'll be making a new release, we should mention this behavior.

dizcza:
Regarding the line in the developer's guide.
```
First, check that the version string in :file:`elephant/VERSION` is correct.
```

I don't understand the point of putting this line in the doc in the way it is now. Consider either removing this line at all (it's not clear what "correct" version means) or explaining why the developer should care about the version file. For example, if he/she fixes a bug, the minor digit 0.6.x has to be incremented.

mdenker:
The line is under the heading "Making a release" and was kept as a reference to make sure all developers/maintainers create releases in the same manner to ensure consistency.

To avoid the confusion you spotted, we could think of dividing these two points into to separate pages, one for _developers_ (i.e., contributors) and one for _maintainers_ (i.e., those making releases etc).

dizcza:
Now merged in https://github.com/NeuralEnsemble/elephant/pull/222

elephant/__init__.py
jpgill86: I raised #237 before noticing this old discussion. Let me just add my two cents.

In my opinion, removing these imports is not a good idea. It breaks existing code without gaining anything.

> Avoid the bad practice of using `from elephant import *` style.

There is a big difference between using `from . import (statistics, ...)` in `__init__.py` and using `from elephant import *` in some downstream project. The latter is rightly [vilified](https://stackoverflow.com/a/2360808/3314376) because it pollutes namespaces in unpredictable ways. The use of `from . import (statistics, ...)` in `__init__.py` does not do this for end-users. Furthermore, it does not pollute the internal `elephant` namespace because `statistics` etc. remain in their own sub-namespaces (`elephant.statistics`, etc.; contrast this to `from .statistics import *`).

I think one valid reason for removing top-level imports like these is for efficiency. If (1) the submodules are independent enough such that most files do not need to be read by the interpreter to load and execute any particular function, and (2) the submodules are so numerous and large that importing all is costly, then this type of performance optimization could make sense. However, I haven't noticed any discussion of that here, nor have I noticed importing elephant to be particularly slow.

dizcza: You are right. Package's `__init__.py` should list available modules, so that users don't need to go to docs and check what's there. And `from elephant import *` is up to the user. That is something I overthought in this PR.

PR #232, opened 2019-06-27
fixed readthedocs.yml

dizcza:
In one of the pull requests I made I forgot to merge readthedocs.yml, and currently, the docs badge shows the docs build is failed.
So here it is.
Built docs: https://elephant.readthedocs.io/en/docs_fix2/

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,7 +1,15 @@
-name: elephant
+# readthedocs version
+version: 2
 
 build:
     image: latest
 
+sphinx:
+  builder: html
+  configuration: doc/conf.py
+
 python:
     version: 3.7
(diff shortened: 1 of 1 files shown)


mdenker:
[Merging despite CI checks not showing complete -- on travis all is fine]

dizcza:
On travis all is fine because travis does not check the docs build. Currently, readthedocs does not support API not to merge the PR if the docs are failing. They're working on it, however. So basically we never sure about the docs unless it's merged in the master and then readthedocs starts building the docs, and only then we can see if it's successful or not. We might think to add the docs build as another test on travis. However, this breaks the logic of CI since one instrument starts doing the job of another one...

PR #233, opened 2019-07-03
Integrated GPFA

dizcza:
Only `gpfa` method is integrated here among all other choices in neural_trajectory.
We'll add new functionality once it's implemented by our researchers.


elephant/gpfa.py
```diff
@@ -0,0 +1,396 @@
...
+    ------
+    AssertionError
+        If `seqs` es empty.
+
+    """
+    assert len(seqs) > 0, "Got empty trials."
+
+    # Set cross-validation folds
+    num_trials = len(seqs)
+    # fdiv = np.linspace(0, num_trials, num_folds + 1).astype(np.int) if \
```
alperyeg: Can be removed

elephant/gpfa.py
```diff
@@ -0,0 +1,396 @@
...
+    y_all = np.hstack(seqs_train['y'])
+    y_dim = y_all.shape[0]
+
+    if np.linalg.matrix_rank(np.cov(y_all)) < y_dim:
+        errmesg = 'Observation covariance matrix is rank deficient.\n' \
+                  'Possible causes: ' \
+                  'repeated units, not enough observations.'
+        raise ValueError(errmesg)
+
+    print('Number of training trials: {}'.format(len(seqs_train)))
```
alperyeg: Maybe this should be optional. Either as logging, or as an optional verbose variable as it is done in the `gpfa_core.py`.

elephant/gpfa_src/gpfa_util.py
```diff
@@ -0,0 +1,579 @@
...
+    Returns the solution to x b = a. Equivalent to MATLAB right matrix
+    division: a / b
+    """
+    return np.linalg.solve(b.T, a.T).T
+
+
+def logdet(A):
+    """
+    log(det(A)) where A is positive-definite.
+    This is faster and more stable than using log(det(A)).
```
alperyeg: Is this equivalent in python/numpy?

elephant/utils.py
```diff
@@ -0,0 +1,21 @@
+import quantities as pq
+
+
+def check_quantities(var, var_name):
```
alperyeg: Is this module a general file to provide utilities for `elephant`? That should be indicated.

elephant/gpfa.py
dizcza: yeah, a logger is another thing I'm thinking on

dizcza:
As a side note, gpfa can be implemented as a class. Doing so will reduce a large number of repeated arguments (like `verbose`, `num-of-iters`, etc.) that are passed all over the gpfa functionality. I guess, if we agree to this, it'll be a separate PR.

elephant/gpfa_src/gpfa_util.py
dizcza: This is equivalent to `np.log(np.linalg.det(A))` and, what is more important, numerically stable. However, this function is actually _slower_ than numpy's version. Maybe because I've installed MKL.
A good catch. I'll add a special test.
Figure_1

elephant/test/test_gpfa.py
```diff
@@ -0,0 +1,174 @@
+# -*- coding: utf-8 -*-
+"""
+Unit tests for the neural_trajectory analysis.
```
Junji110: neural_trajectory --> GPFA

elephant/test/test_gpfa.py
```diff
@@ -0,0 +1,174 @@
...
+else:
+    HAVE_SKLEARN = True
+    from elephant.gpfa_src import gpfa_util
+    from elephant.gpfa import gpfa
+
+python_version_major = sys.version_info.major
+
+
+@unittest.skipUnless(HAVE_SKLEARN, 'requires sklearn')
+class NeuralTrajectoryTestCase(unittest.TestCase):
```
Junji110: NeuralTrajectoryTestCase --> GPFATestCase

elephant/test/test_gpfa.py
```diff
@@ -0,0 +1,174 @@
...
+        # generate a positive definite matrix
+        matrix = np.random.randn(20, 20)
+        matrix = matrix.dot(matrix.T)
+        logdet_fast = gpfa_util.logdet(matrix)
+        logdet_ground_truth = np.log(np.linalg.det(matrix))
+        assert_array_almost_equal(logdet_fast, logdet_ground_truth)
+
+
+def suite():
+    suite = unittest.makeSuite(NeuralTrajectoryTestCase, 'test')
```
Junji110: NeuralTrajectoryTestCase --> GPFATestCase

elephant/utils.py
```diff
@@ -0,0 +1,30 @@
...
+        input argument
+    var_name : str
+        variable name
+
+    Raises
+    -------
+    AssertionError
+        If the input `var` is not a `pq.Quantity`
+    """
+    assert isinstance(var, pq.Quantity), \
```
Junji110: I don't see a merit in defining this function.
Does this increase readability/maintainability of code so much?

Junji110:
I found there are number of cases where `assert` is used to check the type or shape of input variables, but I don't think this is a proper usage of `assert`; you should rather raise `ValueError` in such cases.
To my understanding, `assert` should be used to check the internal logic of the code, but not to check whether users use the code in a proper way.

elephant/gpfa.py
```diff
@@ -0,0 +1,402 @@
...
+                  'repeated units, not enough observations.'
+        raise ValueError(errmesg)
+
+    if verbose:
+        print('Number of training trials: {}'.format(len(seqs_train)))
+        print('Number of test trials: {}'.format(len(seqs_test)))
+        print('Latent space dimensionality: {}'.format(x_dim))
+        print('Observation dimensionality: {}'.format(has_spikes_bool.sum()))
+
+    min_var_frac = 0.01
```
Junji110: This hard-coded number should rather be given as a keyword-argument (with a default of 0.01) of the function

elephant/utils.py
dizcza: Yes for maintainability and consistency. It does not improve readability.
Reason: instead of typing the same thing each time in each function.

dizcza:
> I found there are number of cases where `assert` is used to check the type or shape of input variables, but I don't think this is a proper usage of `assert`; you should rather raise `ValueError` in such cases.
> To my understanding, `assert` should be used to check the internal logic of the code, but not to check whether users use the code in a proper way.

Actually yes, `ValueError` better fits the philosophy of checking the function input, while asserts are used as "I swear this cannot happen". I'll go through input checking asserts and change them to ValueError.

elephant/utils.py
dizcza: Forgot to mention. I'll add this function check to all elephant's public functions.

Junji110: Then you should define this function in that PR, not here.

Junji110:
I've refactored unit tests and added a new test on dimensionality reduction performance of GPFA

elephant/gpfa.py
```diff
@@ -0,0 +1,407 @@
...
+                GP noise variances
+            d: ndarray of shape (#units, 1)
+                observation mean
+            C: ndarray of shape (#units, #latent_vars)
+                mapping between the neuronal data space and the latent variable
+                space
+            R: ndarray of shape (#units, #latent_vars)
+                observation noise covariance
+
+    seqs_train: numpy.recarray
```
mdenker: I am not sure I understand the advantage of using `recarray`  instead of a classic list of `dict` here? Would
```python
seqs_train['y'][0] # recarrays, trial 0
```
be preferrable over
```python
seqs_train[0]['y'] # list of dicts, trial 0
```
from a user perspective?

elephant/gpfa.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+def gpfa(data, bin_size=20*pq.ms, x_dim=3, em_max_iters=500):
+    """
+    Prepares data and calls functions for extracting neural trajectories.
+
+    Parameters
+    ----------
+
+    data : list
```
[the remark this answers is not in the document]
dizcza: I also agree that the trial ID is useless here because it always goes as `range(len(data))`. I also added TODO below. Since this is GPFA-related issue only, we might change this behaviour now, not waiting for the next issue and corresponding fix (some people might start using it immediately). Discussion is needed.

elephant/gpfa.py
```diff
@@ -0,0 +1,407 @@
...
+                neural data
+            * xsm: ndarray of shape (#latent_vars, #bins)
+                posterior mean of latent variables at each time bin
+            * Vsm: ndarray of shape (#latent_vars, #latent_vars, #bins)
+                posterior covariance between latent variables at each
+                timepoint
+            * VsmGP: ndarray of shape (#bins, #bins, #latent_vars)
+                posterior covariance over time for each latent variable
+
+    seqs_test: numpy.recarray
```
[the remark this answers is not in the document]
dizcza: Since cross-validation is not implemented yet, we can remove `seqs_test` at all. Discussion is needed.

elephant/gpfa.py
```diff
@@ -0,0 +1,407 @@
...
+neural trajectories. The parameters (C, d, R) are estimated from the data using
+FACTOR ANALYSIS technique. GPFA is simply a set of Factor Analyzers (FA),
+linked together in the low dimensional space by a Gaussian Process (GP).
+
+The analysis consists of the following steps:
+
+0) bin the data to get a sequence of N dimensional vectors for each time
+    bin, and choose the reduced dimension x_dim;
+
+1) run
```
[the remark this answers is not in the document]
dizcza: Agree. Some smart words are needed.

elephant/gpfa.py
```diff
@@ -0,0 +1,407 @@
...
+            * method: str, Method name.
+
+    Raises
+    ------
+    AssertionError
+        If `data` is an empty list.
+        If `bin_size` if not a `pq.Quantity`.
+        If `data[0][1][0]` is not a `neo.SpikeTrain`.
+
+    Examples
```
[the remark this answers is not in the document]
dizcza: This is a numpy-doc reserved keyword. There is no "Example" for a singular one.

elephant/gpfa.py
```diff
@@ -0,0 +1,407 @@
...
+    >>> from elephant.spike_train_generation import homogeneous_poisson_process
+    >>> data = []
+    >>> for trial in range(50):
+    >>>     n_channels = 20
+    >>>     firing_rates = np.random.randint(low=1, high=100,
+    >>>         size=n_channels) * pq.Hz
+    >>>     spike_times = [homogeneous_poisson_process(rate=rate)
+    >>>         for rate in firing_rates]
+    >>>     data.append((trial, spike_times))
+    >>> params_est, seqs_train, seqs_test, fit_info = gpfa(
```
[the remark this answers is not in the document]
dizcza: We discussed it with Junji a bit. The most obvious "explanation" would be a plot. But it requires matplotlib. The docs example is shown only to show the usage code, not the usefulness. Discussion is needed.

requirements-extras.txt
```diff
@@ -1,3 +1,4 @@
 pandas>=0.14.1
 scikit-learn
 mpi4py>=3.0.1
+tqdm
```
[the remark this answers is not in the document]
dizcza: Yes, I'll add tqdm in other functions as well. I already added some todos in the following PRs where tqdm should replace handwritten progress bar.

elephant/gpfa.py
dizcza: I fully agree not using `np.recarray`. The authors should provide a clear reason why they stick to np.recarray over python-friendly structures. Apart from confusing the users (and probably developers as well), it has several disadvantages during the test setup and future maintenance.

Junji110: I also though to define `seqs_train` as a dict rather than a recarray, but I found that there are many part where `seqs_train['y']` is subjected to vector operation, which is not quickly done with a dict.
We could still define it as a dict of arrays (of the same length, i.e., # of trials) representing `y`, `x`, and so on, but it would require a total rewrite of the whole code.

By the way, for a recarray `seqs_train`, `seqs_train['y'][0]` is identical to `seqs_train[0]['y']`.

elephant/gpfa.py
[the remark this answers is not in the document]
Junji110: The trial ID is inherited from the original MATLAB code, and the idea behind this would be that trial IDs can start from arbitrary number instead of 0 or 1, and there could even be missing numbers in between.
I personally don't think this feature is necessary, and if we get rid of this, the input structure would be much simpler.

dizcza:
TODO list:
1. Replace `np.recarray` -> `dict`: hide `np.recarray` from the user in GPFA internal private functional, but still operate with dicts as input and output type.
2. Add an example where test data is used as not a part of the training data.
3. Move functions to OOP.

Issue #220, opened 2019-05-20
Butterworth filtering and theta bandpass


jmbudd:
Hi 

I have encountered what I think may be an issue with bandpass filtering for theta band using Elephant. For my working, see [1] and [2]; I have used a test signal with a sampling rate (fs) of 5000 Hz to match data I am analysing but which I can't share. 

I think this may arise because [signal_processing.py](https://github.com/NeuralEnsemble/elephant/blob/master/elephant/signal_processing.py) uses scipy 'ba' rather than 'sos' output for Butterworth filters and this problem arose for order > 3 (current default value is 4). 

Used scipy (1.2.1), elephant (0.6.2) and neo (0.7.1).

N.B. Originally posted on [NeuralEnsemble](https://groups.google.com/d/msg/neuralensemble/Q9Zs8p-8sJM/SIzCYDmnAQAJ)

[1]  [bandpass_test.pdf](https://github.com/NeuralEnsemble/elephant/files/3197578/bandpass_test.pdf)
[2] [bandpass_test.ipynb](https://github.com/jmbudd/misc/blob/master/bandpass_test.ipynb)

mdenker:
Thanks for sharing your analysis and putting this report here from the mailing list. For a start, I definitely can reproduce the behavior you demonstrate.

dizcza:
`scipy.butter` by default uses 'ba' output. I don't know why they stick to 'ba' rather than 'sos', I have little experience in frequency filtering. But I think, as it's done in all great python packages, it's due to the fact that, on average, it's working well, even though in some particular cases - one that you generously demonstrated - it might not work as expected. Since the first implementation of `signal_processing.butter`, 2015, to the best of my knowledge, you're the first one who encountered such an issue. So changing the default scipy's `ba` to `sos` just because of one particular case seems for me too much.

What we can do, however, is to add `output=ba` keyword in `signal_processing.butter` function. So you can change it to whatever you like. What do you think, is it worth doing so?

jmbudd:
Thanks for your response. Yes, having the option to use 'sos' output would be helpful but the code would also need to include sosfiltfilt() filtering. But I am not sure I agree that 'ba' should remain the default because it seems best practice not to use this representation. This is not because there is a problem with scipy implement but that this method is numerically unstable (for discussion, see SO answer [1]): 

- Scipy [2] states " It’s recommended to use second-order sections format when filtering, to a void numerical error with transfer function (ba) format". See also scipy github [3] issue. 
[...]

PR #234, opened 2019-07-05
Butterworth supports sosfiltfilt filter_function

dizcza:
Fixes https://github.com/NeuralEnsemble/elephant/issues/220.

### DIFF
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -304,10 +304,15 @@ def butter(signal, highpass_freq=None, lowpass_freq=None, order=4,
     order : int
         Order of Butterworth filter. Default is 4.
     filter_function : string
-        Filtering function to be used. Either 'filtfilt'
-        (`scipy.signal.filtfilt()`) or 'lfilter' (`scipy.signal.lfilter()`). In
-        most applications 'filtfilt' should be used, because it doesn't bring
-        about phase shift due to filtering. Default is 'filtfilt'.
+        Filtering function to be used. Available filters:
+            * 'filtfilt': `scipy.signal.filtfilt()`;
+            * 'lfilter': `scipy.signal.lfilter()`;
+            * 'sosfiltfilt': `scipy.signal.sosfiltfilt()`.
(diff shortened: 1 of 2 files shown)


elephant/test/test_signal_processing.py
```diff
@@ -398,7 +398,13 @@ def test_butter_filter_function(self):
         _, psd_lfilter = spsig.welch(
```
mdenker: Please update comment on line 388/389 above to include `sosfilter`

elephant/signal_processing.py
```diff
@@ -304,10 +304,15 @@ def butter(signal, highpass_freq=None, lowpass_freq=None, order=4,
...
-        most applications 'filtfilt' should be used, because it doesn't bring
-        about phase shift due to filtering. Default is 'filtfilt'.
+        Filtering function to be used. Available filters:
+            * 'filtfilt': `scipy.signal.filtfilt()`;
+            * 'lfilter': `scipy.signal.lfilter()`;
+            * 'sosfiltfilt': `scipy.signal.sosfiltfilt()`.
+        In most applications 'filtfilt' should be used, because it doesn't
+        bring about phase shift due to filtering. For numerically stable
+        filtering, use 'sosfiltfilt' (see issue
+        https://github.com/NeuralEnsemble/elephant/issues/220).
```
mdenker: From what I understood, sos filtering decomposes a FIR filter or some order into a series of second order filters, so that this is useful in particular to generate higher filter orders (e.g., https://de.wikipedia.org/wiki/Filter_mit_endlicher_Impulsantwort#Kaskadierte_Form , sadly not translated in the English version of wikipedia). If this is true, maybe one could hint at this by saying, e.g., 
`For numerically stable filtering, in particular at higher filter orders, use sosfiltfilt...`

dizcza: Yeah, I wanted to mention higher order filtering but missed it at some point. I'll add the note.

mdenker:
I could confirm this function now works with the notebook originally submitted in #220

Issue #10, opened 2014-07-10
Dependency versions


toddrjen:
Our minimum version requirements include numpy 1.5.0 and scipy 0.9.0.  However, our Ubuntu tests only test the latest version of all packages, and the miniconda tests only support numpy >= 1.6.2 and scipy >= 0.11.0.  

I am a bit uncomfortable listing dependencies that our tests cannot verify work properly.  Installing either numpy or scipy (not to mention both) manually in travis takes forever, so that is not a feasible approach. 

Even someone tests manually before official releases, a lot of small and subtle bugs sneak through that can be very hard to track to a specific change.  

So I am thinking it might be best to set our minimum dependencies versions to be the versions supported by travis or greater.  What does everyone else think?

alperyeg:
@apdavison perhaps Numpy 1.9 is downloaded because in the requirements it is stated that the version should be higher than 1.6.2. So if Elephant is installed per pip e.g. in a virtualenv numpy 1.9 will be automatically downloaded and installed.

apdavison:
@alperyeg That's not how I interpreted the message. He stated that he "needed" to install 1.9, which suggests to me that he already had an earlier version installed, and it didn't work. I guess I should follow up to find out which earlier version he tried and what the error message was.

dizcza:
> set our minimum dependencies

We provide the versions as `scipy>=X.Y.Z` but this does not mean all X.Y.Z will work because it might clash with the requirement of another package. Therefore, the solution to this issue requires building a dependency tree, which, of course, is something we as the developers won't do.

PR #235, opened 2019-07-09
fixed min requirements

dizcza:
Currently, the min package requirements versions are broken. I found the corresponding issue https://github.com/NeuralEnsemble/elephant/issues/10.
This commit fixes it and includes "pip 3.6 requirements-extras min version" test in travis to make sure we will not break dependencies again.

* Min scikit-learn version that supports python 2.7 and 3.5 is 0.19.0. That triggers numpy>=1.9.3 but v1.9.3 is broken, and the next stable one is 1.10.1.
* Min quantities version that works well with the latest neo==0.7.1 is 0.12.1, not 0.10.1 (neo's guys should change min quantities version as well).
* `DataFrame.sort_index(level=...)` was introduced in pandas v0.18.0.

One problem that I see now is that "pip 3.6 requirements-extras min version" test case takes 20 minutes while the other ones finish in less than 10 min. Building numpy, scipy and pandas old versions takes a lot of time. This is the caveat that is thankfully avoided in later and up-to-date versions. Consider excluding this test once again if that will harden the continuous development and integration.

-----
I also added `.coveragerc` file to ignore tests from being counted as source code when the coverage command is run.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+# .coveragerc to control coverage.py
+[report]
+omit = elephant/test/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,8 @@ matrix:
       python: 3.6
       env: DISTRIB="pip"
       before_install:
-        - sudo apt install libblas-dev liblapack-dev libatlas-base-dev gfortran
+        - sudo apt install -y libopenmpi-dev openmpi-bin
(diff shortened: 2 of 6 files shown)


mdenker:
Is this something to include in 0.6.3, or rather in 0.7? I approved, but did not merge.

dizcza:
I suggest merging this PR after we release v0.6.3 since the requirements are changed a lot.

PR #243, opened 2019-07-30
STTC speedup by using numpy instead of loops

bjoern1001001:
The current implementation of STTC is following the implementation provided along the original paper very closely. This makes it easy to understand and it's very clear the implementation does the same. 
However, directly transcribing C code to Python makes it really slow. 
This PR provides a considerable increase in performance (multiple hours => few minutes) due to using numpy at the cost of not being easily comparable to the original implementation.

The two main parts of this PR are:
- in run_T: Instead of looping over the spike time differences one by one, this is now vectorized, computing a vector of differences and summing over it. 
This should be relatively easy to understand and compare to the original code.
- in run_P: What this function does, is that for every spike in `spiketrain_1` it checks if there is any spike in `spiketrain_2` within `dt` of it. Instead of looping over every single spike, the implementation here uses a binary search algorithm. This INCREASES runtime complexity, but due to the numpy implementation it outperforms the current version even for very large spiketrains.

I'm not sure if moving this far away from the original code is desired, so please let me know what you think.

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -656,16 +656,30 @@ def spike_time_tiling_coefficient(spiketrain_1, spiketrain_2, dt=0.005 * pq.s):
         within dt
         """
 
+        # Search spikes of spiketrain_1 in spiketrain_2
+        # ind will contain index of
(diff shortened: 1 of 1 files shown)


bjoern1001001:
Replaced by #244

PR #252, opened 2019-09-24
Feature: Joint-ISI dithering

pbouss:


elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
+'''
+Module to generate surrogates of spike trains by using the joint-ISI dithering.
```
stellalessandra: Few comments:
- I think that the joint-isi dithering class should be included in the spike train surrogates module
- I also think that the structure should be similar to the spiketrain surrogate module
I don't know if this is a massive job though

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+    HAVE_PLT = True
+except ImportError:
+    HAVE_PLT = False
+
+class Joint_ISI_Space:
+    '''
+    The class :class:`Joint_ISI_Space` is implemented for Joint-ISI dithering
+    as a continuation of the ideas of Louis et al. (2010) and Gerstein (2004).
+
+    When creating an class instance all necessary preprocessing steps are done,
```
stellalessandra: This is principle should be changed. Spiketrains with too low firing rate should be dithered by the joint isi anyway: maybe adapting the width of the gaussian kernel accordingly (?)

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        If True, also the way of how the dithered spikes are evaluated
+        is returned so 'uniform' for uniform and dithering and 'jisid' for
+        joint-ISI-dithering
+        Default: False
+    use_sqrt: boolean
+        if use_sqrt == True a sqrt is applied to the joint-ISI histogram,
+        following Gerstein et al. 2004
+        Default: False
+    method: string
+        if 'fast' the entire diagonals of the joint-ISI histograms are
```
stellalessandra: rewrite? The description is cryptic

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        Default: 1
+    dither: pq.Quantity
+        The range of the dithering for the uniform dithering,
+        which is also used for the method 'window'.
+        Default: 15.*pq.ms
+    unit: pq.unit
+        The unit of the spiketrain in the output.
+        Default: pq.s
+    window_length: pq.Quantity
+        The extent in which the joint-ISI-distribution is calculated.
```
stellalessandra: Write more explicitly what you mean with extent?

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        Default: 120*pq.ms
+    num_bins: int
+        The size of the joint-ISI-ditribution will be num_bins*num_bins.
```
stellalessandra: typo -> distribution
ps: why would you want to change this parameter?

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        Default: 120
+    sigma: pq.Quantity
+        The standard deviation of the Gaussian kernel, with which
+        the data is convoluted.
+        Default: 0.001*pq.s
+    isi_median_threshold: pq.Quantity
```
stellalessandra: look at comment up

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        Only if the median of the ISI distribution is smaller than
+        isi_median_threshold the Joint-ISI dithering is applied, if not the
+        uniform dithering is used.
+        Default: 30*pq.ms
+    alternate: boolean
+        If alternate == True: then first all even and then all odd spikes are
+        dithered. Else: in acending order from the first to the last spike, all
```
stellalessandra: typo -> ascending

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        spikes are moved.
+        Default: True.
+    show_plot: boolean
```
stellalessandra: do you think it makes sense to add a plotting function in the method? I thought we do not add plotting into elephant code

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        if use_sqrt == True a sqrt is applied to the joint-ISI histogram,
+        following Gerstein et al. 2004
+        Default: False
+    method: string
+        if 'fast' the entire diagonals of the joint-ISI histograms are
+        used if 'window' only the values of the diagonals are used, whose
+        distance is lower than dither
+        Default: 'fast'
+    cutoff: boolean
+        if True than the Filtering of the Joint-ISI histogram is
```
stellalessandra: typo -> then

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        limited to the lower side by the minimal ISI.
+        This can be necessary, if in the data there is a certain dead time,
```
stellalessandra: better call it refractory period

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        which would be destroyed by the convolution with the 2d-Gaussian
+        function.
+        Default: True
+    min_spikes: int
```
stellalessandra: see my comment up

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        self.cutoff = cutoff
+        self.min_spikes = min_spikes
+
+        self.preprocessing()
+
+    def preprocessing(self):
+        '''
+        All preprocessing steps for the joint-ISI dithering are done here.
+
+        So first to checks are done. If they are not passed, self.method is
```
stellalessandra: rephrase first sentence?

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        set to 'uniform'. The first one asks for the number of spikes.
```
stellalessandra: with 'one' do you intend the check?

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+        function is calculated.
+
+        If method is 'window':
+        For each point in the joint-ISI distribution a on the line parallel to
+        the antidiagonal all points up to the dither-parameter are included, to
+        calculate the cumulative distribution function.
+
+        The function has no output, but stores its result inside the class.
+        '''
+        if len(self.st) < self.min_spikes:
```
stellalessandra: I would shorten the explanation in the documentation, but add it here in the code, check by check, step by step

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,462 @@
...
+            self.window_length = self.window_length.rescale(
+                self.unit).magnitude
+        if isinstance(self.sigma, pq.Quantity):
+            self.sigma = self.sigma.rescale(self.unit).magnitude
+
+        self.sampling_rhythm = self.alternate + 1
+
+        self.bin_width = self.window_length/self.num_bins
+
+        def index_to_isi(ind):
```
stellalessandra: I think in principle you dont need to define such simple function. Maybe just state what you do in a comment?

elephant/joint_isi_dithering_class.py
pbouss: 1. I also added it to the surrogate method, as you may have seen. And there you can access it as the other surrogate methods (which for the user means less freedom). 
But I liked to give this space to the JOint-ISI. first that it is possible to separate preprocessing and processing as we will use it in SPADE. Also because it makes it much easier to investigate the properties changing the parameters on the fly.

elephant/joint_isi_dithering_class.py
pbouss: Here I'm still thinking of how to do it. And we can maybe discuss within the next days, how to procede with this. I have some ideas, regarding this.

elephant/joint_isi_dithering_class.py
pbouss: Okay, I kicked it off.

elephant/joint_isi_dithering_class.py
pbouss: Because it gives me a finer grid on which the cumulative distribution functions are calculated. And thus the effects of the discretization are less. I think this is normally more import for the low ISI-regime. - Instead it could introduce a binsize parameter a just assert, that it is a multiple of the window length.

elephant/joint_isi_dithering_class.py
pbouss: I do it, so that it gets more clear and also that I the 1/2 is added once, and there is no need to remember it again.

elephant/joint_isi_dithering_class.py
mdenker: I am not really sure I understand the argument to (1) as such. Although I didn't go through the code thoroughly, I understand the reason for the class implementation, and the solution to have both a class and a normal function wrapper is good. However, having a separate module for this one class seems to me massive overkill, the class should be part of the surrogates module.

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,438 @@
...
+        spiketrains were obtained.
+
+    Output:
+    dithered_sts
+        List of spiketrains, dithered with Joint-ISI-dithering
+    mode (only if print_mode=True)
+        string: Indicates, which method was used to dither the spikes.
+            'jisih' if joint-ISI was used,
+            'uniform' if the dense_rate was too low and uniform dithering was
+             used.
```
mdenker: I guess this is the output of the dithering method. Then I am not sure it needs to go here, it's already in the doc string of the function.

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,438 @@
...
+    as a continuation of the ideas of Louis et al. (2010) and Gerstein (2004).
+
+    When creating an class instance all necessary preprocessing steps are done,
+    to use the method dithering().
+
+    To make the dithering-procedure stable, the spiketrain needs to pass two
+    thresholds, that the Joint-ISI dithering is applied, if not uniform
+    dithering is used.
+
+    Attributes
```
mdenker: All these attributes are "public", is this intentional, or should these rather be the arguments of the class constructor?

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,438 @@
...
+        For this spiketrain the surrogates will be created
+
+    n_surr: int, optional
+        Number of surrogates to be created.
+        Default: 1
+    dither: pq.Quantity
+        The range of the dithering for the uniform dithering,
+        which is also used for the method 'window'.
+        Default: 15.*pq.ms
+    unit: pq.unit
```
mdenker: Is changing the units of the output spiketrain an important feature to include in this function? It seems to me that in 99.9% of cases you would expect the same units as the input spike train, and else, one would rescale the result of the jisi dither.

elephant/joint_isi_dithering_class.py
```diff
@@ -0,0 +1,438 @@
...
+            self.sigma = self.sigma.rescale(self.unit).magnitude
+
+        self.sampling_rhythm = self.alternate + 1
+
+        self.bin_width = self.window_length / self.num_bins
+
+        def index_to_isi(ind):
+            return (ind + 0.5) * self.bin_width
+
+        self.index_to_isi = index_to_isi
```
mdenker: Why not just define a private method? This assignment seems overly complex.

elephant/joint_isi_dithering_class.py
stellalessandra: I understand your arguments! Then I would add it as well to the surrogate modules, keeping the class structure, as @mdenker says

elephant/joint_isi_dithering_class.py
pbouss: This was intentional. My idea was to allow, changed this parameters afterwards. Which should allow to investigate easier all parameters. But maybe it is too much and not in all case useful. Since anyway, for most of them parts of the preprocessing needs to be redone.

elephant/joint_isi_dithering_class.py
pbouss: @stellalessandra @mdenker No I placed the class inside the module of the surrogates.

elephant/joint_isi_dithering_class.py
pbouss: But effectively, of course it would be better to define update functions, where only the part of the preprocessing is done again, which is needed....
I will try to work it in the next days.

elephant/spike_train_surrogates.py
```diff
@@ -450,9 +460,691 @@ def jitter_spikes(spiketrain, binsize, n=1):
...
+                    alternate=alternate,
+                    use_sqrt=use_sqrt,
+                    method=method,
+                    cutoff=cutoff,
+                    expected_refr_period=expected_refr_period,
+                    min_spikes=min_spikes
+                    ).dithering()
+
+
+class JointISI:
```
dizcza: always subclass from `object` class - we still need to support python 2.

PR #254, opened 2019-10-09
Fix/spade

stellalessandra:
In this pull request we fixed a few issues in the spade.py module. We
- fixed the calculation of the duration of a pattern, in the output of spade. Previously, in the function restructuring the output format and making it more 'human readable' (concept_output_to_patterns), the calculation of the pattern duration was not always correct, depending on the order of lags and neurons. We fixed the bug by accounting the duration independently from the order of neurons;
- added raise ValueErrors in code when the output_format is not one of the two parameters allowed
- fixed bug of undefined spectrum type when running spade without surrogates (only FIM)
In addition:
- corrected typos in documentation and removed redundant calls to numpy

### DIFF
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -1927,8 +1927,8 @@ def concept_output_to_patterns(concepts, winlen, binsize, pvalue_spectrum=None,
             output_dict['pvalue'] = -1
         # Signature (size, n occ) of the pattern
         elif spectrum == '3d#':
-            sgnt = (len(conc[0]), len(conc[1]), max(
-                np.abs(np.diff(np.array(conc[0]) % winlen))))
+            duration = (max(conc[0]) - min(conc[0])) % winlen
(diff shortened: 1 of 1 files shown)


dizcza:
Hi Alessandra,

can you please edit your comment 

> fixed few bugs in spade.py module

and describe in a few words _what_ exactly it fixes? I suppose, it relates much to `elif spectrum == '3d#'` but it's hard to guess why previous was wrong and now is right.

stellalessandra:
> Hi Alessandra,
> 
> can you please edit your comment
> 
> > fixed few bugs in spade.py module
> 
> and describe in a few words _what_ exactly it fixes? I suppose, it relates much to `elif spectrum == '3d#'` but it's hard to guess why previous was wrong and now is right.

Just updated the description. Please tell me if something is still not clear!

dizcza:
Alessandra, you did it more than enough! :) I very liked your explanation. Consider being a documentation writer for Elephant :+1:

Issue #253, opened 2019-10-08
Plans for compatibility with Neo 0.8.0?


jpgill86:
Neo 0.8.0 was released on September 30. Presently, elephant is marked as incompatible with it:

https://github.com/NeuralEnsemble/elephant/blob/c78af7ea6cff3645bab306393349aa7d89f05a13/requirements.txt#L1

Running ``pip install neo elephant`` in a new environment results in

```
ERROR: elephant 0.6.3 has requirement neo<0.8.0,>=0.7.1, but you'll have neo 0.8.0 which is incompatible.
Installing collected packages: numpy, quantities, neo, scipy, six, elephant
Successfully installed elephant-0.6.3 neo-0.8.0 numpy-1.17.2 quantities-0.12.3 scipy-1.3.1 six-1.12.0
```

Although pip printed an error, the installation completes, and the packages seem to work together.

However, continuous integration tests for projects that use Neo and elephant can be tripped up by this error. For example, my [neurotic](https://github.com/jpgill86/neurotic) project's tests are failing when previously they were passing because of this pip install error, and the only change was the release of Neo 0.8.0: [Travis CI build report](https://travis-ci.com/jpgill86/neurotic/jobs/228366279#L1467)

I could pin my project's requirements to the old version of Neo to get around this, but there are features in 0.8.0 I'd like to use, and this might not even be an option for other projects.

Are there plans for the near future to release a version of elephant with the Neo entry in ``requirements.txt`` bumped to allow version 0.8.0? Are there major obstacles to full compatibility with Neo 0.8.0 that I might be able to help with?

dizcza:
Hi Jeffrey,

I'm currently looking at switching elephant to neo v0.8.0. There is one bug, however, reproduced several times in `test_neo_tools.py` that I don't understand - some breaking changes that, unfortunately, are not documented in neo v0.8.0 release notes. Now, `fake_neo('Block', seed=0)` returns a Block whose `array_annotations` field is an empty dict. Here is the traceback

```python
ERROR: test__extract_neo_attrs__epoch_noparents_array (elephant.test.test_neo_tools.ExtractNeoAttrsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ulianych/PycharmProjects/elephant/elephant/test/test_neo_tools.py", line 649, in test__extract_neo_attrs__epoch_noparents_array
    'durations': array_annotation_save['durations'],
KeyError: 'durations'
```

[...]

PR #256, opened 2019-10-09
neo v0.8.0 compatible

dizcza:
Fixes https://github.com/NeuralEnsemble/elephant/issues/253


elephant/neo_tools.py
```diff
@@ -45,17 +45,14 @@ def extract_neo_attrs(obj, parents=True, child_first=True,
...
-                # These would be duplicate
-                if a not in [_[0] for _ in obj._necessary_attrs + obj._recommended_attrs]:
-                    if "array_annotations" not in attrs:
-                        attrs["array_annotations"] = {}
-                    attrs["array_annotations"][a] = obj.array_annotations[a].copy()
-        except AttributeError:
-            pass
+    if not skip_array and hasattr(obj, "array_annotations"):
+        # Exclude labels and durations (and maybe other attributes) that are
+        # handled as array_annotations. These would be duplicate.
```
jpgill86: [According to](https://github.com/NeuralEnsemble/elephant/issues/253#issuecomment-539903686) @apdavison, epoch labels and durations were not assimilated into array_annotations with Neo 0.8.0, so this commend should be updated.

dizcza: Updated the comment

jpgill86:
> Need to wait till [NeuralEnsemble/python-neo#746](https://github.com/NeuralEnsemble/python-neo/issues/746) is resolved and released with a new version.

I'm not quite sure how the NeuralEnsemble/python-neo#746 impinges on elephant.

In that issue, you mention functions called "`cut_segment_by_epoch`" and "`seg_time_slice`". Are these elephant functions that are intended to be dropped in the next elephant release because their job is now handled in Neo 0.8.0, but because the performance is unsatisfactory you don't want to drop elephant's implementations yet? I'm having trouble finding them in elephant.

Sorry I'm not understanding!

dizcza:
Some Elephant users have developed and since then have been using these functions that now became a part of neo. But since I know them, I can mention this behavior and caution all other users in the following Elephant release.
As you probably saw, it's not only about `cut_segment_by_epoch` and `seg_time_slice`. And this is only one example I ran, I'm not sure how other examples are affected. But true, it should not harden the development and it's better to make minor releases often than waiting for a change.

Let's merge it.

dizcza:
Added todos for the fix of https://github.com/NeuralEnsemble/python-neo/issues/749

PR #258, opened 2019-10-18
keep array_annotations in the output of signal_processing functions

rgutzen:
In the function signal processing function _zscore_,  _butter_, and _hilbert_ the output is created by using the Neo functionality _duplicate_with_new_data_. However, this function purposely ignores array_annotations. Quote from the source code: 
 ```
        # Note: Array annotations are not copied here, because it is not ensured
        # that the same number of signals is used and they would possibly make no sense
        # when combined with another signal
```
In the case of the above functions, the array_annotations are still valid for the output, therefore, this PR copies them manually to the output.

### DIFF
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -126,6 +126,7 @@ def zscore(signal, inplace=True):
             sig_normalized = sig
         else:
             sig_normalized = sig.duplicate_with_new_data(sig_normalized)
+            sig_normalized.array_annotate(**sig.array_annotations)
         sig_dimless = sig_normalized / sig.units
         result.append(sig_dimless)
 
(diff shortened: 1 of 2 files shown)


dizcza:
Thanks for catching it. Currently, our test objects have empty array_annotations. It makes sense then to add a test (or rather update existing ones) that checks if array_annotations are preserved.

Anyway, I'm not sure that copying array_annotations should be handled by user, not neo. I'm not sure because I don't know all possible cases of array_annotations, whether they all have a length, etc. We'll check this at one of our Elephant & Co meetings.

mdenker:
Discussed in code meeting that the copying of array annotations should become a flag in the `duplicate_with_new_data()` function in Neo. For now, this fix is ok. Default behavior of the Neo function would be to not copy the array annotations, so there should not be a problem.

dizcza:
Relates to https://github.com/NeuralEnsemble/python-neo/issues/752

Issue #163, opened 2018-05-22
Integration of optimized ASSET


alperyeg:
Add the optimized ASSET module to Elephant. Atm the code lies in a private repository ([here](https://github.com/INM-6/pcp_use_cases/tree/clean_up)).
TODOs (more to add):

- [ ] Compare and merge `asset.py` with the code in the module
- [ ] Create an `asset` folder and put the necessary files in there. The module requires `cython` and other `c-based` files
- [ ] Adapt `setup.py` to compile the `c` files and include corresponding folder

dizcza:
Integrated in the mentioned pull reques,t however, we stripped off Cython.

PR #259, opened 2019-10-18
[WIP] Integrate optimised ASSET

Kleinjohann:
Addresses #163 

I've been using an adapted version based on [this one](https://github.com/INM-6/pcp_use_cases/tree/clean_up) mentioned in the issue for a while. Creating this PR now to avoid duplication and help with confusion over the multiple ASSET versions. 

There are still lots of TODOs left, however most of them are just ideas for possible performance improvements which may or may not yield significant speed-ups. 

I got rid of the cython part in order not to overload the requirements, it was only a minor performance improvement. As for MPI, the current version works both with and without it. 

Open for feedback, suggestions, ...


dizcza:
The tests are failing with the current impl and I don't know the correct behavior. Here is what I did so far:

1. reverted `_xrange` back to `range`. We still support python2, but computationally we should make an effort for py3. The overhead of creating immediate lists with `range` in python2 is negligible, compared to what ASSET (any elephant function) does.
2. removed `_sample_quantiles()`, `_sample_pvalue()` and `_time_slice()` functions since they are not used.
3. covered the docs for the first 10% of functions. Then I lost the understanding of what is going on. I'm not sure if we need to go through all functions again or just look at those that you've changed in this PR.

PR #260, opened 2019-10-22
Improved doc string of covariance() and corrcoef()

PaulinaDabrowska:
In the docs of spike_train_correlation.covariance() and ...corrcoef() vectors `b_i` and `b_j` were erroneously referred to as binary, which is now corrected to binned spike trains. I also added the definition of _l_ used in the formula for covariance.


dizcza:
Yes, it makes more sense now.
Thank you.

PR #261, opened 2019-10-24
Homogeneous Poisson Process with refr. period

pbouss:


elephant/spike_train_generation.py
```diff
@@ -1060,5 +1063,77 @@ def compound_poisson_process(rate, A, t_stop, shift=None, t_start=0 * ms):
...
+
+    rate = rate.rescale(1 / t_start.units).magnitude
+    duration = (t_stop-t_start).magnitude
+    mean_spike_count = rate * duration
+    spike_count = np.random.poisson(lam=mean_spike_count)
+
+    refr_period = refr_period.rescale(t_start.units).magnitude
+    eff_duration = duration - (spike_count - 1) * refr_period
+    if eff_duration <= 0.:
+        raise ValueError('The duration is too short to place this number of ' +
```
dizcza: I don't think we should raise the error here.

pbouss: The point is that there is an inherently limit of the firing rate, as 1/ refractory period. For example with a refractory period of 3 ms, a maximal firing rate of 333 Hz, would mean that there is just one spike every 3 ms. A higher rate can't be achieved with this refr. period. 

But this the number_of_spikes should follow a Poisson distribution, and can't just raise the error directly as a comparison of rate with the inverse of the refr. period. 

What I could do is only raise a warning and return a spiketrain with equally spaces spikes each separated by the refractory period.

What do you think?

pbouss: maybe it is better too state the error as such:
raise ValueError('The fire rate must be sufficiently smaller than the inverse of the refractory period, otherwise the spikes will merely follow each other equidistantly. ')

elephant/spike_train_generation.py
```diff
@@ -1060,5 +1064,86 @@ def compound_poisson_process(rate, A, t_stop, shift=None, t_start=0 * ms):
...
+    All numerical values should be given as Quantities, e.g. 100*Hz.
+
+    Parameters
+    ----------
+
+    rate : Quantity scalar with dimension 1/time
+           The rate of the discharge.
+    refr_period : Quantity scalar with dimension time
+                  The time period the after one spike no other spike is
+                  emitted.
```
stellalessandra: I think you should specify everywhere the default parameters. Btw: is 3ms appropriate?

elephant/spike_train_generation.py
```diff
@@ -1060,5 +1064,86 @@ def compound_poisson_process(rate, A, t_stop, shift=None, t_start=0 * ms):
...
+    duration = (t_stop - t_start).magnitude
+    mean_spike_count = rate_mag * duration
+    spike_count = np.random.poisson(lam=mean_spike_count)
+
+    # Check that the number of spikes drawn from the Poisson distributions,
+    # can fit in the duration regarding the refractory period.
+    if spike_count >= 1. + duration / refr_period_mag:
+        spike_count = np.ceil(duration / refr_period_mag).astype(int)
+
+    eff_duration = duration - (spike_count - 1) * refr_period_mag
```
stellalessandra: maybe add a comment on what the effective duration is, or the logic of your procedure

elephant/spike_train_generation.py
stellalessandra: @dizcza I agree with @pbouss actually. I think there is an upper limit for the rate. Maybe we do not understand what you mean?
@pbouss I would rewrite the error message though, the sentence is a bit complicated

elephant/spike_train_generation.py
```diff
@@ -1060,5 +1064,91 @@ def compound_poisson_process(rate, A, t_stop, shift=None, t_start=0 * ms):
...
+
+    st = np.sort(np.random.random(spike_count)) * eff_duration
```
dizcza: I don't think that sorting is necessary here.
Otherwise, you implicitly make each interspike interval longer than the previous, which is not a feature of randomness.

pbouss: No, I just make each inter-spike interval positive...

elephant/spike_train_generation.py
pbouss: @stellalessandra I rewrote the error message. How do you think, does it fit now?

elephant/spike_train_generation.py
```diff
@@ -1128,33 +1144,33 @@ def homogeneous_poisson_process_with_refr_period(rate,
...
-    # shortened by (spike-count - 1) times the refr. period.
-    eff_duration = duration - (spike_count - 1) * refr_period_mag
-
-    # In this effective time interval each spike is placed uniformly, and after sorting to each ISI
-    # one refractory period is added.
-    st = np.sort(np.random.random(spike_count)) * eff_duration
-    st += refr_period_mag * np.arange(spike_count)
-    st += t_start.magnitude
+    if duration <= refr_period:
+        st = []
```
pbouss: I would here also return a empty Spiketrain or an empty numpy-array. That it is consistent with the other output.

elephant/spike_train_generation.py
dizcza: It was hard for me to grasp the first Peter's version that's why I asked for help. Now I get it. If the current implementation is just a rewrite of the old one, then I admit I was wrong.
Thank you!

elephant/spike_train_generation.py
dizcza: yeah, then it'd require an additional if-check

pbouss: And a the case, that the duration is smaller than the refractory_period. You could still get one spike inside the spiketrain isn't it?

dizcza: hm, true

PR #266, opened 2019-10-31
Calculate spike train auto-correlation time ('timescale')

AlexVanMeegen:
Hey!

This PR provides an implementation to calculate the auto-correlation time of a binned spike train. The definition of the auto-correlation time is ambiguous, here the definition used in [1, 2] is implemented because the square in the integral (see docstring) reduces the influence of noise due to finite observation times.

The implementation was joint work with @morales-gregorio.

Cheers,
Alex

[1] Wieland, S., Bernardi, D., Schwalger, T., & Lindner, B. (2015). Slow fluctuations in recurrent networks of spiking neurons. Physical Review E, 92(4), 040901.
[2] van Meegen, A., & van Albada, S. J. (2019). A Microscopic Theory of Intrinsic Timescales in Spiking Neural Networks. arXiv preprint arXiv:1909.01908.


AlexVanMeegen:
Mmm, should I turn the docstrings into raw strings to satisfy PEP8?

elephant/spike_train_correlation.py
```diff
@@ -749,3 +751,68 @@ def run_T(spiketrain):
...
+    [1] Wieland, S., Bernardi, D., Schwalger, T., & Lindner, B. (2015).
+        Slow fluctuations in recurrent networks of spiking neurons.
+        Physical Review E, 92(4), 040901.
+    """
+    T = binned_st.t_stop - binned_st.t_start
+    binsize = binned_st.binsize
+
+    rate = np.sum(binned_st.to_array()) / T
+
+    tau_max.units = binsize.units
```
dizcza: this is something we should not do

elephant/spike_train_correlation.py
```diff
@@ -749,3 +751,68 @@ def run_T(spiketrain):
...
+    tau_max_bins = int(np.round(tau_max/binsize))
```
dizcza: `np.round` is useless here

elephant/spike_train_correlation.py
```diff
@@ -749,3 +751,68 @@ def run_T(spiketrain):
...
+                                               window=cch_window)
+    # CCH is dimensionless. Should have dimension 1/time^2, thus 1/dt^2.
+    # Furthermore, a multiplicative factor dt/T arises from the convolution.
+    # Subtract the squared first moment to arrive at the correlation function.
+    corrfct = cch / binsize / T - rate**2
+    # Take only t > 0 values, in particular neglecting the delta peak.
+    corrfct_pos = corrfct.time_slice(binsize/2, corrfct.t_stop).flatten()
+
+    # Calculate the timescale using trapezoidal integration
+    integr = np.abs((corrfct_pos / corrfct_pos[0]).magnitude)**2
```
dizcza: During the unit test, `corrfct_pos` has the dimension of `1/(s*ms)`.
I guess we need to make it like `.simplified.magnitude`
`.simplified` brings the units to `1/s**2`

elephant/test/test_spike_train_correlation.py
```diff
@@ -638,5 +641,38 @@ def test_exist_alias(self):
...
+
+        timescale_num = []
+        for _ in range(50):
+            spikes = homogeneous_gamma_process(2, 2*nu, 0*pq.ms, T)
+            spikes_bin = BinnedSpikeTrain(spikes, binsize)
+            timescale_i = spike_train_timescale(spikes_bin, 10*timescale)
+            timescale_i.units = timescale.units
+            timescale_num.append(timescale_i.magnitude)
+        correct = np.isclose(timescale.magnitude, timescale_num, rtol=1e-1)
+        self.assertTrue(correct.mean() > 0.9)
```
dizcza: I didn't take a closer look in the first place,
this check is overcomplicated. do you assume it should be the same for 90% of cases?
please use `assert_array_almost_equal`

dizcza:
> Mmm, should I turn the docstrings into raw strings to satisfy PEP8?

Maybe, it requires two slashes: `\\tau_\\mathrm{max}`

elephant/spike_train_correlation.py
dizcza: or did you mean not taking the integer part but rather rounding to the nearest integer? if so, this behavior should be avoided - use `math.ceil`. no integer casting is required afterward.

elephant/spike_train_correlation.py
```diff
@@ -749,3 +751,68 @@ def run_T(spiketrain):
...
+      necessary to introduce a cutoff for the numerical integration - this
+      cutoff should be neither smaller than the true auto-correlation time
+      nor much bigger.
+    * The binsize of binned_st is another critical parameter as it defines the
+      discretisation of the integral :math:`d\\tau`. If it is too big, the
+      numerical approximation of the integral is inaccurate.
+
+    References
+    ----------
+    [1] Wieland, S., Bernardi, D., Schwalger, T., & Lindner, B. (2015).
```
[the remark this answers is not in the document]
AlexVanMeegen: It is in the paper, Eq. (6) - as is written in the docstring (line 759). Here is the doi in case you looked at the wrong paper: 10.1103/PhysRevE.92.040901

elephant/spike_train_correlation.py
AlexVanMeegen: please suggest an alternative

elephant/spike_train_correlation.py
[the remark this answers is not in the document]
dizcza: ah, true, i missed it!

elephant/spike_train_correlation.py
dizcza: remove the line

elephant/spike_train_correlation.py
AlexVanMeegen: thought it doesn't matter because it divide `corrfct_pos / corrfct_pos[0]` which should be dimensionless regardless from where we start - or are there intricacies in `quantities` that I don't know about?

elephant/spike_train_correlation.py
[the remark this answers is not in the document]
AlexVanMeegen: no prob

elephant/spike_train_correlation.py
AlexVanMeegen: yup, the division was the issue I tried to address. like this works. for the `.ceil` see my comment below.

elephant/spike_train_correlation.py
dizcza: true, it should not matter here

elephant/spike_train_correlation.py
AlexVanMeegen: the problem is something like this: `int(6.6e0/3.3e-2)` which yields `199` instead of `200` due to numerical inaccuracies in float division. `ceil` would fix this example, but not if the floating division slightly overestimates the result. the only solution I see is rounding first and then taking the integer.

AlexVanMeegen: `math.ceil(3.6/1.2e-3)` for example yields `3001`

elephant/spike_train_correlation.py
AlexVanMeegen: cool - can you resolve, then? thx!

AlexVanMeegen:
> Maybe, it requires two slashes: \\tau_\\mathrm{max}
The two slashes are causing PEP8 to complain. The only way to resolve this is to turn the strings into raw strings, that way I can get rid of the double slashes (it does not interpret them anymore as escape sequences).

elephant/spike_train_correlation.py
AlexVanMeegen: What about this?
```
    assert (tau_max/binsize).units == pq.dimensionless
    tau_max_bins = int(np.round((tau_max/binsize).simplified.magnitude))
```

elephant/spike_train_correlation.py
```diff
@@ -749,3 +751,68 @@ def run_T(spiketrain):
...
+        Slow fluctuations in recurrent networks of spiking neurons.
+        Physical Review E, 92(4), 040901.
+    """
+    T = binned_st.t_stop - binned_st.t_start
```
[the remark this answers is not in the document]
AlexVanMeegen: done

elephant/spike_train_correlation.py
```diff
@@ -749,3 +751,68 @@ def run_T(spiketrain):
...
+    binsize = binned_st.binsize
+
+    rate = np.sum(binned_st.to_array()) / T
```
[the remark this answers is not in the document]
AlexVanMeegen: done

elephant/test/test_spike_train_correlation.py
AlexVanMeegen: tried to address, please check

AlexVanMeegen:
Here you go, @dizcza, new things to check ;)

AlexVanMeegen:
Thanks for the review!

elephant/test/test_spike_train_correlation.py
```diff
@@ -638,5 +641,38 @@ def test_exist_alias(self):
...
+
+        timescale_num = []
+        for _ in range(10):
+            spikes = homogeneous_gamma_process(2, 2*nu, 0*pq.ms, T)
+            spikes_bin = BinnedSpikeTrain(spikes, binsize)
+            timescale_i = spike_train_timescale(spikes_bin, 10*timescale)
+            timescale_i.units = timescale.units
+            timescale_num.append(timescale_i.magnitude)
+        target = np.allclose(timescale.magnitude, timescale_num, rtol=2e-1)
+        self.assertTrue(target)
```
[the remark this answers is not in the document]
AlexVanMeegen: Good to know, thanks for the explanation

Issue #269, opened 2019-11-07
[Bug] Spike triggered phase extraction with heterogenous time units


mdenker:
**Describe the bug**
A bug occurred in `elephant..phase_analysis.spike_triggered_phase` when the spike train and the analog signal had different time units, and these time units contained compound units. This lead to an incorrect calucation of the time indices of where spikes are in the analog signal.

**Fix**
The culprit was an expression in the function that calculates a dimensionless Quantities value vom the various time values, such as `t_start`. However, without doing a `simplify` on this experssion, it may be dimensionless, but potentially something like "1/3000*dimensionless".

A fix is in progress.

PR #270, opened 2019-11-07
Correctly handle Compound Units in `elephant.phase_analysis.spike_triggered_phase`

mdenker:
This is a fix and corresponding regression test for #269.

The new unit test reproduces the behavior: if the spike train times (or other time quantities of the input data) contain a CompoundUnit, a division in the code could lead to a unit like
`1/30000*dimensionless`, i.e., one could not take the magnitude of the term to obtain the actual outcome of the division. Use of `simplified` solves the issue.

### DIFF
--- a/elephant/phase_analysis.py
+++ b/elephant/phase_analysis.py
@@ -108,7 +108,8 @@ def spike_triggered_phase(hilbert_transform, spiketrains, interpolate):
         # Find index into signal for each spike
         ind_at_spike = np.round(
             (spiketrain[sttimeind] - hilbert_transform[phase_i].t_start) /
-            hilbert_transform[phase_i].sampling_period).magnitude.astype(int)
+            hilbert_transform[phase_i].sampling_period). \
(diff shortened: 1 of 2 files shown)


mdenker:
For completeness, the error message resulting from the incorrect division was something like
`IndexError: index 2970000 is out of bounds for axis 0 with size 100000`

Fixes #269

mdenker:
[Successful test may also linked to issue #763 in python-neo]

dizcza:
Doesn't it mean that we should correct all `.magnitude` to `.simplify.magnitude` for all elephant functions?

mdenker:
Not sure if it's for all, indeed it could be an issue for many functions. Not sure if there's a shortcut...

dizcza:
I guess, there is no such a shortcut. But it's not a problem.
Yet, to be safe, I'll change all usages of `.magnitude` to `.simplified.magnitude` in the near future.

Issue #264, opened 2019-10-30
CCH normalization / y-offset


AlexVanMeegen:
Hey!

@morales-gregorio and I stumbled across the following issues when using `cross_correlation_histogram` to calculate the time-lagged auto-correlation function. These might not be actual issues but boil down to different conventions with their respective (dis-)advantages. Just want to raise it here to make you aware.

If I understand the `cross_correlation_histogram` function correctly, it calculates 
```
CCH_{s1,s2}[k] = sum_n s1[n+k] * s2[n]
```
where s1 and s2 are `BinnedSpikeTrain.to_array()`s. All the following points are based on this.

**Normalization (time)**
For me, a (binned) spiketrain has dimension 1/time such that it's integral equals the total number of spikes. Hence, I expect the CCH to have the dimension 1/time^2. However, in agreement with the above definition, it is dimensionless. This amounts to a factor `1/binsize**2` difference.

**Normalization (convolution)**
The `sum_n` in the CCH depends on the length of the binned spiketrains. Thus, the CCH of a realization of a stationary process that is twice as long as another realization leads to a CCH that differs by a factor of two. To avoid this, I think the CCH should be divided by the length of the binned spiketrains. This amounts to a factor `binsize/T` difference.

**y-offset**
For long times, the time-lagged correlation should (for reasonable processes) decay to zero. However, in the above definition it decays to the squared rate because s1 and s2 are not centered. This amounts to a subtraction of the product of the mean values of s1 and s2.

Taking all three points into account, I arrive at
```
CCH_new = CCH_old / binsize / T - rate**2
```
for the auto-correlation (for the cross-correlation the rate might of course be different).

Just to be clear: none of the above amounts to the question whether the variance should be divided out imho.

dizcza:
It took a long time to think about what you wrote and what we actually do with it in CCH.

As you saw, CCH function has `cross_corr_coef` flag. However, it does slightly different than your proposal. But the equation for computing the cross-correlogram for poisson spike trains seems to be established since 1982 and used in papers interchangeably, so we stick to the equation, which is given in our book which in turn refers to [Eggermont 1992a](https://www.physiology.org/doi/pdf/10.1152/jn.1992.68.4.1216).

But if you look closely at the equation and our implementation of `cross_corr_coef()`, you find differences. That makes me wonder if the implementation is correct at all.  It has to be correct though because we have a test to compare CCH + cross_corr_coef output with `corrcoef()` function.

[...]

PR #271, opened 2019-11-11
Feature timescale

AlexVanMeegen:
Hey!

This is a small correction to #266 - apologies, I should have been more careful there. In particular, I
* cleaned the imports in the test,
* made the test more relevant by using different units,
* added an assertion that `tau_max` is divisible by the binsize, and
* use the `cross_corr_coef=True` option of `cross_correlation_histogram` to make the implementation independent of the discussion #264.

Cheers,
Alex

PS: linking @morales-gregorio FYI


AlexVanMeegen:
>  By convention, we use ValueError instead of AssertionErrors that are meant for debugging
Thanks for explaining, that's always helpful.

PR #111, opened 2017-07-20
CCHcoeff

mvonpapen:
added option in **elephant.spike_train_correlation.cch** named **cross_corr_coeff** that allows to _normalize the CCH output to obtain a cross-correlation coefficient function_ according to [Eggermont (2010) Pair-correlation in the time and frequency domain](https://link.springer.com/chapter/10.1007%2F978-1-4419-5675-0_5), Ch.5 in: Analysis of parallel spike trains, eds: S. Gruen, S. Rotter, Equation (5.10).

added a unit test in test_cross_correlation_histogram of **test_spike_train_correlation.py** that compares the output of CChcoef to the one of elephant.spike_train_correlation.corrcoeff for several displacements tau.

Joint work with @rgutzen

def _cross_corr_coef(cch_result, binned_st1, binned_st2):
        # Normalizes the CCH to obtain the cross-correlation 
        # coefficient function ranging from -1 to 1
        N  = max(binned_st1.num_bins, binned_st2.num_bins)
        Nx = len(binned_st1.spike_indices[0])
        Ny = len(binned_st2.spike_indices[0])
        spmat = [binned_st1.to_sparse_array(), binned_st2.to_sparse_array()]
        bin_counts_unique = []
        for s in spmat:
            bin_counts_unique.append(s.data)
        ii = np.dot(bin_counts_unique[0], bin_counts_unique[0])
        jj = np.dot(bin_counts_unique[1], bin_counts_unique[1])
        rho_xy = (cch_result - Nx*Ny/N) / np.sqrt( (ii-Nx**2./N)*(jj-Ny**2./N) )
        return rho_xy

PR #273, opened 2019-11-15
Use fftconvolve instead of np.correlate in cross_correlation_histogram

dizcza:
1. CCH speed-up with FFT: `np.correlate` has the time complexity of `O(N)` while `fftconvolve` has just a constant. For small arrays `np.correlate` is preferred, but the main problem, of course, with large arrays (see picture).
2. Introduced `CrossCorrHist` class that encapsulates the "private" functions in the former implementation.
3. Added `get_num_of_spikes()` in `BinnedSpikeTrain`.
4. Vectorized `.spike_indices` property in `BinnedSpikeTrain`.
5. Added `binarize()` function in `BinnedSpikeTrain`.

Figure_1

### DIFF
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -458,7 +458,7 @@ class BinnedSpikeTrain(object):
     # t_start, num_bins, t_stop
     # t_start, bin_size, t_stop
     # t_stop, num_bins, binsize
-    # ==========================================================================
+    # =========================================================================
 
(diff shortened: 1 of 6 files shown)


dizcza:
Need to look at `corrcoef` Pearson correlation coefficient and check if it's correct.

dizcza:
#111 
Ask Robin

dizcza:
The implementation of `cross_corr_coef()` in #111 and hence here is correct.
I'll open a separate PR to address this.

PR #275, opened 2019-11-27
Feature/joint isi dithering

pbouss:
Integration of the Joint-ISI dithering method, written as a class.


elephant/spike_train_surrogates.py
```diff
@@ -450,9 +461,451 @@ def jitter_spikes(spiketrain, binsize, n=1):
...
+        limited to the lower side by the minimal ISI.
+        This can be necessary, if in the data there is a certain refractory
+        period, which would be destroyed by the convolution with the
+        2d-Gaussian function.
+        Default: True
+    refr_period:
+        Since this dither-method should conserve the refractory period,
+        It is internally calculated as the minimum of the value given here
+        and the least ISI in the spiketrain.
+        Default: 4.*pq.ms
```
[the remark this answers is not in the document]
pbouss: This is still a question. In the end, this value is only important for sparse spiketrains. It helps than conserving at least this distance between two spikes. 
I think an option would be to make it bigger, but I would not make it smaller.

elephant/conversion.py
```diff
@@ -833,6 +833,7 @@ def remove_stored_array(self):
         """
         Unlinks the matrix with counted time points from memory.
 
+    def binarize(self):
```
dizcza: remove this commented line

pbouss: Okay, sorry

PR #279, opened 2019-12-05
added function for spike width extraction

PaulinaDabrowska:


elephant/waveform_features.py
```diff
@@ -9,11 +9,41 @@
...
+    width : float
+        Width of a waveform expressed as a number of data points
+    '''
+    waveform = np.squeeze(waveform)
+    if np.ndim(waveform) != 1:
+        raise ValueError('Expected 1-dimensional waveform.')
+    if len(waveform) < 2:
+        raise ValueError('Too short waveform.')
+
+    min_border = int(len(waveform) * 3 / 4)
```
dizcza: make `3/4` as `cutoff` parameter

dizcza:
@shashwatsridhar please take a look. We started it to be simple, though maybe you also know or have experienced other implementations that cover cases that Paulina's implementation misses.

I'm interested in general how the complexity of waveform width extraction grows while trying to cover all experimental cases.

elephant/waveform_features.py
```diff
@@ -2,17 +2,63 @@
...
-import quantities as pq
-import neo
 import warnings
 
+import numpy as np
+
+
+def waveform_width(waveform, cutoff=0.75):
+    """
+    Calculate the width (through-to-peak TTP) of a waveform.
```
shashwatsridhar: Typo in the word "trough"

dizcza: a catchy one, thanks

PR #280, opened 2019-12-10
Feature/spade speed up

pbouss:
Speed up for SPADE:
- deleting erroneous while-loop
- inclusion of sparse representation of Binned Spike Trains


dizcza:
I think we can get rid of the `for col in range(window_idx, window_idx + winlen)` line in `_build_context()` function. But it requires thinking about how to do this, and I'm not sure if it brings benefits. In any case, if you plan to optimize it further, I suggest first checking out `line_profiler` pip package.

PR #283, opened 2019-12-17
Homog. Poiss. with refr. Period: Bug fixed

pbouss:
Bug fixed while dealing with the units of the refr. period


dizcza:
Posting your detailed message,
> As it is now, we compare one object without duration against one with a unit: refr_period. This gives an error, if refr_period does not have the same units as t_start.

PR #285, opened 2019-12-20
SPADE refactoring

pbouss:
Trivial reduction of number of nested blocks.

### DIFF
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -1498,10 +1498,12 @@ def approximate_stability(concepts, rel_matrix, n_subsets, delta=0, epsilon=0):
         stab_int /= min(n_subsets, 2 ** len(intent))
         output.append((intent, extent, stab_int, stab_ext))
 
-    recv_list = comm.gather(output, root=0)
-    if rank == 0:
-        for i in range(1, len(recv_list)):
-            output.extend(recv_list[i])
+    if size != 1:
+        recv_list = comm.gather(output, root=0)
+        if rank == 0:
+            for i in range(1, len(recv_list)):
+                output.extend(recv_list[i])
+
     return output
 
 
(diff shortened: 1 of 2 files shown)


elephant/spade.py
```diff
@@ -289,37 +291,27 @@ def spade(data, binsize, winlen, min_spikes=2, min_occ=2, max_spikes=None,
...
-            time_pvalue_spectrum))
-        # Storing pvalue spectrum
-        output['pvalue_spectrum'] = pv_spec
-    elif 0 < alpha < 1:
-        warnings.warn('0<alpha<1 but p-value spectrum has not been '
-                      'computed (n_surr==0)')
+
+    if output_format not in ['concepts', 'patterns']:
+        raise ValueError(
+            "The output_format value has to be one between"
```
dizcza: 'one between' is confusing
change to `has to be 'x', 'y', 'z',...` or `has to be one of ...`

elephant/spade.py
```diff
@@ -977,19 +971,19 @@ def _fast_fca(context, min_c=2, min_z=2, max_z=None,
...
-            for (z, c) in np.transpose(np.where(spec_matrix != 0)):
-                spectrum.append((z + 1, c + 1, int(spec_matrix[z, c])))
 
-        if report == '3d#':
-            for (z, c, l) in np.transpose(np.where(spec_matrix != 0)):
-                spectrum.append(
-                    (z + 1, c + 1, l, int(spec_matrix[z, c, l])))
-        del spec_matrix
-        return spectrum
+    del concepts
```
dizcza: [not now]
We need to discuss all these `del x` statements that, in general, should be avoided.
I bet all of them came from C++, where it makes sense. But not in python.

dizcza:
And fix PEP8.
Other than that, it looks strange but OK

PR #291, opened 2020-01-21
cross_correlation_histogram relaxed the constraint for t_start and t_stop

### LINKED ISSUE #272: [BUG] Valid mode in CCH function for different t_start and t_stop
Hi,
I have a probem with the window parameter "valid" in the function cross_correlation_histogram.

I think for this mode it is nessecary to give different long spiketrains, but then an Error is raised. Here one example code:
```
start = 0.0*s
end = 10.0*s
st2 = BinnedSpikeTrain(homogeneous_poisson_process(rate=10.0*Hz, t_start=0.0*s, t_stop=10.0*s), binsize = 10*ms)
st1 = BinnedSpikeTrain(homogeneous_poisson_process(rate=10.0*Hz, t_start=2.5*s, t_stop=7.5*s), binsize = 10*ms)

cch, bins = cross_correlation_histogram(st1, st2, window = "valid")
plt.bar(bins, np.concatenate(cch))
plt.show()
```

The Error message is:
```

  File "<ipython-input-10-92b03047740d>", line 1, in <module>
    runfile('/home/lewen/Dropbox/Masterarbeit/sandkasten.py', wdir='/home/lewen/Dropbox/Masterarbeit')

  File "/home/lewen/anaconda3/lib/python3.7/site-packages/spyder_kernels/customize/spydercustomize.py", line 827, in runfile
    execfile(filename, namespace)

  File "/home/lewen/anaconda3/lib/python3.7/site-packages/spyder_kernels/customize/spydercustomize.py", line 110, in execfile
    exec(compile(f.read(), filename, 'exec'), namespace)

  File "/home/lewen/Dropbox/Masterarbeit/sandkasten.py", line 29, in <module>
    cch, bins = cross_correlation_histogram(st1, st2, window = "valid")

  File "/home/lewen/anaconda3/lib/python3.7/site-packages/elephant/spike_train_correlation.py", line 543, in cross_correlation_histogram
    raise AssertionError("Spike train must have same t start")

AssertionError: Spike train must have same t start
```

Expected behavior:
That the CCH is calculated like in numpy.correlate and the lags are given with quantities. Lag 0 would be the value if the smaller spiketrain is centered within the bigger one. And the difference in the starting times are ignored.

dizcza:
Fixes #272 

Previously, binned spiketrains are required to have the same num of bins, same t_start, and t_stop. Under these conditions, `valid` mode outputs only 1 value.

1. Jan implemented `valid` mode for spiketrains with different `t_start` and `t_stop`. I extended this idea to cover all cases of `window` parameter.
2. Jan added support for `memory` mode calculation of CCH with `valid` mode.
3. I made `CrossCorrHist` class protected by prepending an underscore so that `cross_correlation_histogram` is the only one public function, doing this.


dizcza:
That's also a good opportunity to discuss why the arguments are swapped:
`elephant.cch(x, y) == np.correlate(y, x)`.

elephant/spike_train_correlation.py
```diff
@@ -588,64 +600,87 @@ def cross_correlation_histogram(
...
+                                 window=(left_edge, right_edge))
     if method == 'memory':
-        cross_corr = cch_builder.correlate_memory()
+        cross_corr = cch_builder.correlate_memory(cch_mode=cch_mode)
     else:
         cross_corr = cch_builder.correlate_speed(cch_mode=cch_mode)
-    bin_ids = np.arange(left_edge, right_edge + 1)
+
     if border_correction:
         cross_corr = cch_builder.border_correction(cross_corr)
```
Kleinjohann: If I understand this correctly the border correction does not make any sense in the case `window='valid'`. If everything is calculated at full overlap then the same amount of bins is being used for each calculation, so there is no bin loss at the edges. In fact, a correct first order border correction should be zero in this case, so I suspect there is an issue in the calculation of the correction. 
I think I know what is wrong, but in PR reviews on github it is not possible to comment on lines untouched by the PR. Would you prefer to fix that here or rather open a new PR?
In any case there should probably be a warning that setting both `window='valid'` and `border_correction=True` does not make sense.

dizcza: > a correct first order border correction should be zero in this case, so I suspect there is an issue in the calculation of the correction.

I agree, this is also my thought. Till the issue is fixed, I decided to ignore border correction when window mode is `valid`. But let's better create a separate fix for this.

Apart from border correction, have you found any other issues with this PR? I wanted someone to take a look (not a closer one, just conceptional) before merging.

Kleinjohann: Alright! Haven't found any other issues, looks good!

PR #292, opened 2020-01-24
Homogen_poisson_with_refractory_period

pbouss:
- calculates these function now from the ISI-distribution
- changed other easily seen things.

### DIFF
--- a/elephant/spike_train_generation.py
+++ b/elephant/spike_train_generation.py
@@ -1153,8 +1153,8 @@ def homogeneous_poisson_process_with_refractory_period(
     rate = rate.rescale(1 / t_start.units)
     rate_magnitude = rate.magnitude
 
-    refractory_period_magnitude = \
-        refractory_period.rescale(t_start.units).magnitude
+    refractory_period = refractory_period.rescale(t_start.units)
+    refractory_period_magnitude = refractory_period.magnitude
     effective_rate_magnitude = \
         rate_magnitude / (1. - rate_magnitude*refractory_period_magnitude)
 
@@ -1162,8 +1162,12 @@ def homogeneous_poisson_process_with_refractory_period(
         return refractory_period_magnitude +\
                np.random.exponential(1./effective_rate_magnitude, size)
 
-    return _homogeneous_process(isi_generator, (), rate,
-                                t_start, t_stop, as_array)
+    spiketrain = _homogeneous_process(isi_generator, (), rate,
+                                      t_start-refractory_period, t_stop,
+                                      as_array)
+    if not as_array:
+        spiketrain.t_start = t_start
+    return spiketrain
 
 
 # Alias for the homogeneous poisson process with refractory period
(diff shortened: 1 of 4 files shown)


elephant/spike_train_generation.py
```diff
@@ -150,9 +146,13 @@ def threshold_detection(signal, threshold=0.0 * mV, sign='above'):
 
     assert threshold is not None, "A threshold must be provided"
 
-    if sign is 'above':
+    if sign not in ('above', 'below'):
+        raise AttributeError("sign should be 'above' or 'below'")
```
dizcza: please change to ValueError

elephant/spike_train_generation.py
```diff
@@ -204,14 +204,19 @@ def peak_detection(signal, threshold=0.0 * mV, sign='above', format=None):
     """
     assert threshold is not None, "A threshold must be provided"
 
-    if sign is 'above':
+    if sign not in ('above', 'below'):
+        raise AttributeError("sign should be 'above' or 'below'")
+
+    if format is not None and format != 'raw':
+        raise AttributeError("Format argument must be None or 'raw'")
```
dizcza: 1. change all asserts and AttributionErros to ValueError
2. did you get AttributeError from our examples styles_guide.rst?
3. better to write `if format not in (None, 'raw')` - the same syntax as you did above;
4. actually, the current version is totally OK, because in the example below the block is small and it's up to a developer to either
```
    if sign is 'above':
        cutout = np.where(signal > threshold)[0]
        peak_func = np.argmax
    elif sign in 'below':
        cutout = np.where(signal < threshold)[0]
        peak_func = np.argmin
    else:
        raise ValueError("sign must be 'above' or 'below'")
```

or

```
if sign not in ('above', 'below'):
        raise ValueError("sign should be 'above' or 'below'")

if sign == 'above':
        cutout = np.where(signal > threshold)[0]	
        peak_func = np.argmax
else:
        cutout = np.where(signal < threshold)[0]
        peak_func = np.argmin
```

The second approach is better in cases where we have long if-blocks.

we should specify this in our guide as well.

elephant/spike_train_generation.py
```diff
@@ -316,33 +321,89 @@ def homogeneous_poisson_process(rate, t_start=0.0 * ms, t_stop=1000.0 * ms,
...
+              Default: 0 ms
+    t_stop : Quantity scalar with dimension time, optional
              The end of the spike train.
-    as_array : bool
+             Default: 1000 ms
+    as_array : bool, optional
                If True, a NumPy array of sorted spikes is returned,
                rather than a SpikeTrain object.
+               Default: False
+    refractory_period : None or Quantity, optional
```
dizcza: it should be `pq.Quantity` for all arguments, actually. Refer to https://github.com/NeuralEnsemble/elephant/blob/master/elephant/unitary_event_analysis.py#L679 for example

PR #293, opened 2020-01-28
deprecation: pandas bridge will be removed

dizcza:
pandas_bridge module will be removed in Elephant v0.8.x
since nobody I've talked to so far has been using pandas bridge + it misses documentation + it's not used at all in Elephant + ... it's convoluted.


dizcza:
Need to bump up to v0.8.0

PR #295, opened 2020-01-31
added refractory period to inhomogeneous PP

pbouss:
added refractory period integration by thinning algorithm
further Code style improvements


elephant/spike_train_generation.py
```diff
@@ -461,13 +482,50 @@ def inhomogeneous_poisson_process(rate, as_array=False):
...
-    spikes = homogeneous_poiss[u < rate_interpolated.flatten()]
-    if as_array:
-        spikes = spikes.magnitude
+    random_uniforms = np.random.uniform(size=len(homogeneous_poiss)) * max_rate
+    spikes = homogeneous_poiss[random_uniforms < rate_interpolated.flatten()]
+
+    if refractory_period is not None:
+        refractory_period = refractory_period.rescale(
+            rate.t_stop.units).magnitude
+        spikes = _thinning_for_refractory_period(spikes.magnitude,
```
dizcza: doing this reduces the expected firing rate
are you sure this is the right way to include a refractory period in inhomogeneous PP?

elephant/spike_train_generation.py
```diff
@@ -613,11 +672,12 @@ def _n_poisson(rate, t_stop, t_start=0.0 * pq.ms, n=1):
...
-    # Check that the provided input is Hertz of return error
+    # Check that the provided input is Hertz
+    if not isinstance(rate, pq.Quantity):
+        raise ValueError('rate must be a pq.Quantity')
     try:
-        for r in rate.reshape(-1, 1):
-            r.rescale('Hz')
-    except AttributeError:
+        rate.rescale(pq.Hz)
+    except TypeError:
```
dizcza: should be ValueError as well

elephant/spike_train_generation.py
```diff
@@ -746,8 +805,7 @@ def single_interaction_process(
         if rate < 0 * pq.Hz:
             raise ValueError(
                 'rate (=%s) must be non-negative.' % str(rate))
-        rates_b = np.array(
-            [rate.magnitude for _ in range(n)]) * rate.units
+        rates_b = np.array([rate.magnitude] * n) * rate.units
```
dizcza: use np.repeat

elephant/spike_train_generation.py
```diff
@@ -775,16 +833,15 @@ def single_interaction_process(
     # Generate the array of times for coincident events in SIP, not closer than
     # min_delay. The array is generated as a pq.Quantity.
     if coincidences == 'deterministic':
-        n_coincidences = int(
-            ((t_stop - t_start) * rate_c).rescale(pq.dimensionless))
+        n_coincidences = int(round(
```
dizcza: why did you add round?

elephant/spike_train_generation.py
```diff
@@ -1004,55 +1063,52 @@ def _cpp_hom_stat(A, t_stop, rate, t_start=0 * pq.ms):
...
-            times[train_id] = mother[row].view(pq.Quantity)
+            spiketrains[train_id] = mother[row].view(pq.Quantity)
 
     except MemoryError:  # Slower (~2x) but less memory-consuming approach
         print('memory case')
-        times = [[]] * N
-        for t, l in zip(mother, labels):
-            train_ids = random.sample(range(N), l)
+        for mother_spiketrain, label in zip(mother, labels):
+            train_ids = random.sample(range(n_spiketrains), label)
```
dizcza: maybe we could use fast np.random routines?
google it

elephant/spike_train_generation.py
```diff
@@ -1066,33 +1122,44 @@ def _cpp_het_stat(A, t_stop, rate, t_start=0. * pq.ms):
...
-    N = len(rate)  # number of output spike trains
-    A_exp = np.dot(A, range(N + 1))  # expectation of A
-    r_sum = np.sum(rate)  # sum of all output firing rates
-    r_min = np.min(rate)  # minimum of the firing rates
-    r1 = r_sum - N * r_min  # rate of the uncorrelated CPP
-    r2 = r_sum / float(A_exp) - r1  # rate of the correlated CPP
-    r_mother = r1 + r2  # rate of the hidden mother process
+    n_spiketrains = len(rates)  # number of output spike trains
+    # amplitude expectation
+    expected_amplitude = np.dot(amplitudes, range(n_spiketrains + 1))
```
dizcza: range is generator
it's better to use np.arange when it's been a part of math operation

elephant/spike_train_generation.py
```diff
@@ -1118,7 +1185,7 @@ def compound_poisson_process(rate, A, t_stop, shift=None, t_start=0 * pq.ms):
           - a single value, all spike trains will have same rate rate
           - an array of values (of length len(A)-1), each indicating the
             firing rate of one process in output
-    A : np.ndarray
+    amplitudes : np.ndarray
```
dizcza: don't rename parameters in public functions!
revert

elephant/spike_train_generation.py
pbouss: It is maybe a bit confusing, but I first updated the firing rate to an effective firing rate. In line 471. And then I thin it out. In average both effects cancel to result in the wished average firing rate.

elephant/spike_train_generation.py
pbouss: True

elephant/spike_train_generation.py
pbouss: I tried, but then some strange error occurring. In the following for loops. Thus I capitulated.

elephant/spike_train_generation.py
pbouss: because if in average I expect 80.99 coincidences. It is more normal to use 81 than 80.

elephant/spike_train_generation.py
pbouss: In test_statistics.py the spike-count is checked against its expectation. It works well.

dizcza:
@p-bouss one SPADE test is faling - test_pattern_set_reduction

elephant/spike_train_generation.py
dizcza: While it works for the scenario you encountered, does it makes sense to round for all possible cases in the Universe of how this function can be called?

pbouss: Yes, since we want the closest approximation to the average coincidence count.

dizcza:
It became a huge module refactoring.
The main part, inhomogeneous PP with reft period, however, is broken.
I added a test that checks the effective firing rate and mean ISI with the expected valeus for both homogeneous and inhomogeneous processes. The test is called `test_effective_rate_refractory_period` and it fails for inhomogeneous. Since it's the title of the PR, please fix your implementation and make the test passing (as in the case with homogeneous PP).

We can sit and think together or you can do it by yourself. Waiting.

elephant/spike_train_generation.py
```diff
@@ -516,13 +513,9 @@ def _thinning_for_refractory_period(spiketrain, refractory_period):
...
         thinned out spiketrain
     """
-    thinned_spiketrain = []
-    previous_spike = -refractory_period
-    for spike in spiketrain:
-        if spike > previous_spike + refractory_period:
-            thinned_spiketrain.append(spike)
-            previous_spike = spike
-    return np.array(thinned_spiketrain)
+    mask_valid = np.diff(spiketrain, prepend=0) > refractory_period
```
pbouss: It is this mask_valid, that does not work. I mean the for loop does not look nice, I know. But it can provide, compare only with the spikes that are actually taken. 
But now, let's say, e.g. the firing rate is very high, at the spiketrain every ISI is smaller than the refractory period. In this mask, say would be all False. And thus all spikes would be discarded...
So I would return to the old version until a very great idea arises on how to do it with numpy.

PR #297, opened 2020-02-04
added uniform dithering method with refr period

pbouss:
+ uniform dithering with the conservation of refr. period
+ added unit tests for this function
+ some code style issues


elephant/spike_train_surrogates.py
```diff
@@ -144,13 +140,95 @@ def dither_spikes(spiketrain, dither, n=1, decimals=None, edges=True):
...
+        Default: 4.*pq.ms
+
+    Returns
+    -------
+    dithered_spiketrains : list of neo.SpikeTrain
+        Each element is an independent surrogate spiketrain of 'spiketrain',
+        where each spike is moved.
+    """
+    if len(spiketrain) == 0:
+        return [spiketrain] * n_surrogates
```
dizcza: this is not a copy!
I fixed it.

elephant/spike_train_surrogates.py
```diff
@@ -144,13 +140,95 @@ def dither_spikes(spiketrain, dither, n=1, decimals=None, edges=True):
...
+                if random_id < len(spiketrain) - 1 \
+                else t_stop + refractory_period
+
+            previous_dither = min(dither,
+                                  spike - previous_spike - refractory_period)
+            next_dither = min(dither,
+                              next_spike - spike - refractory_period)
+
+            dithered_spiketrain[random_id] += (previous_dither + next_dither)\
+                * random.random() - previous_dither
```
dizcza: I'm sure we can vectorize the dithering and get rid of the internal for-loop. Try adding a random uniform vector, things like that.
Think of how to do this. We can think together.

elephant/spike_train_surrogates.py
```diff
@@ -147,13 +143,95 @@ def dither_spikes(spiketrain, dither, n=1, decimals=None, edges=True):
...
 
-    # Return the surrogates as SpikeTrains
+    # Return the surrogates as list of neo.SpikeTrain
     return [neo.SpikeTrain(s,
                            t_start=spiketrain.t_start,
                            t_stop=spiketrain.t_stop).rescale(spiketrain.units)
             for s in surr]
 
 
+def dither_spikes_with_refractory_period(spiketrain, dither, n_surrogates=1,
```
dizcza: maybe it makes sense to add `refractory_period` argument to `dither_spikes()` function as you eventually did with homogeneous Poisson generation?

pbouss: Yes, I will think about how to do this or to give it another name, it's anyway lengthy.

elephant/spike_train_surrogates.py
```diff
@@ -147,13 +143,95 @@ def dither_spikes(spiketrain, dither, n=1, decimals=None, edges=True):
...
+                if random_id > 0 \
+                else t_start - refractory_period
+            next_spike = dithered_spiketrain[random_id + 1] \
+                if random_id < len(spiketrain) - 1 \
+                else t_stop + refractory_period
+
+            previous_dither = min(dither,
+                                  spike - previous_spike - refractory_period)
+            next_dither = min(dither,
+                              next_spike - spike - refractory_period)
```
dizcza: are you sure it should not be like 
`min(dither, next_spike - spike + refractory_perioud)`?

elephant/spike_train_surrogates.py
dizcza: I found a way to vectorize this if the following logic is equivalent to
```python
dithered_spiketrain = np.copy(spiketrain.magnitude)
        for i, random_id in enumerate(random_ordered_ids):
            spike = spiketrain.magnitude[random_id]

            prev_spike = spiketrain.magnitude[random_id - 1] \
                if random_id > 0 \
                else t_start - refractory_period
            next_spike = spiketrain.magnitude[random_id + 1] \
                if random_id < len(spiketrain) - 1 \
                else t_stop + refractory_period

            prev_dither = min(dither,
                                  spike - prev_spike - refractory_period)
            next_dither = min(dither,
                              next_spike - spike - refractory_period)

            dt = (prev_dither + next_dither) * random_vals[i] - prev_dither
            dithered_spiketrain[random_id] += dt
```

but it breaks the test.
So do we need to use `dithered_spiketrain` instead of `spiketrain.magnitude` in the for-loop?

elephant/spike_train_surrogates.py
pbouss: Yes, I'm sure. This keeps the distance between the spike and the next_spike, to be greater than the refractory period.

elephant/spike_train_surrogates.py
pbouss: I think it breaks because prev_spike and next_spike needs to be taken off the updated spike train, i.e. of dithered_spiketrain.

pbouss: This is the logic 
`dithered_spiketrain = np.copy(spiketrain.magnitude)
        for i, random_id in enumerate(random_ordered_ids):
            spike = spiketrain.magnitude[random_id]
                 % or spike = dithered_spiketrain[random_id]
                 % both forms are equivalent here since every spike is only moved once

            prev_spike = dithered_spiketrain[random_id - 1] \
                if random_id > 0 \
                else t_start - refractory_period
            next_spike = dithered_spiketrain[random_id + 1] \
                if random_id < len(spiketrain) - 1 \
                else t_stop + refractory_period

            prev_dither = min(dither,
                                  spike - prev_spike - refractory_period)
            next_dither = min(dither,
                              next_spike - spike - refractory_period)

            dt = (prev_dither + next_dither) * random.random() - prev_dither
            dithered_spiketrain[random_id] += dt`

dizcza: OK, sad. Then there is no way to vectorize the for loop.

dizcza:
Then I'm waiting for a change when you merge `dither_spikes_with_refractory_period` with `dither_spikes()`.

Other than that, no further comments.

elephant/spike_train_surrogates.py
pbouss: Yes, that is was I thought.

elephant/spade.py
```diff
@@ -1265,6 +1265,10 @@ def pvalue_spectrum(spiketrains, binsize, winlen, dither, n_surr, min_spikes=2,
         if surr_method == 'joint_isi_dithering':
             surrs = [instance.dithering()[0] for
                      instance in joint_isi_instances]
+        elif surr_method == 'dither_spikes_with_refractory_period':
+            surrs = [surr.dither_spikes(
+                spiketrain, dither=dither, n=1, refractory_period=binsize)[0]
```
dizcza: `refractory_period=binsize` looks strange but who am I to question spade devs

pbouss: Since it is the initial refractory period, which then is corrected by the smallest ISI, I thought that it would be nice, to have the least value, that still keeps one spike per bin, for spike trains where anyway there are not two spikes close to each other.

stellalessandra: I also agree with Peter that this is an appropriate idea. In principle here our only worry is that the refractory period is at least one binsize, in order to prevent that we don't get spikes that end up very close to each other (in same bin).

elephant/spike_train_surrogates.py
```diff
@@ -46,25 +46,61 @@
...
+
+def _dither_spikes_with_refractory_period(spiketrain, dither, n,
+                                          refractory_period):
+    units = spiketrain.units
+    t_start = spiketrain.t_start.rescale(units).magnitude
+    t_stop = spiketrain.t_stop.rescale(units).magnitude
+
+    dither = dither.rescale(units).magnitude
+    refractory_period = refractory_period.rescale(units).magnitude
+    refractory_period = np.min(np.diff(spiketrain.magnitude),
```
stellalessandra: maybe add a comment here?

elephant/spike_train_surrogates.py
```diff
@@ -46,25 +46,61 @@
...
+    for _ in range(n):
+        dithered_st = np.copy(spiketrain.magnitude)
+        random_ordered_ids = np.arange(len(spiketrain))
+        np.random.shuffle(random_ordered_ids)
+
+        for random_id in random_ordered_ids:
+            spike = dithered_st[random_id]
+            prev_spike = dithered_st[random_id - 1] \
+                if random_id > 0 \
+                else t_start - refractory_period
```
stellalessandra: why do you subtract the refractory period for the first spike?

elephant/spike_train_surrogates.py
```diff
@@ -46,25 +46,61 @@
...
+            next_spike = dithered_st[random_id + 1] \
+                if random_id < len(spiketrain) - 1 \
+                else t_stop + refractory_period
+            prev_dither = min(dither, spike - prev_spike - refractory_period)
```
stellalessandra: I think that the word dither here is misleading...in principle it's just the space from your spike to the previous spike minus the refractory period, right? You are not actually dithering the previous spike if I understood correctly...

elephant/spade.py
dizcza: Peter, please put Alessandra's comment there.

elephant/spike_train_surrogates.py
dizcza: Isn't it already described in `dither_spikes()` docstring?

elephant/spike_train_surrogates.py
pbouss: With this I allow the first spike to move up to the beginning of the spike train. Note that this refractory period is then subtracted again.

elephant/spike_train_surrogates.py
pbouss: I added a comment in the code. 
I stands for the dither range in the direction of the previous spike. This is either the dither parameter itself or the distance between the spikes minus the refractory period.

elephant/spike_train_surrogates.py
pbouss: I added another comment there to explain it also in the code.

elephant/spade.py
pbouss: OK.

PR #303, opened 2020-03-06
[spectral] replaced _welch() by scipy.signal.welch()

dizcza:
Fixed the TODO which says

"This function (`_welch()`) should be replaced by `scipy.signal.csd()`, which will appear in SciPy 0.16.0."

1. Now it returns onesided fft array instead of two-sided (old behavior)
2. the phase is inverted - a conjugate transform to what was before. but both ways are "correct" and we decided to stick to the current scipy way of defining the phase

### DIFF
--- a/elephant/spectral.py
+++ b/elephant/spectral.py
@@ -422,10 +422,7 @@ def welch_cohere(x, y, num_seg=8, len_seg=None, freq_res=None, overlap=0.5,
     _, Pxy = scipy.signal.csd(xdata, ydata, **params)
 
     coherency = np.abs(Pxy) ** 2 / (Pxx * Pyy)
-
-    # conjugate is taken to follow the old behavior of _welch() function
-    # that was a copy-paste of scipy v0.16.0
-    phase_lag = np.angle(Pxy.conj())
+    phase_lag = np.angle(Pxy)
 
     # attach proper units to return values
     if isinstance(x, pq.quantity.Quantity):
--- a/elephant/spectral.py
+++ b/elephant/spectral.py
@@ -9,219 +9,10 @@ spectrum).
 
 from __future__ import division, print_function, unicode_literals
 
-import warnings
-
-import numpy as np
-import scipy.signal
-import scipy.fftpack as fftpack
-import scipy.signal.signaltools as signaltools
-from scipy.signal.windows import get_window
-from six import string_types
-import quantities as pq
 import neo
-
-
-def _welch(x, y, fs=1.0, window='hanning', nperseg=256, noverlap=None,
-           nfft=None, detrend='constant', scaling='density', axis=-1):
-    """
-    A helper function to estimate cross spectral density using Welch's method.
-
-    Welch's method [1]_ computes an estimate of the cross spectral density
-    by dividing the data into overlapping segments, computing a modified
-    periodogram for each segment and averaging the cross-periodograms.
-
-    Parameters
-    ----------
-    x : list or tuple or np.ndarray
-        Time series of measurement values for the first signal.
-        It will be converted to `np.ndarray`.
-    y : list or tuple or np.ndarray
-        Time series of measurement values for the second signal.
-        It will be converted to `np.ndarray`.
-    fs : float, optional
-        Sampling frequency of the `x` and `y` time series in units of Hz.
-        Default: 1.0.
-    window : str or tuple or np.ndarray, optional
-        Desired window to use.
-        If `window` is a string, see `scipy.signal.get_window` for a list of
-        windows and required parameters.
-        If `window` is tuple or `np.ndarray`, it will be used directly as the
-        window, and its length will be used for `nperseg`. If not string,
-        it will be treated as `np.ndarray`, and the result must be a vector
-        with length less than or equal to the length of the last axes of `x`
-        and `y`.
-        Default: 'hanning'.
-    nperseg : int, optional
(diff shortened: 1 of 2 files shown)


elephant/test/test_spectral.py
```diff
@@ -210,9 +211,9 @@ def test_welch_cohere_behavior(self):
         freqs_np, coherency_np, phase_lag_np =\
             elephant.spectral.welch_cohere(x.magnitude.flatten(), y.magnitude.flatten(),
                 fs=1/sampling_period, freq_res=freq_res)
-        self.assertTrue((freqs == freqs_np).all() and
-                        (coherency[:, 0] == coherency_np).all() and
-                        (phase_lag[:, 0] == phase_lag_np).all())
+        assert_array_almost_equal(freqs.simplified.magnitude, freqs_np)
+        assert_array_almost_equal(coherency[:, 0], coherency_np)
+        assert_array_almost_equal(phase_lag[:, 0], phase_lag_np)
```
Junji110: These assertions pass with the original test (using assertTrue) in my configuration.
Do they fail in other configurations?
If they do, what configurations?

elephant/test/test_spectral.py
```diff
@@ -223,9 +224,9 @@ def test_welch_cohere_behavior(self):
             elephant.spectral.welch_cohere(x_multidim, y_multidim)
         freqs_T, coherency_T, phase_lag_T =\
             elephant.spectral.welch_cohere(x_multidim.T, y_multidim.T, axis=0)
-        self.assertTrue(np.all(freqs==freqs_T))
-        self.assertTrue(np.all(coherency==coherency_T.T))
-        self.assertTrue(np.all(phase_lag==phase_lag_T.T))
+        assert_array_almost_equal(freqs, freqs_T)
+        assert_array_almost_equal(coherency, coherency_T.T)
+        assert_array_almost_equal(phase_lag, phase_lag_T.T)
```
Junji110: Here as well.

elephant/spectral.py
```diff
@@ -627,11 +417,15 @@ def welch_cohere(x, y, num_seg=8, len_seg=None, freq_res=None, overlap=0.5,
...
-    phase_lag = np.angle(Pxy)
+    freqs, Pxx = scipy.signal.welch(xdata, **params)
+    _, Pyy = scipy.signal.welch(ydata, **params)
+    _, Pxy = scipy.signal.csd(xdata, ydata, **params)
+
+    coherency = np.abs(Pxy) ** 2 / (Pxx * Pyy)
+
+    # conjugate is taken to follow the old behavior of _welch() function
+    # that was a copy-paste of scipy v0.16.0
+    phase_lag = np.angle(Pxy.conj())
```
Junji110: It seems that our original implementation of _welch() was wrong.
Please remove this conjugate.

dizcza: then we should change the ground truth of one of the tests
remove the conjugate and try running tests

elephant/test/test_spectral.py
dizcza: All numpy floating-point value arrays must be checked with the ground truth arrays by `assert_array_almost_equal()`. This is the numpy's testing convention and it simplifies the debugging because when the test fails it shows useful information about how many values differ, max absolute and relative difference instead of just
got False, expected True
message

dizcza: They failed once I started debugging and that's why I changed them to ones how it should be

Junji110: Then can't we use assert_array_equal?
These arrays should be exactly the same, as the computation inside the function should not depend on how the data were given.

dizcza: For example,
```python
    phase_lag_new = np.angle(Pxy.conj())

    from numpy.testing import assert_array_almost_equal
    assert_array_almost_equal(phase_lag, phase_lag_new)  # phase_lag is our old _welch()
```

```
AssertionError: 
Arrays are not almost equal to 6 decimals

Mismatched elements: 2 / 1000 (0.2%)
Max absolute difference: 6.28318531
Max relative difference: 2.
 x: array([[ 3.141593, -2.602966,  2.297019,  2.649172, -1.171436, -0.943932,
        -0.565363, -1.253926,  2.561917, -0.813005, -0.940448, -1.601776,
         2.712933,  2.158754,  0.957191,  1.645848, -0.283867, -1.849341,...
 y: array([[-3.141593, -2.602966,  2.297019,  2.649172, -1.171436, -0.943932,
        -0.565363, -1.253926,  2.561917, -0.813005, -0.940448, -1.601776,
         2.712933,  2.158754,  0.957191,  1.645848, -0.283867, -1.849341,...
```

this is actually how it behaves now - it does not much precisely even with `.conj()`

elephant/spectral.py
Junji110: I didn't find a test where the actual values of phase difference were tested (existing tests are always comparison between the results from different usages of the same function).
Where is it?

dizcza: fixed the ground truth

elephant/test/test_spectral.py
Junji110: Please use also here assert_array_equal().

dizcza: done

Junji110:
All fine.
I'm not authorized to merge, so please do that by yourself or ask someone else.

PR #307, opened 2020-03-26
Parallel module

dizcza:
* ProcessPoolExecutor
* MPIPoolExecutor
* MPICommExecutor


mdenker:
Please check in particular the doc and online tutorial at https://elephant-docs-development.readthedocs.io/en/feature-embarassingly_parallel/tutorials/parallel.html

elephant/parallel/mpi.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    The `MPIPoolExecutor` class uses a pool of MPI processes to execute calls
+    asynchronously.
+
+    `MPIPoolExecutor` is recommended to use on cluster nodes which support
+    MPI-2 protocol.
+
+    Notes
+    -----
+    `-m mpi4py.futures` command line option is needed to execute python scripts
```
JuliaSprenger: According to the code style guideline 'python' should be always capitalized 
```suggestion
    `-m mpi4py.futures` command line option is needed to execute Python scripts
```

elephant/parallel/mpi.py
```diff
@@ -0,0 +1,71 @@
...
+    with MPI:
+
+    .. code-block:: sh
+
+       mpiexec -n numprocs python -m mpi4py.futures pyfile [arg] ...
+
+    For more information of how to launch MPI processes in python refer to
```
JuliaSprenger:
```suggestion
    For more information of how to launch MPI processes in Python refer to
```

postBuild
```diff
@@ -1 +1,5 @@
-pip install .[tutorials]
\ No newline at end of file
+pip install .[tutorials]
+
+# Install conda version of mpi4py for mybinder service
+conda install mpi4py
```
JuliaSprenger: Does this imply that the tests are never run with the pip installed version of mpi4py?

setup.py
```diff
@@ -21,7 +21,8 @@
 with open("README.md") as f:
     long_description = f.read()
 with open('requirements/requirements.txt') as fp:
-    install_requires = fp.read()
+    install_requires = fp.read().splitlines()
+    install_requires.append("futures; python_version == '2.7'")
```
JuliaSprenger: Shouldn't this be part of the requirements.txt instead of being hard coded here?

JuliaSprenger:
Thanks for implementing the new parallelization feature!
In addition to the comments here I also made some suggestions for typos to correct in https://github.com/INM-6/elephant/pull/73

elephant/parallel/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+
+from .parallel import ProcessPoolExecutor, SingleProcess
+
+try:
+    from .mpi import MPIPoolExecutor, MPICommExecutor
+except ImportError:
+    # mpi4py is missing
+    # please run command `pip install -r requirements-extras.txt`
```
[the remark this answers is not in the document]
dizcza: I agree it's not clear.

1. If I print a message, then all other users will see the message "Cannot import blah-blah. Please install extras." while `import elephant`.

The only solution is 

2. still put there a print but remove `from . import parallel` from `__init__.py`. This will, however, disturb slightly our ecosystem - we've been always importing all modules in `__init__.py` and now we make an exception.

Anyway, I'd rather do 2.
Yes, nor?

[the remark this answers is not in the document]
dizcza: `pip install elephant[extras]` is definitely better.

doc/tutorials/parallel.ipynb
```diff
@@ -0,0 +1,364 @@
+{
```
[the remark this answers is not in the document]
dizcza: It should. But readthedocs hangs to execute MPI part of the tutorial even though mpi4py is installed. It's because of the intervention of two things: conda mpi4py package is somewhat broken (compared to linux libopenmpi package) and all the scripts that have mpi should be executed with mpiexec command.

postBuild
dizcza: pip install mpi4py requires libmpi linux package to be installed. This is the workaround for both readthedocs and mybinder.

setup.py
dizcza: Yes, a good catch.

doc/tutorials/parallel.ipynb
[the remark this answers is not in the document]
JuliaSprenger: and the pip requirement for mpi4py also doesn't work on readthedocs?

elephant/__init__.py
```diff
@@ -19,7 +19,6 @@
                current_source_density,
                change_point_detection,
                phase_analysis,
-               parallel,
```
JuliaSprenger: I think it would make sense to add a comment here, explaining why parallel is not autoimported, such that we don't try to 'fix' this in the future.

doc/tutorials/parallel.ipynb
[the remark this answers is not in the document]
dizcza: precisely

PR #313, opened 2020-04-21
Fix kernels and statistics (instantaneous_rate & mean_firing_rate)

### LINKED ISSUE #107: alpha kernel is not starting at zero
Contrary to the definition in the docs, the alpha kernel does not start with non-negative values only after `t = 0` in elephant `0.4.1`.

to demonstrate

```python
import neo
import elephant as ele
import scipy as sp
import matplotlib.pyplot as plt
import quantities as pq

st = neo.SpikeTrain([1]*pq.s,t_start=0*pq.s,t_stop=2*pq.s)
kernel = ele.kernels.AlphaKernel(200*pq.ms)
fs = 0.1 * pq.ms
asig = ele.statistics.instantaneous_rate(st,t_start=st.t_start,t_stop=st.t_stop,sampling_period=fs,kernel=kernel)
plt.plot(asig.times.rescale(pq.s),asig)
[plt.axvline(t) for t in st.times]
```
produces
alpha_kernel_offset

edit: which is almost the median (not exactly probably the kernel never decays back to zero). I looked at the `AlphaKernel` class in `kernels.py` and it looks fine, is there some median centering of the kernel going on somewhere else, that might need to be disabled?

### LINKED ISSUE #288: [Bug] Instantaneous rate estimation fails for inconsistent kernel selection
When providing a kernel on a much smaller time scale than sampling rate requested the instantaneous rate estimation will fail on numpy level with
`ValueError: zero-size array to reduction operation minimum which has no identity`

Here's a minimal failing example
```python
from elephant import kernels
spiketrain = neo.spiketrain_generation.homogeneous_poisson_process(10*Hz, t_start=array(0.) * ms, t_stop=array(10000.) * ms)
kernel = kernels.AlphaKernel(sigma = 5*ms, invert = True)
rate = instantaneous_rate(spiketrain, sampling_period = 200*ms, kernel=kernel)
```

I would expect this to be caught by the instantaneous rate and providing a more telling error message instead of this failing on a numpy level.

The full traceback is 
```
ValueError                                Traceback (most recent call last)
<ipython-input-167-5ac422ce9801> in <module>
      1 from elephant import kernels
      2 kernel = kernels.AlphaKernel(sigma = 5*ms, invert = True)
----> 3 rate = instantaneous_rate(spiketrain1, sampling_period = 200*ms, kernel=kernel)

~/anaconda3/envs/t/lib/python3.7/site-packages/elephant/statistics.py in instantaneous_rate(spiketrain, sampling_period, kernel, cutoff, t_start, t_stop, trim)
    743 
    744     if not trim:
--> 745         r = r[kernel.median_index(t_arr):-(kernel(t_arr).size -
    746                                            kernel.median_index(t_arr))]
    747     elif trim:

~/anaconda3/envs/t/lib/python3.7/site-packages/elephant/kernels.py in median_index(self, t)
    215         """
    216         return np.nonzero(self(t).cumsum() *
--> 217                           (t[len(t) - 1] - t[0]) / (len(t) - 1) >= 0.5)[0].min()
    218 
    219     def is_symmetric(self):

~/anaconda3/envs/t/lib/python3.7/site-packages/numpy/core/_methods.py in _amin(a, axis, out, keepdims, initial)
     30 def _amin(a, axis=None, out=None, keepdims=False,
     31           initial=_NoValue):
---> 32     return umr_minimum(a, axis, None, out, keepdims, initial)
     33 
     34 def _sum(a, axis=None, dtype=None, out=None, keepdims=False,

ValueError: zero-size array to reduction operation minimum which has no identity
```

dizcza:
Breaking changes:
* statistics:
  * when the input is a neo.SpikeTrain or a Quantity and t_start and/or t_stop are given, they (start, stop) should be of type Quantity (previously a float is accepted - error-prone);

Improvements:
* kernels:
  * fixed #288
  * get rid of the magic in `boundary_enclosing_area_fraction` function of the Kernel base class
  * fixed a slightly wrong estimate of the true `boundary_enclosing_area_fraction` function of AlphaKernel class --> more than x100 speedup for the AlphaKernel --> the runtime of the tests decreased from 7 sec to 70 ms;
  * fixed the wrong `median_index()` kernel function and added tests
  * added built-in matplotlib plots for each kernel in the docs - check-out [AplhaKernel](https://elephant-docs-development.readthedocs.io/en/fix-kernels/reference/toctree/kernels/elephant.kernels.AlphaKernel.html#elephant.kernels.AlphaKernel) readthedocs
* statistics:
  * vectorized binning in `instantaneous_rate()`
  * refactored `mean_firing_rate()`, see Breaking changes 

Not done/Unsure:
* [x] previously, `test_rate_estimation_consistency()` had a test with `trim=True`. Testing with `trim=True` does not make sense, does it? - Becuse the `x` axis is shorter (trimmed). Previously, it worked because of the incorrect estimation of `kernel.median_index()`. If I'm right, this is a perfect example of BUG x BUG = NO BUG paradigm in programming.
* [ ] EpanechnikovLikeKernel pdf differs from scipy implementation - we provide the valid kernel, the user decides to use it or not;
* [ ] `mean_firing_rate` can easily support a list of spiketrains. Moreover, we can leverage the advantage of our parallel module here transparently for the user. I'm tempted to quickly do it in this PR, but it should be a separate topic.
* [ ] `_check_consistency_of_spiketrains()` is very strict and the requirements can be relaxed
* [x] why `tau=sqrt(3)*sigma` and not just `sigma` in RectangularKernel? - to enclose approx. the same PDF mass as in the rest of the kernels;
* [x] `instantaneous_rate()` todo (addressed in the last commits):
```python
    if not trim:
        # TODO: -kernel_array_size + mi + 1 ? to be consistent with 'valid'
        #  mode
        rate = rate[median_id: -kernel_array_size + median_id]
```

Known issues:
* EpanechnikovLikeKernel class doc reference is incorrectly rendered. This is the known numpydoc [issue](https://github.com/numpy/numpydoc/issues/214) which has been addressed recently in https://github.com/numpy/numpydoc/pull/259. It should be fixed automatically with the next release of numpydoc.

--------
Update

* added `center_kernel` option in `instantaneous_rate()` function (fixes #107)
* added CDF, ICDF in kernels to correctly estimate the median index


elephant/statistics.py
```diff
@@ -605,9 +612,15 @@ def instantaneous_rate(spiketrain, sampling_period, kernel='auto',
                       sampling_period.rescale(units).magnitude,
                       sampling_period.rescale(units).magnitude) * units
 
+    if center_kernel:
+        fft_mode = 'full'
```
dizcza, line 616: trimming is performed after the convolution

PR #316, opened 2020-04-29
Naming convention (binsize -> bin_size, etc.)

dizcza:
**IMPORTANT**. The old arguments are deprecated but not _removed_. Even though it looks like they are removed from the code, the old signature will still work - it just throws a warning, which can become a hard error in 2 or 3 releases since then, if we want.

Below is a compiled list, made by elephant doc team with :heart: 

### Elephant-wise
* `binsize --> bin_size`
* `num_bins --> n_bins` in all modules except GPFA `transform_info` dict that has the key `'num_bins'`; commit 0408896b2e98903b98ca7fd0f63e8c7bb1a33bf2
[](https://api.gh-polls.com/poll/01ECSY4HH88SCB441MTT2FSRH3/I%20want%20num_bins%20back%21/vote)
* `binned_st[s] --> binned_spiketrain` because `binned_spiketrain`**s**, as `spiketrains`, would imply a list of BinnedSpikeTrain objects
* `fs -> sampling_frequency`. `fs` is used in 4 elephant functions: butter, wavelet_transform, welch_psd, and welch_cohere.
[](https://api.gh-polls.com/poll/01ECSXV9KSVQHBV9M6YPTGA87B/fs/vote)
[](https://api.gh-polls.com/poll/01ECSXV9KSVQHBV9M6YPTGA87B/sampling_frequency/vote)
* `_freq -> _frequency`? `freq` is used 4 times: butter, welch_psd, welch_coherence, and GPFA
[](https://api.gh-polls.com/poll/01ECT19QM9476W40GYDM91REMG/freq/vote)
[](https://api.gh-polls.com/poll/01ECT19QM9476W40GYDM91REMG/frequency/vote)
* a pair of similar signals, spiketrains, and other variables will have suffixes `_i` and `_j`. Below is a list of affected functions.
```
elephant/asset.py:    def __init__(self, spiketrains, spiketrains_y=None, bin_size=3 * pq.ms,
elephant/asset.py:                 t_start_x=None, t_start_y=None, t_stop_x=None, t_stop_y=None
elephant/spike_train_correlation.py:    _CrossCorrHist.__init__(self, binned_spiketrain1, binned_spiketrain2, window):
elephant/spike_train_correlation.py:    def get_valid_lags(binned_spiketrain1, binned_spiketrain2):
elephant/spike_train_correlation.py:    def spike_time_tiling_coefficient(spiketrain_1, spiketrain_2, dt=0.005 * pq.s):
elephant/spike_train_correlation.py:    def run_P(spiketrain_1, spiketrain_2):
elephant/spike_train_correlation.py:    def cross_correlation_histogram(binned_spiketrain1, binned_spiketrain2,
elephant/spectral.py:def welch_cohere(x, y, n_segments=8, len_segment=None, freq_resolution=None,
```
[](https://api.gh-polls.com/poll/01ECSZVA9GG27WZ5JM5TFPKQSK/_x%20and%20_y/vote)
[](https://api.gh-polls.com/poll/01ECSZVA9GG27WZ5JM5TFPKQSK/_1%20and%20_2/vote)
[](https://api.gh-polls.com/poll/01ED8T9JT9WQQW7A6X9479V9TG/_i%20and%20_j/vote)

### Function-specific
* cubic
  - `data -> histogram`
  - `ximax -> max_iterations`
* victor_purpura_dist
  - `trains --> spiketrains`
  - `q -> cost_factor`; is there a particular reason, a paper, where `q` is used?
* cv2 and lv `v --> time_intervals`
* cell_assembly_detection 
  - `data --> binned_spiketrain`
  - `maxlag --> max_lag`
  - `min_occ -> min_occurences`
  - `same_config_cut -> same_configuration_pruning`
* cross_correlation_function 
  - `ch_pairs --> channel_pairs`
  - `nlags --> n_lags`
  - `env -> hilbert_envelope`
* `lst_input -> input_spiketrains` in BinnedSpikeTrain; closes #311
* hilbert `N --> padding` (`N` is passed in `scipy.signal.hilbert`, which has this argument)
* wavelet_transform
  - `nco -> n_cycles`
* welch_psd, welch_cohere
  - `num_seg -> n_segments`
  - `len_seg -> len_segment`
  - `freq_res -> freq_resolution`
* `welch_cohere() -> welch_coherence()`. As a remark, scipy, to my surprise, uses `scipy.signal.coherence` and not `scipy.signal.cohere`.
* spike_extraction `extr_interval -> interval` - the function name already has the key "extraction"
* peak_detection `format -> as_array`
* single_interaction_process
  - `rate_c -> coincidence_rate`
  - `return_coinc -> return_coincidences`
  - `n -> n_spiketrains`
* JointISI `refr_period -> refractory_period`
* surrogates `surr_method -> method`
* `sskernel --> optimal_kernel_bandwidth`
  - `w -> bandwidth`
  - `tin -> times`
* spike_train_surrogates module
  - `n --> n_surrogates`
* estimate_csd `coords -> coordinates`
* kernels `t -> time`
* `extract_neo_attrs  -> extract_neo_attributes`
  - `obj -> neo_object`
* generate_lfp
  - `ele_xx -> x_positions` or `positions_x`
  - `xlims -> x_limits` or `limits_x`? Currently, it's `x_limits`, but then why do we use `max_something` and not `something_max`?
  - `res -> resolution`;
* multiple_filter_test
  - `dt -> time_step or step`
* empirical_parameters
  - `dt -> time_step or step`
* victor_purpura_dist -> victor_purpura_distance
  - `trains -> spiketrains`
  - `q -> cost_factor`
* van_rossum_dist -> van_rossum_distance
  - `trains -> spiketrains`
  - `tau -> time_constant`
* spike_train_timescale
  `tau_max -> max_tau`
* compound_poisson_process `A -> amplitude_distribution`
* cross_correlation_histogram `cross_corr_coef -> cross_correlation_coefficient`
[](https://api.gh-polls.com/poll/01ECT0FFY8DSX08MRRAR6ZT2J7/cross_corr_coef/vote)
[](https://api.gh-polls.com/poll/01ED8TGT8PYBQNRJK4HV5NPFDQ/cross_corr_coefficient/vote)
[](https://api.gh-polls.com/poll/01ECT0FFY8DSX08MRRAR6ZT2J7/cross_correlation_coefficient/vote)
* `corrcoef() -> correlation_coefficient()`
[](https://api.gh-polls.com/poll/01ECT0DC6NXGDS07PSZNBD1NT2/corrcoef/vote)
[](https://api.gh-polls.com/poll/01ECT0DC6NXGDS07PSZNBD1NT2/corr_coef/vote)
[](https://api.gh-polls.com/poll/01ECT0DC6NXGDS07PSZNBD1NT2/correlation_coefficient/vote)

## Not refactored

Because they require special care (ASSET already follows the naming convention):
* SPADE
* Unitary Event Analysis
* GPFA

Also, I didn't pay attention to _protected functions not because they can avoid the naming convention but because I didn't want to spend time refactoring them. Keep in mind that any decent renaming may lead to breaking the logic and thus should be justified.

## Won't change

* In `homogeneous_gamma_process()` function `a` is the shape and `b` is the **rate** of a distribution will not be renamed to avoid a misinterpretation of the **rate** of the outcome spiketrain, which is equal to `b/a` and not `b`.
* In `surrogates()` function `dt` won't be changed because it means different things depending on the surrogate method choice.
* In `multiple_filter_test()` and `empirical_parameters()` functions `t_final` won't be changed to `t_stop` because it serves a different purpose.
* `nfft` became a standard over the years, even though it does not comply with our convention, which renders this variable as `n_fft`.
[](https://api.gh-polls.com/poll/01ECSY833X15GQX7GH4F2RPM0W/nfft/vote)
[](https://api.gh-polls.com/poll/01ECSY833X15GQX7GH4F2RPM0W/n_fft/vote)


elephant/spike_train_generation.py
```diff
@@ -200,9 +204,16 @@ def peak_detection(signal, threshold=0.0 * pq.mV, sign='above', format=None):
         Determines whether to count threshold crossings that cross above or
         below the threshold.
         Default: 'above'.
+    as_array : bool, optional
+        If True, a NumPy array of the resulted peak times is returned instead
```
JuliaSprenger, line 208: This should be `resulting`

elephant/spike_train_generation.py
```diff
@@ -736,16 +754,16 @@ def single_interaction_process(
...
-          entries must be larger than `rate_c`.
-    rate_c : pq.Quantity
+          entries must be larger than `rate_coincidence`.
+    rate_coincidence : pq.Quantity
         Coincidence rate (rate of coincidences for the n-dimensional SIP).
-        The SIP spike trains will have coincident events with rate `rate_c`
-        plus independent 'background' events with rate `rate-rate_c`.
+        The SIP spike trains will have coincident events with rate `rate_coincidence`
+        plus independent 'background' events with rate `rate-rate_coincidence`.
     n : int, optional
```
JuliaSprenger, line 767: Is `n` telling enough here?

elephant/statistics.py
```diff
@@ -947,14 +956,14 @@ def sskernel(spiketimes, tin=None, w=None, bootstrap=False):
     ----------
     spiketimes : np.ndarray
         Sequence of spike times (sorted to be ascending).
-    tin : np.ndarray, optional
+    time_points : np.ndarray, optional
```
JuliaSprenger, line 959: Maybe `times` would be even sufficient here?

elephant/current_source_density.py
```diff
@@ -201,43 +201,52 @@ def estimate_csd(lfp, coords=None, method=None,
...
+        Defaults to [0.,1.], use only in 3D case
+    resolution : int, optional
+        The resolution of the integration
+        Defaults to 50
+
+    Returns
+    -------
+    LFP : neo.AnalogSignal
+       The potentials created by the csd profile at the electrode positions.
+       The electrode positions are attached as RecordingChannel's coordinate.
```
JuliaSprenger, line 249: This sentence is outdated. `RecordingChannels` are removed in current Neo versions in favour of `ChannelIndex` objects.

elephant/spike_train_generation.py
dizcza: right.
moved to our questions

kohlerca:
Regarding `cross_coef`/`cross_corr_coef` in `spike_train_correlation`:

The parameter in `cross_correlation_histogram` function is passed to the protected class `_CrossCorrHist`. Therefore, according to the conventions of just using `corr` as an abbreviation for correlation, `cross_corr_coefficient` will make the information clearer. The fact that it is passed internally, I don't see as an annoyance of a larger parameter name. 

Moreover, in `_CrossCorrHist`, the class method should be `cross_correlation_coefficient` to keep the standard as in `spike_time_tiling_coefficient`, and to align to what we discussed regarding keeping the Elephant function names without abbreviations. The latter also applies to function `corrcoef`. Although NumPy has this function, in Elephant we are talking about an implementation that works with `elephant.BinnedSpikeTrain` inputs. I don't see any benefit in shadowing NumPy.

Finally, in the class `_CrossCorrHist`, the current `cross_corr_coef` function takes an argument named only `cross_corr`. It should also define which is the input, as `cross_corr*` is a prefix for histogram or matrices. Therefore, something in the direction of `cross_corr_array` or `cross_corr_matrix` will make the code clear.

kohlerca:
Regarding `_x` and `_y` or `_1` and `_2` in `welch_cohere` and `ASSET`:

I don't think they can be analyzed together. The idea of `_1` and `_2` is for the functions where the parameter order does not matter, which is the case of `welch_cohere`. 

In ASSET, the `spiketrains_y` will actually reflect in the interpretation of the matrices axes. Therefore, `_x` and `_y` is meaningful with respect to the method output. ASSET is tough because you don't need both parameters, but I think it is better to name the first (non-optional) as `_x` and the optional as `_y`.

elephant/spike_train_dissimilarity.py
```diff
@@ -276,11 +280,12 @@ def _victor_purpura_dist_for_st_pair_intuitive(
...
                 scr[i, j] = min(scr[i-1, j]+1, scr[i, j-1]+1)
                 scr[i, j] = min(scr[i, j], scr[i-1, j-1] + np.float64((
-                               q*abs(train_a[i-1]-train_b[j-1])).simplified))
+                                                                              cost_factor * abs(spiketrain_a[i - 1] - spiketrain_b[j - 1])).simplified))
     return scr[nspk_a, nspk_b]
 
 
-def van_rossum_dist(trains, tau=1.0 * pq.s, sort=True):
+@deprecated_alias(trains='spiketrains')
+def van_rossum_dist(spiketrains, tau=1.0 * pq.s, sort=True):
```
kohlerca, line 288: Can't `tau` be replaced by `time_constant`? In Van Rossum (2001), the formula does not have `tau`. It mentions `t_c` in the formula for the parameter of the exponential function. Therefore, not only we should rename this to the meaningful parameter, but also make the proper formula reference in the Notes.

Also, the function could be renamed to `van_rossum_distance`.

elephant/spike_train_dissimilarity.py
```diff
@@ -35,8 +37,9 @@ def _create_matrix_from_indexed_function(
     return mat
 
 
-def victor_purpura_dist(
-        trains, q=1.0 * pq.Hz, kernel=None, sort=True, algorithm='fast'):
+@deprecated_alias(trains='spiketrains', q='cost_factor')
+def victor_purpura_dist(spiketrains, cost_factor=1.0 * pq.Hz, kernel=None,
```
kohlerca, line 41: If also renaming functions, `_dist` should be `_distance`

elephant/spike_train_correlation.py
```diff
@@ -906,7 +925,8 @@ def run_T(spiketrain):
 sttc = spike_time_tiling_coefficient
 
 
-def spike_train_timescale(binned_st, tau_max):
+@deprecated_alias(binned_st='binned_spiketrain')
+def spike_train_timescale(binned_spiketrain, tau_max):
```
kohlerca, line 929: `tau_max` should be `max_tau` per the convention for min/max values. Regarding the Greek letter, this function seems to be specifically tied to the mathematical definition showed in the docstring, so I would leave it as `tau` unless another meaningful name can come.

mdenker: True, but of course, here we again have the stupid case where `\tau_{max}` is a parameter of the function. Since I think the decision was to be consistent in not calling variable names as latex-like mathematical variables, I guess that's good to change...

elephant/spike_train_dissimilarity.py
mdenker: Indeed, this `dist` could be misread as distribution...

elephant/spike_train_dissimilarity.py
mdenker: I would agree to rename to `time_constant` in this case, making the reference to t_c is also a good idea.

mdenker:
@kohlerca Regarding _x and _y or _1 and _2 in welch_cohere and ASSET.

Also in welch_cohere, the order of parameters makes a big differene when interpreting the phase_lag return value, since the order defines the reference and the referenced, so to say. In this way, I think the situation is very similar to ASSET. Now, I think the main reason we dislike 1 and 2 is because it looks terrible in the variable name, and I think we dislike x and y because it is commonly associated to a position. How about a compromise and using `i` and `j`? This is commonly used to identify this type of relationship...

kohlerca:
@mdenker How about a compromise and using `i` and `j`? This is commonly used to identify this type of relationship...

I guess this makes total sense and it is better indeed than either of the previous options.

elephant/spike_train_dissimilarity.py
dizcza: Renamed to van_rossum_distance, replaced arg name by time_constant, and added a reference to the paper.

elephant/spike_train_correlation.py
dizcza: put in the "unsure" section

dizcza:
> @mdenker How about a compromise and using `i` and `j`? This is commonly used to identify this type of relationship...
> 
> I guess this makes total sense and it is better indeed than either of the previous options.

Added `_i and _j` option in the poll.

PR #320, opened 2020-05-01
Fixed bug in sorting of dither spikes

pbouss:


elephant/test/test_spike_train_surrogates.py
```diff
@@ -41,6 +42,7 @@ def test_dither_spikes_output_format(self):
             self.assertEqual(surrogate_train.t_start, spiketrain.t_start)
             self.assertEqual(surrogate_train.t_stop, spiketrain.t_stop)
             self.assertEqual(len(surrogate_train), len(spiketrain))
+            assert_array_less(0., np.diff(surrogate_train))  # check ordering
```
dizcza, line 45: is it the regression test that shows that the bug has been fixed where previously the `np.diff(surrogate_train)` was not negative?

dizcza: If yes, it's not clear where the axis comes into play.
Have you thought of a proper test where axis makes a difference?

pbouss: Yes, exactly before there were elements of np.diff(surrogate_train) that were negative. Now it shows that they are all positive and thus properly sorted.

pbouss: If you would run, it for the old axis version this test fails.

PR #328, opened 2020-05-27
Timescale function option to return nan if there are few spikes (<=2)

morales-gregorio:
Hey!

I have added an option to the `spike_train_timescale` function to return nan if there are 2 or less spikes in the given spiketrains.

To this I have also added two tests:
1. Testing the error raising (which probably should have been done in the original PR)
2. Testing that indeed the nans are returned for spiketrains with 0, 1 or 2 spikes and that whatever is returned for spiketrains with 3 or 4 spikes is not a nan.

Please have a look and let me know if anything should be changed. 

Best,
Aitor

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -906,7 +906,7 @@ def spike_time_tiling_coefficient(spiketrain_1, spiketrain_2, dt=0.005 * pq.s):
 sttc = spike_time_tiling_coefficient
 
 
-def spike_train_timescale(binned_st, tau_max, with_nan=False):
+def spike_train_timescale(binned_st, tau_max):
     r"""
     Calculates the auto-correlation time of a binned spike train.
     Uses the definition of the auto-correlation time proposed in [[1]_,
@@ -924,15 +924,14 @@ def spike_train_timescale(binned_st, tau_max, with_nan=False):
     binned_st : elephant.conversion.BinnedSpikeTrain
         A binned spike train containing the spike train to be evaluated.
     tau_max : pq.Quantity
-        Maximal integration time of the auto-correlation function.
-    with_nan : bool
-        If True returns NaN for spike trains with 2 or less spikes.
-        Default: False
+        Maximal integration time of the auto-correlation function. It needs to
+        be a multiple of the binsize of `binned_st`.
 
     Returns
     -------
     timescale : pq.Quantity
-        The auto-correlation time of the binned spiketrain.
+        The auto-correlation time of the binned spiketrain. If `binned_st` has
+        less than 2 spikes `np.nan` is returned.
 
     Notes
     -----
@@ -952,6 +951,9 @@ def spike_train_timescale(binned_st, tau_max, with_nan=False):
         Slow fluctuations in recurrent networks of spiking neurons.
         Physical Review E, 92(4), 040901.
     """
+    if binned_st.get_num_of_spikes() < 2:
+        return np.nan
+
     binsize = binned_st.binsize
     if not (tau_max / binsize).simplified.units == pq.dimensionless:
         raise ValueError("tau_max needs units of time")
@@ -961,9 +963,6 @@ def spike_train_timescale(binned_st, tau_max, with_nan=False):
     if not np.isclose(tau_max.simplified.magnitude,
                       (tau_max_bins * binsize).simplified.magnitude):
         raise ValueError("tau_max has to be a multiple of the binsize")
-
-    if len(np.where(binned_st.to_array())[1]) <= 2 and with_nan:
-        return np.nan
 
     cch_window = [-tau_max_bins, tau_max_bins]
     corrfct, bin_ids = cross_correlation_histogram(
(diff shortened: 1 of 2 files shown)


elephant/spike_train_correlation.py
```diff
@@ -925,6 +925,9 @@ def spike_train_timescale(binned_st, tau_max):
         A binned spike train containing the spike train to be evaluated.
     tau_max : pq.Quantity
         Maximal integration time of the auto-correlation function.
+    with_nan : bool
```
dizcza, line 928: bool, optional

elephant/spike_train_correlation.py
```diff
@@ -951,13 +954,16 @@ def spike_train_timescale(binned_st, tau_max):
...
+        raise ValueError("tau_max needs units of time")
 
     # safe casting of tau_max/binsize to integer
     tau_max_bins = int(np.round((tau_max / binsize).simplified.magnitude))
     if not np.isclose(tau_max.simplified.magnitude,
                       (tau_max_bins * binsize).simplified.magnitude):
-        raise AssertionError("tau_max has to be a multiple of the binsize")
+        raise ValueError("tau_max has to be a multiple of the binsize")
+
+    if len(np.where(binned_st.to_array())[1]) <= 2 and with_nan:
```
dizcza, line 965: a more efficient and clear version is
`if with_nan and binned_st.get_num_of_spikes() <= 2`

However, I'm not sure that we should return NaN _only_ when `with_nan` is specified. My vision is that if the num. of spikes is less than a particular threshold (zero, one or two - you pick the threshold!), then we must return NaN, followed by a warning that a NaN is returned because there are not enough spikes. If this is the case, we don't need `with_nan` argument at all!

That is my concern. We'll discuss this on Tue morning. In the meanwhile, what do you think?

dizcza: Alright, we agreed on the following:
* remove with_nan
* if n_spikes < 2:
   warning.warn(blah-blah)
   return np.nan

because if the num. of spikes is less than 2, it does not make sense to compute CCH anyway.

morales-gregorio: Hey @dizcza! Thanks for reviewing!

I will implement the changes right now. 

I only have one major concern, I would not raise a warning for this, this to me is expected behavior and I would not want my terminal flooded with warnings when I am applying this function to hundreds of slices of data.

Or does warnings have some fancy option to only warn for the same thing once in the same session?

dizcza: yes, use `warnings.simplefilter('once', UserWarning)` either inside the function or in the module (py file) itself - need to test. If inside the function works, leave it there - just before calling warnings.warn.

morales-gregorio: I am quite sure it would not work from inside the function (due to multiple calls from different executions), but I will add it to my workflow and that will hopefully work!

morales-gregorio:
Just added the warning and the other comments. If no new issues arise (hopefully the tests will pass) 

Let me know if any other thing needs fixing, otherwise I am ready for merging on my side.

Thanks! Aitor

dizcza:
Hm, somehow this PR was out of my scope for a week. I don't see any other problems of why it should not be merged.

Your comment about warning only once per session is not addressed though. I'm not sure if it makes sense to warn only once because then all other functions in this module will warn only once, which might be actually the desired behavior since the users don't like to be flooded with warnings.

We'll discuss this particular aspect - whether or not add `warnings.simplefilter('once', UserWarning)` at the top of the module - next Monday.

morales-gregorio:
Hey any updates on this? I would just let the user deal with filtering the warnings.

dizcza:
I need to look if adding `warnings.simplefilter('once', UserWarning)` in one module does not escalate the behavior to runtime.
If it's not the case, I'll ask you to add it there.
If it's the case, no additional changes are needed.

dizcza:
@morales-gregorio 

There are two issues: one is on your side and the other is scipy [issue](https://github.com/scipy/scipy/issues/12449).

Add the following check that the warning is triggered in `test_timescale_nan`:
```python
with self.assertWarns(UserWarning):
    timescale = sc.spike_train_timescale(bst, tau_max)
```
followed by fixing the bug in the function itself - there is an extra `,` in the warning message (see the suffix). Just remove the comma.

Note that `assertWarns` requires py2.7 (yeah, we're still supporting it) and thus you need to wrap the test function like so
```python
@unittest.skipUnless(python_version_major == 3, "assertWarns requires 3.2")
def test_timescale_nan(self)
```

I could fix these minor things for you if the pull request originated from INM-6/elephant.

If there are still questions about how to fix them, let me know.
Otherwise, once you address your part (don't bother with many warnings, it's on scipy side), I merge the PR.

morales-gregorio:
Thanks @dizcza for having a closer look!
 I hope everything is fine after the last commits

dizcza:
I meant
```python
        for st in [st0, st1]:
            bst = conv.BinnedSpikeTrain(st, binsize)
            with self.assertWarns(UserWarning):
                timescale = sc.spike_train_timescale(bst, tau_max)
            self.assertTrue(math.isnan(timescale))
```

sorry

:)

morales-gregorio:
Hahaha I should have realised that what I wrote made no sense ;)

PR #332, opened 2020-06-24
Time-domain pairwise Granger causality 

Qkley:
This is the implementation of time-domain Granger causality for two signals. The documentation needs more work and we are also discussing one issue regarding an edge case. However, all the discussion can be moved to here from this point forward since the implementation is at its final stages.

Alongside the creation of the new module, we have added ourselves to the doc/authors.rst and doc/modules.rst.


Qkley:
We are discussing a degenerate case where two **identical** signals are passed to the function. In this case, no additional information is provided by the second time-series. Currently, we are getting small negative values. At least one other implementation of spectral causality addresses such issues by explicitly setting directional causalities of identical signals as NaN.

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+    optimal_coeffs = np.zeros((dimension, dimension, optimal_order))
+    optimal_cov_matrix = np.zeros((dimension, dimension))
+
+    if information_criterion == 'bic':
+        evaluate_ic = bic
+    elif information_criterion == 'aic':
+        evaluate_ic = aic
+    else:
+        raise ValueError(
+            f"Information criterion {information_criterion} not valid")
```
dizcza, line 278: it's easily solved by passing `information_criterion=bic` meaning that information_criterion is a callable and not a string. Then you don't need this checking block at all.

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+            optimal_coeffs = coeffs
+            optimal_cov_matrix = cov_matrix
+
+    return optimal_coeffs, optimal_cov_matrix, optimal_order
+
+
+def pairwise_granger(signals, max_order, information_criterion='bic'):
+    r"""
+    Determine Granger Causality of two time series
+    Note: order parameter should be removed
```
dizcza, line 297: don't see order parameter anymore

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+    Parameters
+    ----------
+    signals : np.ndarray or neo.AnalogSignal
+        time series data
+    max_order : int
+        maximal order of autoregressive model
+    information_criterion : string
+        bic for Bayesian information_criterion,
+        aic for Akaike information criterion
+    Returns
```
dizcza, line 307: any two doc sections should be separated with a new line
apply to all public functions

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+    return optimal_coeffs, optimal_cov_matrix, optimal_order
+
+
+def pairwise_granger(signals, max_order, information_criterion='bic'):
+    r"""
+    Determine Granger Causality of two time series
+    Note: order parameter should be removed
+    Parameters
+    ----------
+    signals : np.ndarray or neo.AnalogSignal
```
dizcza, line 300: Is signals variable always a (2, N) array?

Qkley:
Here is the [actual data](https://github.com/NeuralEnsemble/elephant/files/4825771/granger_timeseries_groundtruth_data.zip) for ground truth test generated on my machine.

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+from neo.core import AnalogSignal
+
+Causality = namedtuple('causality',
+                       ['directional_causality_x_y',
+                        'directional_causality_y_x',
+                        'instantaneous_causality',
+                        'total_interdependence'])
+
+
+def bic(cov, order, dimension, length):
```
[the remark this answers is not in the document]
Qkley: I would suggest that bic and aic initialisms are quite conventional. Other packages, [vars R package](https://cran.r-project.org/web/packages/vars/vars.pdf) for example, use them extensively.

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+    criterion : float
+       Bayesian Information Criterion
+    """
+    criterion = 2 * np.log(np.linalg.det(cov)) \
+        + 2*(dimension**2)*order*np.log(length)/length
+
+    return criterion
+
+
+def aic(cov, order, dimension, length):
```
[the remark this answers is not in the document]
Qkley: Please see above reply to the bic comment.

elephant/causality/granger.py
Qkley: In the context of `pairwise_granger` yes.

dizcza: wouldn't it make more sense to replace `signals` by `signals_x` and signals_y` in the argument list?

Qkley:
Merged master into this branch (commit: https://github.com/NeuralEnsemble/elephant/pull/332/commits/3e138bc976b8ba829495bdab349e320200bf996b)

elephant/causality/granger.py
dizcza: besides, we're still with py2 and py3.5 that don't know about nice in-place string formatting of a form `f"{}"`

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+        dimensionality of the data
+    length : int
+        number of time samples
+
+    Returns
+    -------
+    criterion : float
+       Bayesian Information Criterion
+    """
+    criterion = 2 * np.log(np.linalg.det(cov)) \
```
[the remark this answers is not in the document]
Qkley: Addressed with commit 4eb0aaa

elephant/causality/granger.py
Qkley: Addressed in commit 934b047

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+        x: d-dimensional signal
+        x^T: transpose of d-dimensional signal
+        N: number of time points
+        \tau: lag
+
+        C(\tau) = \sum_{i=0}^{N-\tau} x[i]*x^T[\tau+i]
+
+    """
+    length = np.size(signals[0])
+    assert (length >= max_lag), 'maximum lag larger than size of data'
```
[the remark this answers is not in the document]
Qkley: Addressed in commit dda3c86

elephant/causality/granger.py
```diff
@@ -0,0 +1,382 @@
...
+
+:copyright: Copyright 2014-2020 by the Elephant team, see `doc/authors.rst`.
+:license: Modified BSD, see LICENSE.txt for details.
+"""
+
+import numpy as np
+from collections import namedtuple
+from neo.core import AnalogSignal
+
+Causality = namedtuple('causality',
```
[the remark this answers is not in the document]
Qkley: Addressed in commit a7db281

elephant/causality/granger.py
Qkley: Addressed in commit 77e87f3

elephant/causality/granger.py
Qkley: Addressed in commit daa2bc6

elephant/causality/granger.py
Qkley: We believe that for the time being we would like to keep it the way it is. Once we add more functions, such as conditional Granger causality, we can discuss this point again.

Qkley:
> We are discussing a degenerate case where two **identical** signals are passed to the function. In this case, no additional information is provided by the second time-series. Currently, we are getting small negative values. At least one other implementation of spectral causality addresses such issues by explicitly setting directional causalities of identical signals as NaN.

We have addressed this in commit 4eb0aaa. This was partially solved by switching from `np.linalg.det()` to `np.linalg.slogdet()`. We check the sign of the `np.linalg.slogdet()` and if it is not positive, it raises a ValueError exception. A unit test for such a case can be added as well, or I could replace the:
```python

    def test_pairwise_granger_error(self):
        null_signals = np.array([[0, 0], [0, 0]])
        with self.assertRaises(ValueError):
            elephant.causality.granger.pairwise_granger(
                null_signals, max_order=2)
```
with something like:
```python
    def test_pairwise_granger_error(self):
        same_signal = np.vstack([self.signal[0], self.signal[0]])
        with self.assertRaises(ValueError):
            elephant.causality.granger.pairwise_granger(
                same_signal, max_order=2)
```
What is your opinion?

elephant/causality/granger.py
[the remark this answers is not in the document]
dizcza: Unless you know that slogdet returns two elements, the sign and the abs value, it's hard to read why you take `[1]` from the output. A better approach would be `sign, logdet = slogdet()` where you explicitly specify the return values.

elephant/causality/granger.py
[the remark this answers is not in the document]
dizcza: Additionally, you could inform the user of what the values are.
"Maximum lag ({lag}) larger than the size ({size}) of data".format(lag=lag, size=size).

elephant/causality/granger.py
```diff
@@ -0,0 +1,395 @@
...
+
+    Parameters
+    ----------
+    signals : np.ndarray
+        time series data
+    dimension : int
+        dimensionality of the data
+    max_order : int
+        maximal order to consider
+    information_criterion : string
```
dizcza, line 265: it's not a string anymore, it's a `callable`. Then, because `bic` and `aic` are special keywords that need to be rendered as the functions on readthedocs, they need to be enclosed in ticks '\`', and the doc should look like
```
information_criterion : callable
        `bic` for Bayesian information_criterion,
        `aic` for Akaike information criterion
```

Or you even better

":func:\`aic\`"

on your choice.

elephant/causality/granger.py
dizcza: Got it.

doc/reference/causality-overview.rst
```diff
@@ -0,0 +1,24 @@
...
+"""""""""""
+The user must be mindful of the method's limitations, which are assumptions of covariance stationary data, linearity imposed by the underlying autoregressive modelling as well as the fact that the variables not included in the model will not be accounted for [2].
+
+References
+""""""""""
+[1] Granger CWJ (1969) Investigating causal relations by econometric models and cross-spectral methods. Econometrica 37:424-438.
+
+[2] Seth A (2007) Granger causality. Scholarpedia, 2(7):1667., revision #127333
+
+[3] Ding M, Chen Y, Bressler SL (2006) Granger causality: Basic theory and application to neuroscience. In Schelter S, Winterhalder N, Timmer J. Handbook of Time Series Analysis. Wiley, Wienheim.
```
dizcza, line 20: I don't see any reference to [3].

doc/reference/causality.rst
```diff
@@ -0,0 +1,7 @@
+=========================================
+`causality` - Granger causality measures
```
dizcza, line 2: No ticks in the title. Just call it "Granger causality measures". It could be renamed to just "Causality measures" once you add other measures.

elephant/causality/granger.py
```diff
@@ -0,0 +1,395 @@
...
+---------------------
+Various formulations of Granger causality have been developed. In this module
+you will find function for time-series data to test pairwise Granger causality
+(`pairwise_granger`).
+
+Time-series Granger causality
+~~~~~~~~~~~~~~~
+
+.. autosummary::
+    :toctree: causality/
```
dizcza, line 17: :toctree: toctree/causality/

doc/reference/causality-overview.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Limitations
+"""""""""""
+The user must be mindful of the method's limitations, which are assumptions of covariance stationary data, linearity imposed by the underlying autoregressive modelling as well as the fact that the variables not included in the model will not be accounted for [2].
+
+References
+""""""""""
+[1] Granger CWJ (1969) Investigating causal relations by econometric models and cross-spectral methods. Econometrica 37:424-438.
+
+[2] Seth A (2007) Granger causality. Scholarpedia, 2(7):1667., revision #127333
```
dizcza, line 18: We start using bibtex for references. So you need to move the references to `doc/bib/elephant.bib` and then link them here. For example, see how it's done in unitary events analyses and asset. It should be something like

Add in `doc/bib/elephant.bib`:

`@article{yourcustomcausalityreferencename, ...`

Add in `doc/reference/causality`:

```

References
----------

.. bibliography:: ../bib/elephant.bib
   :labelprefix: gr (or ca)
   :keyprefix: granger- (or causality-)
   :style: unsrt
```

In `.rst` or `.py` file, where the documentation is described (in your case, it's `causality-overview`), use it as 
":cite:\`granger-yourcustomcausalityreferencename\`" in-place.

elephant/causality/granger.py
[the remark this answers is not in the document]
dizcza: Yes, I found that some packages use `bic` and some `bayesian_information_criterion`. It seems that the majority is on `bic` side, but we will come to this issue once we get mature with the current naming convention PR.

dizcza:
>> We check the sign of the np.linalg.slogdet() and if it is not positive, it raises a ValueError exception. A unit test for such a case can be added as well

>  Sounds good

On second thought, maybe it makes sense to allow for small negative values like `-1e-7` due to the floating-point issues and raise the error if `logdet < -1e-7`? What is the correct behavior for degenerate cases when `logdet` is (infinitely close to) zero?

elephant/causality/granger.py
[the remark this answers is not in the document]
Qkley: Thank you. In that case, I am going to set it as 'Resolved' for now.

elephant/causality/granger.py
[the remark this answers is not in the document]
Qkley: This was fixed for BIC and AIC as per your comment in commit a31305d.

elephant/causality/granger.py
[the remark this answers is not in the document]
Qkley: Fixed in commit a31305d.

elephant/causality/granger.py
Qkley: Addressed in commit 5c26f0d. We chose the first option for the time being.

doc/reference/causality-overview.rst
Qkley: Addressed in commit 6e2739d.

doc/reference/causality.rst
Qkley: Could you please elaborate. Do you mean something like this:
```
======================================
causality - Granger causality measures
```
Or do you want to get rid of "causality" altogether?

elephant/causality/granger.py
Qkley: Addressed in the commit 52eda3a.

Qkley:
> > > We check the sign of the np.linalg.slogdet() and if it is not positive, it raises a ValueError exception. A unit test for such a case can be added as well
> 
> > Sounds good
> 
> On second thought, maybe it makes sense to allow for small negative values like `-1e-7` due to the floating-point issues and raise the error if `logdet < -1e-7`? What is the correct behavior for degenerate cases when `logdet` is (infinitely close to) zero?

Anno and I briefly discussed the above points:

The determinant of a covariance matrix cannot be negative. The covariance matrix is positive semi-definite. Therefore, its determinant is non-negative. 

When it comes to the case of logdet being infinitely close to zero, then the estimation becomes problematic. It becomes impossible to calculate causality in such a case, because underlying statistical model breaks down.

dizcza:
> The determinant of a covariance matrix cannot be negative. The covariance matrix is positive semi-definite. Therefore, its determinant is non-negative.

This is clear. I was talking about the machine precision that is `logdet` should be zero but it returns you a small negative value. So if you
* allow zero `logdet` -> you should account for a small negative `logdet` and thus you cannot rely on the sign only; in this case, raise an error if `logdet < -tolerance` where `tolerance=1e-7`;
* don't allow zero -> you can raise an error if the sing is negative and you don't care about the absolute value of `logdet`.

Taking into account your second comment,
> When it comes to the case of logdet being infinitely close to zero, then the estimation becomes problematic. It becomes impossible to calculate causality in such a case, because underlying statistical model breaks down.

the question is should you
* warn a user about the degenerate covariance matrix (`logdet` is close to zero) but still proceed with the computation;
* raise an error, explaining that the following computation does not make sense
?

doc/reference/causality.rst
dizcza: > Or do you want to get rid of "causality" altogether?

yes. I suggest

```
======================
Granger causality measures
```

elephant/causality/granger.py
dizcza: Pay attention to the ticks. There are 3 types
```
`
'
"
```
To highlight a code snippet or a function we use the first type. But it's minor anyway. To clarify that this is not a string, I suggest adding a sentence that it's "A function to compute the information criterion". So consider

```
information_criterion : callable, optional
    A function to compute the information criterion:
        `bic` for Bayesian information_criterion,

        `aic` for Akaike information criterion
    Default: bic
```

Qkley: Fixed in commit 513790d.

doc/reference/causality.rst
Qkley: Addressed in commit 74c933a.

doc/reference/causality-overview.rst
Qkley: This has been done in commits 632604a and e15b99b.

Qkley: Please note that I have an arxiv preprint in the bibliography. This text was later published as a book chapter.

Qkley:
> > The determinant of a covariance matrix cannot be negative. The covariance matrix is positive semi-definite. Therefore, its determinant is non-negative.
> 
> This is clear. I was talking about the machine precision that is `logdet` should be zero but it returns you a small negative value. So if you
> 
>     * allow zero `logdet` -> you should account for a small negative `logdet` and thus you cannot rely on the sign only; in this case, raise an error if `logdet < -tolerance` where `tolerance=1e-7`;
> 
>     * don't allow zero -> you can raise an error if the sing is negative and you don't care about the absolute value of `logdet`.
> 
pairwise_granger [checks if sign is positive and throws an error if it is zero or negative](https://github.com/INM-6/elephant/blob/e15b99b98ca3270402c09b8e536a65bcbccbe617/elephant/causality/granger.py#L364-L368). Therefore, we follow the second option.
> 
> Taking into account your second comment,
> 
> > When it comes to the case of logdet being infinitely close to zero, then the estimation becomes problematic. It becomes impossible to calculate causality in such a case, because underlying statistical model breaks down.
> 
> the question is should you
> 
>     * warn a user about the degenerate covariance matrix (`logdet` is close to zero) but still proceed with the computation;
> 
>     * raise an error, explaining that the following computation does not make sense
>       ?
I'd say that it is better to warn the user and proceed with the computation. This would facilitate batch processing of signals better than raising an error every time there is a problem.

doc/reference/causality-overview.rst
dizcza: > Please note that I have an arxiv preprint in the bibliography. This text was later published as a book chapter.

Qkley:
I have added a UserWarning as per our discussion in commit https://github.com/NeuralEnsemble/elephant/pull/332/commits/efefa2fc1138e7aa3bf6f8b5c308a60b4cd6691c. Please check if the message is not too ambiguous.

dizcza:
Tests are running and all lines are covered with tests - perfect!

What I did:
* set up tests such that the ground truth is created only once. Overall, the tests take only a few seconds.
* replaced decimal precision in one of the tests from 15 to 8: it's sufficient to use the default `decimal=6` in `assert_array_almost_equal`, which is tailored for float32 bits; since you implicitly use float64 (consider, however, float32), I bumped up to 8, but not 15 - it's too much; you never know what is happening on travis.com.

Important remark: The implementation, the code itself, in its current state is OK. Even if it can be vectorized, it's not obvious. And it seems to be running quite fast already for the given test data.

What is missing:
- [ ] In the Returns block of `pairwise_granger` function explain each attribute of the resulting namedtuple. Currently, they are accompanied by math only. For me, it's fine, but consider making it friendly for a typical (experimental) neuroscientist. This is, however, can be explained in a tutorial rather than in the doc.
- [ ] Are you planning to provide a tutorial? It's not mandatory but strongly recommended.
- [ ] both `bic` and `aic` functions miss the math, explaining how the information measure is computed. Moreover, the explanation of how they differ is missing. In other words, under which circumstances, a user should use one function over the other. [Minor] You can also add "See also" block ([link](https://numpydoc.readthedocs.io/en/latest/format.html#overview)) in the docs to cross-reference them.
- [ ] `bic` computation is done in one line. Python one-liners are neat, but not overuse them. If you split the logic into 2 or 3 lines and assign the result of the intermediate computation to meaningful variable names, the code becomes easier to read. I'm talking about 

```python
criterion = 2 * log_det_cov \
        + 2*(dimension**2)*order*np.log(length)/length
```

Qkley:
We addresed the first point in commit 77952f2.
When it comes to the tutorial, I think we would like to have it, but not at this point. We think it would be better to add more functions to the causality module and then prepare a tutorial explaining them.
We are not sure that the math description is necessary for AIC and BIC, because both of them are one line equations.
When it comes to the final point, BIC is a one-line equation. Therefore, it might be reasonable to keep it the way it is.

doc/reference/causality.rst
```diff
@@ -0,0 +1,5 @@
+==========================
+Granger causality measures
```
mdenker, line 2: Is this meant only for the granger functions (as the title suggests), or all causality functions (as the file name suggests)? I think it should read "Causality", which would also sort it correctly in the function overview of the docs.

.gitignore
```diff
@@ -48,6 +48,7 @@ doc/reference/toctree/asset/elephant.asset.synchronous*
 doc/reference/toctree/parallel
 doc/reference/toctree/statistics
 doc/reference/toctree/unitary_event_analysis
+doc/reference/toctree/causality
```
mdenker, line 51: Can't we just put `doc/reference/toctree` here?

doc/bib/elephant.bib
```diff
@@ -1,4 +1,23 @@
-@article{Gruen02_43, 
+@misc{ding2006granger,
```
mdenker, line 1: We should find a common bibtex key schema, I'd suggest [FirstAuthorLastName][Year2Digit]_[FirstPage]

doc/bib/elephant.bib
```diff
@@ -64,6 +83,18 @@ @article{Riehle97_1950
...
+@article{seth2007granger,
+  author = {Seth, A. },
+  title   = {Granger causality},
+  year    = {2007},
+  journal = {Scholarpedia},
+  volume  = {2},
+  number  = {7},
+  pages   = {1667},
+  doi     = {10.4249/scholarpedia.1667},
+  note    = {revision \#127333}
```
mdenker, line 95: I think revision can go out, at least the DOI should be clear to which revision it points.

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
+# -*- coding: utf-8 -*-
```
mdenker, line 1: The causality submodule works a bit different than other modules of Elephant in that it's a proper submodule. In other cases, we had a top level python file that had top-level function calls, whereas `causality` this is more a grouping submodule like `random.random` in numpy. I think it's worth discussing if we want this or if we want to stick to a flat hierarchy (which depends also on what elese we envision to be captured by `causality`. In the former case, somehow the mapping of the module to the documentation is not quite clear to me -- currently it seems to me the generic `causality` and the specific `granger` is still a bit mixed up.

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+
+
+def pairwise_granger(signals, max_order, information_criterion=aic):
+    r"""
+    Determine Granger Causality of two time series
+
+    Parameters
+    ----------
+    signals : np.ndarray or neo.AnalogSignal
+        time series data
```
mdenker, line 334: This must be more verbose. In case of an array, it's unclear what the dimensions are. In case of an AnalogSignal, it's unclear if the second dimension must be 2. Also, this should refer to X and Y.

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+    max_order : int
+        maximal order of autoregressive model
```
mdenker, line 336:
```suggestion
        The maximal order of the autoregressive model.
```

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+
+def pairwise_granger(signals, max_order, information_criterion=aic):
+    r"""
+    Determine Granger Causality of two time series
+
+    Parameters
+    ----------
+    signals : np.ndarray or neo.AnalogSignal
+        time series data
+    max_order : int
```
mdenker, line 335: Is there a common default parameter one would use here, or some value to suggest to a new user? I guess it depends a lot on the sampling of the time series...

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+            log(C|XY_00 / C|XY_11)
+        total interdependence of X,Y given by
+            log( {C|X \cdot C|Y} / det{C|XY} )
+
+    """
+    if isinstance(signals, AnalogSignal):
+        # transpose (N,2) -> (2, N)
+        signals = signals.magnitude.T
+    else:
+        signals = np.asarray(signals)
```
mdenker, line 381: Is this for the case that users pass a list of np.array? Then it should be documented this is an option.

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+
+    if not (signals.ndim == 2 and signals.shape[0] == 2):
+        raise ValueError("Required 2 x N input 'signals' array.")
```
mdenker, line 384:
```suggestion
        raise ValueError("The input 'signals' must be of dimensions 2xN.")
```

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+
+    # signal_x and signal_y are (1, N) arrays
```
mdenker, line 386: Not sure if this comment is required.

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+    signal_x, signal_y = np.expand_dims(signals, axis=1)
+
+    coeffs_x, var_x, p_1 = _optimal_vector_arm(signal_x, 1, max_order,
```
mdenker, line 389: I think it would help to have a fex comments in the code on what you are doing.

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+
+    sign, log_det_cov = np.linalg.slogdet(cov_xy)
+    tolerance = 1e-7
+
+    if sign <= 0:
+        raise ValueError(
+            "Determinant of covariance matrix must be always positive: "
+            "In this case its sign is {}".format(sign))
+
+    if log_det_cov <= tolerance:
```
mdenker, line 404: Is the consequence that the result may not be trusted?

elephant/causality/granger.py
```diff
@@ -0,0 +1,438 @@
...
+
+    pairwise_granger
+
+Information Criterion
+~~~~~~~~~~~~~~~
+
+.. autosummary::
+    :toctree: toctree/causality/
+
+    bic
```
mdenker, line 27: I am wondering if `aic` and `bic` are only important for granger, or also for other causality functions (or even far beyond the causality module). My feeling is that they would at some point become part of a larger utility module.

.gitignore
dizcza: I cannot find a reliable regex that would git-ignore everything except toctrees of ASSET and kernel classes. I can revise the problem again, but not in this PR.

elephant/causality/granger.py
dizcza: In this case, we can write `signals : array-like or neo.AnalogSignal` as numpy functions do.

elephant/causality/granger.py
dizcza: In this case, we can write `signals : array-like or neo.AnalogSignal` as numpy functions do.

elephant/causality/granger.py
dizcza: The comment shows that the signals are not just vectors of length N as one might think.

doc/reference/causality.rst
Qkley: It is meant for all of the causality measures so in that case it should be 'Causality'. However, we could also consider reorganisation into [functional connectivity tree](https://www.frontiersin.org/files/Articles/165147/fnsys-09-00175-HTML/image_m/fnsys-09-00175-g001.jpg). Full paper [here](https://www.frontiersin.org/articles/10.3389/fnsys.2015.00175/full).

doc/bib/elephant.bib
Qkley: I have previously used [firstauthorlastname][year4digit][firstwordoftitle] (e.g. ding2006granger), but it would be fine either way.

doc/bib/elephant.bib
Qkley: Revision is not reflected in the DOI (if that is what you are referring to?):
[Revision #127333 http://dx.doi.org/10.4249/scholarpedia.1667](http://www.scholarpedia.org/article/Granger_causality)
[Revision #91329 http://dx.doi.org/10.4249/scholarpedia.1667](http://www.scholarpedia.org/w/index.php?title=Granger_causality&oldid=91329)

elephant/causality/granger.py
Qkley: We can definitely discuss the pros and cons of this. Also, we could have a functional connectivity module encompassing corr, mutual info, cross-corr, Granger in the time-domain and coherence, phase methods, Granger in the Frequency domain as per [this figure](https://www.frontiersin.org/files/Articles/165147/fnsys-09-00175-HTML/image_m/fnsys-09-00175-g001.jpg). Full paper [here](https://www.frontiersin.org/articles/10.3389/fnsys.2015.00175/full).

elephant/causality/granger.py
Qkley: This might need a discussion. Perhaps we can avoid expanding dims in the pairwise and sort it out in the optimal vector arm?

elephant/causality/granger.py
Qkley: This might need a refactor as some of the variable names are too ambiguous and could improve the readability of the code if changed.

elephant/causality/granger.py
Qkley: A comment was raised earlier that logdet has float precision issues. In that case we need to warn the user to keep in mind that the computation might not make sense. This approach is more suitable for batch processing.

elephant/causality/granger.py
Qkley: I think this might require quite a bit of discussion and planning.

elephant/causality/granger.py
dizcza: The idea, at least why I proposed to include such a tolerance, is not due to logdet having float precision issues (I am not aware of any; can the result of `np.logdet` be negative, say `-1e-8`? I don't know), but rather to warn the user when he (accidentally) passes a covariance matrix all values of which are close to zero. Is it better to warn the user immediately or ignore this rare case and proceed with the computation until an unexplainable error, if any, occurs at a later stage of processing?

elephant/causality/granger.py
dizcza: Yes, I'd also suggest having a default value for the max order; something like 10; does it make sense? Or, if you don't want to put any default value, describe a typical range for max_order (for instance, a range of [1, 10]).

dizcza:
I added an example. Feel free to modify it.
Also, though I changed the type of `signals` to array-like, I didn't describe the dimensions.

Qkley:
We are wondering if we should add a Reference section in the docstring as per [lv](https://github.com/NeuralEnsemble/elephant/blob/91c03c46ec8bf69ce80f5dacf9a90e6fd88a36d2/elephant/statistics.py#L317-L337) function in elephant.statistics. If so, we do not follow the outlined equations verbatim, but have an equivalent implementation in the case of Eq 48 in Ding & Chen 2006. Would we need to explicitly state this, or is it ok just to provide a reference?

elephant/causality/granger.py
```diff
@@ -0,0 +1,512 @@
...
+    return optimal_coeffs, optimal_cov_matrix, optimal_order
+
+
+def pairwise_granger(signals, max_order, information_criterion=aic):
+    r"""
+    Determine Granger Causality of two time series
+
+    Parameters
+    ----------
+    signals : array-like or neo.AnalogSignal
```
Qkley, line 362: By definition array-like should include lists. Therefore, we are wondering if it can stay like this?

dizcza: Yes, I made it array-like to specify that the input can be a list or a numpy array.

dizcza:
> We are wondering if we should add a Reference section in the docstring as per [lv](https://github.com/NeuralEnsemble/elephant/blob/91c03c46ec8bf69ce80f5dacf9a90e6fd88a36d2/elephant/statistics.py#L317-L337) function in elephant.statistics. If so, we do not follow the outlined equations verbatim, but have an equivalent implementation in the case of Eq 48 in Ding & Chen 2006. Would we need to explicitly state this, or is it ok just to provide a reference?

I don't quite get your question. Is your question about whether should you explicitly pin down the formula in the doc string or it's fine to include only a reference to the math in the paper? I think we don't have a policy regarding this, do we @kohlerca ? If the formula you used in the implementation differs from the Eq.48, I also don't have a strong opinion on whether we should (1) explain exactly _how_ it differs or (2) just to point out that the implementation is similar but not entirely equivalent to Eq.48 in the `Notes` section of the documentation of a function.

kohlerca:
@dizcza @Qkley 

> I think we don't have a policy regarding this

No, we don't. The idea of the references is that (regardless of showing the equation in the docstring) you provide the paper which your implementation was based on. If you are using some equation that is different, then I would write that custom equation in the docstring text, and add a note in Notes with something like 

"The formula used in this implementation was based/derived/modified from Eq. 48 in _[1]. [briefly explain how and why]"

Since you came with your own equation, I think it is important to give the rationale for such a decision so that other people can understand when going to the original reference and realizing that it differs.

Qkley:
We have addressed @kohlerca comments in the three new commits. Please check them for clarity. Please feel free to let us know if there is anything else to be done.

dizcza (approved):

PR #339, opened 2020-08-04
Feature/inhomogeneous gamma

pbouss:
Added a function to generate an inhomogeneous Gamma process


elephant/spike_train_generation.py
```diff
@@ -650,6 +650,67 @@ def homogeneous_gamma_process(a, b, t_start=0.0 * pq.ms, t_stop=1000.0 * pq.ms,
...
+        a=shape_factor, b=shape_factor*1.*pq.Hz,
+        t_start=0.*pq.s, t_stop=operational_time[-1]*pq.s, as_array=True)
+
+    indices = np.searchsorted(operational_time, spiketrain_operational_time)
+
+    spiketimes = \
+        rate_times[indices - 1] \
+        + rate.sampling_period.simplified.magnitude / (
+            operational_time[indices]-operational_time[indices-1]) * (
+            spiketrain_operational_time - operational_time[indices-1])
```
dizcza, line 705: ## Part 1

This block is quite complicated, as CodeFactor complains. Split the logic into several lines and assign meaningful names. Possible way (but not limited to) of such a split: 
* `var1 = operational_time[indices]-operational_time[indices-1]`
* `var2 = spiketrain_operational_time - operational_time[indices-1]`
* `rate_times[indices - 1] + rate.sampling_period.simplified.magnitude / var1 * var2`

Also, add comments describing why you're adding `rate_times[indices - 1]` and then what `var1` and `var2` are, if necessary.

## Part 2

Are you sure that `indices` are always greater than zero? For example, did you consider a scenario when
```python
>>> np.searchsorted([4,5,6], 3)
0
```
?
Deal with this case or drop a comment describing that such a case is impossible to occur.

elephant/spike_train_generation.py
```diff
@@ -650,6 +650,67 @@ def homogeneous_gamma_process(a, b, t_start=0.0 * pq.ms, t_stop=1000.0 * pq.ms,
...
+    spiketrain : neo.SpikeTrain or np.ndarray
+        Inhomogeneous Poisson process realization, of type `neo.SpikeTrain`
+        if `as_array` is False (default) and `np.ndarray` otherwise.
+
+    Raises
+    ------
+    ValueError
+        If `rate` contains a negative value.
+    """
+
```
dizcza, line 682: add a check if the `rate` is AnalogSignal and raise an error if not + add this in the `Raises` doc section.

Also, add one test per error to trigger them independently.

elephant/spike_train_generation.py
pbouss: For Part 2: I considered this scenario, therefore I padded a zero to the beginning of the operational time array.

pbouss: For Part 1: I tried to make it easier to read and added some comments.

dizcza (approved):

PR #341, opened 2020-08-17
Added a warning in fanofactor function when the input spiketrains vary in their durations

### LINKED ISSUE #336: Fano factor should check that the spiketrains in the passed list are of the same duration
Currently, Fano factor takes the length of each spike train in the spike train list to perform the computation. The only check that is performed is that the number of spikes in each spike train is not equal to zero. In the case of spike trains that are not even in their duration, suggesting trials of varying durations, the current method would return Fano factor value that could be rather meaningless. Therefore, it might be beneficial to perform a check or raise an error in such a case.

dizcza:
Fixes #336

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -271,8 +271,8 @@ def fanofactor(spiketrains, warn_tolerance=0.1 * pq.ms):
         return np.nan
 
     if all(isinstance(st, neo.SpikeTrain) for st in spiketrains):
-        if not isinstance(warn_tolerance, pq.Quantity):
-            raise TypeError("'warn_tolerance' must be a quantity.")
+        if not is_time_quantity(warn_tolerance):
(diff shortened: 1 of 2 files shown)


elephant/statistics.py
```diff
@@ -251,15 +255,35 @@ def fanofactor(spiketrains):
...
     # Compute FF
     if all(count == 0 for count in spike_counts):
-        fano = np.nan
-    else:
-        fano = spike_counts.var() / spike_counts.mean()
+        # empty list of spiketrains reaches this branch, and NaN is returned
+        return np.nan
+
+    if all(isinstance(st, neo.SpikeTrain) for st in spiketrains):
+        if not isinstance(warn_tolerance, pq.Quantity):
```
dizcza, line 274: should be check is_time_quantity

PR #346, opened 2020-09-01
[Feature] Add revised local variability LvR

morales-gregorio:
Hey,

I was reading [Shinomoto et al. 2009](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1000433) where they present this LvR metric. Since I wanted to try it on my data I just went ahead and put it in elephant. I hope the initiative is fine :) 

Additionally to the metric, I modified the order in which the metrics are presented (cv, cv2, lv, lvr) since it did not feel right to separate cv and cv2 as it was before (cv, lv, cv2). Just my opinion, I can change it back if needed.

Also, I modified the notation slightly in the docstrings (nothing breaking I run the tests) to have `CV -> Cv` and `LV -> Lv` which is the way that Shinomoto writes those (and well, he did invent these things). I can also put it back if you don't like that. 

Cheers,
Aitor


elephant/statistics.py
```diff
@@ -389,39 +460,44 @@ def lv(time_intervals, with_nan=False):
 
 
 @deprecated_alias(v='time_intervals')
-def cv2(time_intervals, with_nan=False):
+def lvr(time_intervals, R=5*pq.ms, with_nan=False):
```
dizcza, line 463: deprecated aliases are made for existing functional

don't use them while adding a new function; otherwise, we will never get rid of them

dizcza (commented):
Thank you, Aitor,
looks good to me. Please remove the deprecation decorator from `lvr`.
We'll discuss it on Mon.

elephant/statistics.py
morales-gregorio: Hey!
Right, I just did not know why that was there so I copied it over again hehe
Should be fine now

morales-gregorio:
Hey Danilo,

Thanks for the quick review, I just 'corrected' the python 2 errors by letting a test be ignored, `assertWarns` does not like python<3, but I guess this should be fine since it is how this is dealt with in the other test cases.

Cheers!
Aitor

dizcza (approved):
We'll discuss it at one of the Mondays' meetings

dizcza (approved):
You changed the naming convention of CV to Cv, etc.
We need to stick to it for future requests if any.

PR #350, opened 2020-09-09
tutorials: changed wget to curl for platform compatibility

pbouss:


dizcza (approved):
It ain't much but it's honest work.
Thank you!

PR #353, opened 2020-09-30
Surrogates get arbitrary sampling_rate

morales-gregorio:
Hey all,

I have included the sampling rate to the generation of surrogates, since the current version does not pass it from the original spike train to the surrogate ones (like it does with the t_start or t_stop). 

Hopefully nothing breaks down.

Best,
Aitor


dizcza (approved):
Thank you for noticing that.

A lot of places. This means we never thought of testing this attribute, hm.
OK, even if some spiketrain-generation code is still not covered with the correct sampling rate, we'll gradually fix them once observed.

dizcza:
Apart from the `sampling_rate`, the following additional parameters are recommended to have:
* name
* description
* sort

Cristiano will work on this in the context of provenance tracking.

PR #338, opened 2020-07-26
Implementation of a new function called spike contrast. 

Broxy7:
We implemented a new function called spike_contrast which is capable of calculating the synchrony of spike trains.
I am fairly new to software testing and would appreciate any kind of feedback.
Furthermore this is my first pull request so I also would appreciate any kind of feedback.

PR #354, opened 2020-10-05
spike train synchrony: new function called spike contrast

Broxy7:
As discussed the new pull request for #338 .


dizcza (approved):
Thank you very much for the method and patience during the review phase!

PR #359, opened 2020-10-09
Feature/conditional granger

Qkley:
This is the PR for conditional GC. We would like to have it reviewed. Please let us know if the number of tests for the new function is sufficient. Also, it would be good to check the documentation for compliance with the Elephant standard.

### DIFF
--- a/elephant/causality/granger.py
+++ b/elephant/causality/granger.py
@@ -578,9 +578,13 @@ def conditional_granger(signals, max_order, information_criterion='aic'):
 
     signals_xz = np.vstack([signal_x, signal_z])
 
-    coeffs_xz, cov_xz, p_1 = _optimal_vector_arm(signals_xz, 2, max_order,
+    coeffs_xz, cov_xz, p_1 = _optimal_vector_arm(signals_xz, dimension=2,
+                                                 max_order=max_order,
+                                                 information_criterion=
                                                  information_criterion)
-    coeffs_xyz, cov_xyz, p_2 = _optimal_vector_arm(signals, 3, max_order,
+    coeffs_xyz, cov_xyz, p_2 = _optimal_vector_arm(signals, dimension=3,
+                                                   max_order=max_order,
+                                                   information_criterion=
                                                    information_criterion)
 
     conditional_causality_xy_z = np.log(cov_xz[0, 0]) - np.log(cov_xyz[0, 0])
--- a/elephant/causality/granger.py
+++ b/elephant/causality/granger.py
@@ -72,7 +72,8 @@ from neo.core import AnalogSignal
 
 __all__ = (
     "Causality",
-    "pairwise_granger"
+    "pairwise_granger",
+    "conditional_granger"
 )
 
 
--- a/elephant/causality/granger.py
+++ b/elephant/causality/granger.py
@@ -579,14 +579,12 @@ def conditional_granger(signals, max_order, information_criterion='aic'):
 
     signals_xz = np.vstack([signal_x, signal_z])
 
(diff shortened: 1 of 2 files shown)


elephant/causality/granger.py
```diff
@@ -526,3 +526,73 @@ def pairwise_granger(signals, max_order, information_criterion='aic'):
...
+
+    # Round conditional GC according to following scheme:
+    #     Note that standard error scales as 1/sqrt(sample_size)
+    #     Calculate  significant figures according to standard error
+    length = np.size(signal_x)
+    asymptotic_std_error = 1/np.sqrt(length)
+    est_sig_figures = int((-1)*np.around(np.log10(asymptotic_std_error)))
+
+    conditional_causality_xy_z_round = np.around(conditional_causality_xy_z,
+                                                 est_sig_figures)
```
dizcza, line 596: why do you round the resulting value? does it make sense other than more human-readable?

elephant/causality/granger.py
```diff
@@ -526,3 +526,73 @@ def pairwise_granger(signals, max_order, information_criterion='aic'):
...
+
+    # transpose (N,3) -> (3,N) for mathematical convenience
+    signals = signals.T
+
+    # signal_x, signal_y and signal_z are (1, N) arrays
+    signal_x, signal_y, signal_z = np.expand_dims(signals, axis=1)
+
+    signals_xz = np.vstack([signal_x, signal_z])
+
+    coeffs_xz, cov_xz, p_1 = _optimal_vector_arm(signals_xz, 2, max_order,
```
dizcza, line 581: use keywords
```
dimension=2,
max_order=max_order,.
...
```
same for dimension 3

elephant/causality/granger.py
```diff
@@ -45,6 +45,7 @@
     :toctree: toctree/causality/
 
     pairwise_granger
+    conditional_granger
```
dizcza, line 48: add conditional_granger in `__all__` as well

dizcza (commented):
The implementation and the docs look good, though maybe Cristiano has a better eye for it.

One thing to add:
Currently, you have provided only one test that checks for zero causality of the ground truth data. Add at least 1 test that checks for non-zero causality (simulate the data).

elephant/causality/granger.py
Qkley: We use rounding to avoid tiny negative causality results, when the actual output should be zero. This is also done in the pairwise causality method as well (https://github.com/NeuralEnsemble/elephant/pull/332/commits/e577ce0b891e010d3a3b59ca4a2de775cc5cb493).

dizcza: The rounding in pairwise granger didn't catch my eye.

All right, it'd be nice if you put your answer into a code comment describing "We use rounding to avoid tiny negative causality results, when the actual output should be zero". Such a comment will help future developers to read the code. This is optional.

elephant/causality/granger.py
Qkley: Addressed in commit https://github.com/NeuralEnsemble/elephant/pull/359/commits/9a8d6272b94511ece1fa2a4ebda7e233fb5d0fa5

elephant/causality/granger.py
Qkley: Addressed in commit https://github.com/NeuralEnsemble/elephant/pull/359/commits/bb4d8fe77adcbefa1bd4944c87376c4c1bedcbb5

Qkley:
The comments have been addressed. In particular, I have extended the conditional causality test case to include non-zero conditional causality unit test.

Issue #361, opened 2020-10-16
[Feature] Fast instantaneous_rate for many spike trains 


gyyang:
**Feature**
Improve instantaneous_rate to be faster when processing many spike trains

**Motivation**
It is often necessary to compute the rate of many neurons or many trials, using the same ```t_start, t_stop, sampling_period```. The current code appears rather slow for this purpose. There are many places that can be sped-up by parallelization and by not using Neo format. The following is some code that speeds things up by **~50-60X** for reasonable data size, i.e. ~1000 units, 10,000 spike each. The code is not extensively tested for edge cases, but does have the same output as the original code when ```kernel.sigma``` and ```t_stop - t_start``` are both integer multiples of ```sampling_period```.

**Alternatives**
```
import warnings
import numpy as np
import scipy

from elephant.statistics import optimal_kernel_bandwidth
import elephant.kernels as kernels


def myrate(spiketrain, sampling_period, kernel='auto',
                       cutoff=5.0, t_start=None, t_stop=None, trim=False,
                       center_kernel=True):
    """
    Estimates instantaneous firing rate by kernel convolution.

    Modified by gryang from elephant.statistics.instantaneous_rate. Much
    faster for many spike trains.

    Parameters
    ----------
    spiketrain : list of lists of spike times in seconds
    sampling_period : float
        Time stamp resolution in seconds of the spike times. The same resolution will
        be assumed for the kernel.

    The rest are the same as elephant.statistics.instantaneous_rate,
    abbreviated here.
    """

    if kernel == 'auto':
        kernel_width_sigma = None
        if len(spiketrain) > 0:
            kernel_width_sigma = optimal_kernel_bandwidth(
                spiketrain.magnitude, times=None, bootstrap=False)['optw']
        if kernel_width_sigma is None:
            raise ValueError(
                "Unable to calculate optimal kernel width for "
                "instantaneous rate from input data.")
        kernel = kernels.GaussianKernel(kernel_width_sigma * spiketrain.units)
    elif not isinstance(kernel, kernels.Kernel):
        raise TypeError(
            "'kernel' must be either instance of class elephant.kernels.Kernel"
            " or the string 'auto'. Found: %s, value %s" % (type(kernel),
                                                            str(kernel)))

    # TODO: do the single spike train case
    n_spiketrain = len(spiketrain)  # Number of spike trains

    # main function:
    nbins = int((t_stop - t_start) / sampling_period) + 1
[...]

PR #362, opened 2020-10-16
Faster instantaneous rate with a bugfix (rate wasn't centered)

### LINKED ISSUE #360: [Bug] instantaneous_rate estimated rates appear shifted by one time step
Rate returned by the instantaneous_rate function appear to be (incorrectly) shifted by one time step.

**To Reproduce**
```
import numpy as np
from neo.core import SpikeTrain
from quantities import s, ms
spiketrain = SpikeTrain(np.array([-0.0001, 0, 0.0001])*s, t_start=-1, t_stop=1.0)
rate = instantaneous_rate(spiketrain, sampling_period=20*ms, kernel=GaussianKernel(50*ms))
print(rate.times[np.argmax(rate)])
```
**Expected behavior**
Since the 3 spikes are centered around 0, I would expect the estimated rate to peak at t=0. However, the rate returned peak at -1 * sample_period.

**Environment**
 - OS (e.g., Linux): macOS
 - How you installed elephant (`conda`, `pip`, source): pip
 - Python version: 3.7
 - `neo` python package version: 0.8.0
 - `elephant` python package version: 0.8.0

dizcza:
* Fixes #360 #361 #363 
* Clamp instantaneous rates to zero from below, because a convolution of non-negative vectors is non-negative by definition, and the warning is not reachable. Previously, the user saw annoyingly small negative values due to the FFT error tolerance.
* Vectorized `instantaneous_rate` function, as described in #361. The speedup is **~25-30X**, compared to the previous implementation. The bottleneck was rescaling quantities, as always.
* Initially added `concatenate=True` parameter to the instantaneous rate function is removed in [this commit](https://github.com/NeuralEnsemble/elephant/pull/362/commits/44d3964d46c9644424d4c425981d67ac2c712e35) to not overwhelm yet already complex API of such simple functionality. Another vote to remove the old behavior - when the input is a list of spiketrains - is described in #363: from the users perspective, when the input is a list of spiketrains and the output is only 1d AnalogSignal, it's more intuitive (and I fully agree) to return an averaged instantaneous rate. Our previous approach, however, returns the sum. Now there won't be any misuse because the output is a 2d analog signal with the same number of channels as the number of input spiketrains. Such "clean" API additionally provides a simple way to get not only the mean or the sum of instantaneous rates, but also the STD, as requested in #363.
* Renamed `spiketrain -> spiketrains` to let users know that this function is optimized now to work with a list of spiketrains.

The fix changes the rate profile because previously the time domain of a kernel was incorrectly sampled - it didn't mind the fact that `t_arr` must contain the origin point (or the median of a kernel in the general case) if `center_kernel` was set to True (see the plots below).

Needs a discussion:
* From the plots below one can easily spot that between `rate.t_stop` and the input `spiketrain.t_stop` there are **two** sample periods, not one, as expected. I think there should be only one lag, which means we have another unreported bug. We even have a test that checks `assertEqual(len(rate), spiketrain.t_stop / sampling_period)` but this is wrong: between `0` and `t_stop` sampled at `sampling_period` there should be `spiketrain.t_stop / sampling_period + 1` points or `spiketrain.t_stop / sampling_period` intervals. The comment from my previous large assessment of kernels states stat "instantaneous rate is a function of intervals" - I remember that I wrote this comment to "fit a theory with the results" and therefore this should be revised.
* Fixed an unreported bug/wrong documentation/wrong function implementation of `ExponentialKernel`. Note that the edge cases of `ExponentialKernel` diverge from `AlphaKernel` - that has been the case for many years in Elephant, probably since the beginning. I assume that `ExponentialKernel` documentation was correct, but not the implementation - need to discuss.
* The author of #361 additionally requested support for numpy arrays without quantities. This can be implemented by providing an additional parameter `units=None` by default and, if the input is a numpy array, the `units` are used to construct the output AnalogSignal rate.


dizcza:
## Differences with the previous implementation for a gaussian kernel
### Typical cases (small sampling period)
sampling_period-1
sampling_period-10
sampling_period-100

### Non typical cases (large sampling period)
sampling_period-200
sampling_period-445
sampling_period-1000

## What happens with the kernel sigma is small and the sampling period is large?
sampling_period-200
sampling_period-1000
The first obvious observation is that the previous implementation results in a zero vector estimate. But why now it's overestimated?

## AlphaKernel convolved with a single spike at the origin
AlphaKernel

PR #379, opened 2020-11-05
Release v0.9.0

dizcza:


apdavison:
Hi @dizcza we are ready to release Neo 0.9.0 from the master branch, but before doing so I wanted to check with you that the current Neo master doesn't break anything in Elephant.

dizcza:
Hi @apdavison 
2 tests fail in test_neo_tools.py with the latest neo master branch:
`test__extract_neo_attrs__spiketrain_parents_childfirst_array`
`test__extract_neo_attrs__spiketrain_parents_parentfirst_array`
These tests are 6 years old. I'm doing a video recording right now, I'll look into the issues in a few hours. Help is appreciated since I already see that the quality of test code is not that great and + it's neo tools, in which I rarely look, so it could be hard for me to realize what is going on.
Maybe @JuliaSprenger have insights?

samuelgarcia:
The test you mention use ChannelIndex and Units.
I think you should remove theses classes in your tests suite.
But this should still work at least for neo 0.9

dizcza:
Are ChannelIndex and Units deprecated now?

apdavison:
yes - for more information see https://neo.readthedocs.io/en/latest/releases/0.9.0.html

JuliaSprenger:
Yes, they are deprecated and not used any more by the IOs. For compatiblity reasons the classes will still be included in version 0.9.0, but should not be used any more.

dizcza:
I removed ChannelIndex-related tests from neo utils. Now it works.
Thanks!

Issue #381, opened 2020-11-09
[Bug] out-of-bounds error for spike_triggered_phase of sliced signals


rgutzen:
**Describe the bug**
When slicing an analogsignal there might be a gap between the last sample point and `t_stop` from as much as 1/sampling_rate.
In the `spike_triggered_phase()` function, spike times are related to the nearest sample point in the analogsignal. When a spike, however, falls into the gap between the last sample point and `t_stop` it might be rounded so that the function tries to get an item that is just outside of the analogsignal slice.

**To Reproduce**
```
asig = neo.AnalogSignal(np.arange(20)*pq.dimensionless,
                        sampling_rate=1*pq.Hz,
                        t_start=0*pq.s)
hilbert = elephant.signal_processing.hilbert(asig)

st = neo.SpikeTrain(np.array([1,4,5,8,9.5,12,17])*pq.s, 
                    t_stop=20*pq.s)

hilbert_slice = hilbert.time_slice(0*pq.s, 10*pq.s)
st_slice = st.time_slice(0*pq.s, 10*pq.s)

phases, magnitude, time = elephant.phase_analysis.spike_triggered_phase(
                hilbert_slice, st_slice, interpolate=False)
```

```
~/anaconda3/envs/ewsk/lib/python3.7/site-packages/elephant/phase_analysis.py in spike_triggered_phase(hilbert_transform, spiketrains, interpolate)
    141             # positive if spike time is later than sample point
    142             print(spike_i, ind_at_spike_j)
--> 143             dv = spiketrain[sttimeind[spike_i]] - times[ind_at_spike_j]
    144 
    145             # Make sure ind_at_spike is to the left of the spike time

~/anaconda3/envs/ewsk/lib/python3.7/site-packages/quantities/quantity.py in __getitem__(self, key)
    372     @with_doc(np.ndarray.__getitem__)
    373     def __getitem__(self, key):
--> 374         ret = super(Quantity, self).__getitem__(key)
    375         if isinstance(ret, Quantity):
    376             return ret

IndexError: index 10 is out of bounds for axis 0 with size 10
```

A quick fix would be to simply check for this specific case with an if-statement and if necessary reduce the index by 1.
Or is there a more elegant solution?

dizcza:
> A quick fix would be to simply check for this specific case with an if-statement and if necessary reduce the index by 1.

Such a fix indicates the wrong logic in the first place.

[...]

PR #382, opened 2020-11-10
Fixing index selection in spike_triggered_phase (see issue #381)

rgutzen:
This fixes replaces the mechanism of finding the indices of the array of sample times left to each spike.
Previously, this was done by calculating the spike position in the times-array via the sampling rate, rounding, and checking if the sampling point is actually to the left of the spike. This approach is a bit convoluted and may cause errors in some conditions (issue #381).
The PR replaces this with the numpy function `searchsorted()` which fixes the bug, is more straightforward, and computes faster.

### DIFF
--- a/elephant/phase_analysis.py
+++ b/elephant/phase_analysis.py
@@ -129,8 +129,10 @@ def spike_triggered_phase(hilbert_transform, spiketrains, interpolate):
         times = hilbert_transform[phase_i].times
 
         # Find index into signal for each spike
-        ind_at_spike = np.searchsorted(times, spiketrain[sttimeind],
-                                       side='right').magnitude - 1
+        ind_at_spike = (
+            (spiketrain[sttimeind] - hilbert_transform[phase_i].t_start) /
+            hilbert_transform[phase_i].sampling_period). \
+            simplified.magnitude.astype(int)
 
         # Append new list to the results for this spiketrain
         result_phases.append([])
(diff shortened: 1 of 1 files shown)


dizcza (commented):
Nice at the first glance, but no.
numpy searchsorted is used for arrays that have different time intervals, and it also does not solve the problem with inserting 2.999999 ms spike into `[0, 3, 6]` intervals.
Compared to `np.floor`, it does some nice things by handling values outside of the interval, but such corner cases must be checked before the execution and raise an error accordingly.
Therefore, your solution is equivalent to np.astype(int) (or np.floor) but with a greater algorithmic complexity.

rgutzen:
What problem do you mean with inserting 2.9 ms spike into `[0,3,6]`? The new code has the same behaviour as the previous to select the sample point to the left. There is an optional additional step if `interpolate=True` in the function to relate to better relate it to it's right sample point.

The alternative to `searchsorted()` would be
```
ind_at_spike = ((spiketrain - hilbert_transform.t_start) / hilbert_transform.sampling_period).simplified.magnitude.astype(int)
```
which yields the same results, and is a bit slower.

Corner cases are anyway excluded in the function (line 125/126) 
```
# Take only spikes which lie directly within the signal segment -
        # ignore spikes sitting on the last sample
        sttimeind = np.where(np.logical_and(
            spiketrain >= start[phase_i], spiketrain < stop[phase_i]))[0]
```

rgutzen:
I switched back to using the calculation via the sampling period and fixed an additional bug as the phases we stored slightly differently when interpolating or not.
In the previous unhandled case that there is a spike occurring after the last sample time and t_stop, and interpolate=True, the interpolation for this spike time is now skipped since there is no right-hand sample point to interpolate between.

dizcza (approved):
Okay, so there are two bugfixes.

> What problem do you mean with inserting 2.9 ms spike into [0,3,6]? 

I wanted to describe a scenario with floating-point precision issues when the spike time is 2.99999999 ms but should be strictly 3.0 ms. Because it's not strictly 3.0, the current fix, just taking the integer part, is not entirely correct, which may be argued upon. More "smart" way is to round to the nearest integer, given the tolerance, as we're currently doing in BinnedSpikeTrain object creation. With the tolerance factor (`~1e-8`), the spike above will have index 1 instead of 0 (current approach).

But let's separate these issues. Keep track if it happens within your tutorial, since the spike trains in 2020_vX nix files have precision errors such as 1e-7 difference in t_start/stop across trials.

I'll merge as soon as the tests pass.

rgutzen:
Yes, floating-point errors like this would lead to a shift in the corresponding sample point. However, just rounding (even with a low tolerance factor) can again introduce the issue of trying to access the last+1 sample point. So, there is some further thinking required. I agree, let's do this one at a time.

PR #389, opened 2020-12-04
Fix LvR values from being off when units are in seconds

morales-gregorio:
Hi! 

I noticed (and fixed) a bug of my own making in the LvR function. When the time units of the time_intervals were not milliseconds the output was way off. I fixed it, added a warning to when there are no explicit units and updated the tests to also check for this. 

I believe this is not a problem in the other ISI variation metrics (Cv, Lv) because they do not have the R parameter which has time units itself. So the issue should be settled with this simple fix.

All best,
Aitor


dizcza (approved):
Thanks and welcome to the units hell.

I'll wait for the tests build and merge then.

PR #385, opened 2020-11-25
Feature/plv

ojoenlanuca:
The phase_analysis module of elephant has yet, just one function, namely the spike_triggered_average() also known as STA. We want to increase this number of functions by implementing the phase_locking_value() or PLV and several more. 
The current version of PLV takes array-objects as input. 

Open Point of Discussion:
Which representations of the input-phases should be available?
neo.AnalogSignal, List of List, ..., etc.

PR #392, opened 2020-12-16
[Feature] pairwise phase consistency

TRuikes:
Hi! 
In our lab we are using 'pairwise phase consistency' as a metric to measure spike-LFP interactions. The conventional method 'Phase Locking Value' can give faulty output if there are few spikes, and PPC improves on this. ([function as implemented by authors](https://www.fieldtriptoolbox.org/reference/ft_connectivity_ppc/). In short you compute the phase of an lfp during spike times (spike_triggered_phase) and then compute the angular distance between each pair of phases.

The output for the function `phase_analysis.spike_triggered_phase` can be directly fed into this function. While PPC can also be used to compute synchony between two EEG/LFP signals, I think there are allready other (better) measures for this and we can leave it at this simple implementation for now.

It can be a nice addition, let me know if you want to add this and if so what you would like me to change/add. In addition I suggest a tutorial on 'spike-field' functionality of elephant, briefly describing basic LFP processing steps (filtering, hilbert transform) and possible metrics/functions (spike_triggered_phase, sta.spike_field_coherence, sta.spike_triggered_average (<-that's PLV)).


dizcza:
Hi Thijs,
we're currently advancing phase statistics (see #385), and we do have PPC on the to-do list, thanks for bringing this!
We're also porting the documentation to the new BibTex citation documentation format, so I've updated your docs & reference in https://github.com/TRuikes/elephant/pull/1. You can merge my PR here. I still don't see you in the authors list of Elephant!

The tests are missing. If you're lazy to write them, I can come up with something, though I'm not an expert in this field.

TRuikes:
Allright, thanks for the quick response! I'll think of some tests (wasn't too lazy, just didn't want spend time before knowing if you were going to use it ;) ) and add myself to authors.

The authors made impoved versions of PPC (this one they call 'ppc0' and then there are two more, I think). Long story short, when computing PPC between pairs, they dont want to draw samples from the same trial (their full reasoning [written here](http://europepmc.org/article/PMC/3394239#impact), due to spiking properties of a neuron). To compute this I'd have to give information on trial nr per phase. I think I'll add an optional keyword `trial_ids` whith same dimensions as `phases` with a number per trial. Then another keyword to select `ppc0` or `ppc1`.

I'm going to work on this next week again (monday), so I'll push something around then.

dizcza:
> I think we don't need to multiply by 2. The matrix dot_prod is completely filled, so each pair 'i','j' is computed twice.

Yes, you're right, I didn't think of that.

Regarding your previous comment about ppc1, if you think that it could take much time or you're unsure how to implement this additional feature that potentially (!) improves the estimate, it's better to come up with another pull request that addresses the improvement after this PR is merged.

elephant/phase_analysis.py
```diff
@@ -235,27 +244,35 @@ def pairwise_phase_consistency(phases):
...
+    if method not in ['ppc0']:
+        raise ValueError('For method choose out of: ["ppc0"]')
+
+    phase_array = np.hstack(phases)
+    n_trials = phase_array.shape[0]  # 'spikes' are 'trials' as in paper
+
+    # Compute the distance between each pair of phases using dot product
+    # Optimize computation time using array multiplications instead of for
+    # loops
+    p_cos_2d = np.tile(np.cos(phase_array), reps=(n_trials, 1))  # TODO: optimize memory usage
```
dizcza, line 256: right, `numpy.broadcast_to` is recommended since it does not consume memory as tile does.

elephant/phase_analysis.py
```diff
@@ -235,27 +244,35 @@ def pairwise_phase_consistency(phases):
...
-        # and should not be included)
-        np.fill_diagonal(dot_prod, 0)
-        ppc = 2 / (n_trials * n_trials - n_trials) * np.sum(dot_prod)
-        result_ppc.append(ppc)
-
-    return result_ppc
+    if method not in ['ppc0']:
+        raise ValueError('For method choose out of: ["ppc0"]')
+
+    phase_array = np.hstack(phases)
```
dizcza, line 250: are you sure that all phases have the same length?

elephant/phase_analysis.py
TRuikes: I was thinking about specifying datatype, using np.float32 would allready reduce half of the memory used. I noticed when trying this on large spiketrains that it consumes too much RAM (nspikes^2). I don't know if broadcast_to will use less memory than tile, I'm not planning on testing it, unless you think it's worth the time.

elephant/phase_analysis.py
TRuikes: I'm sure all phases _do not_ have the same length :p. I implemented this method (slightly against your earlier advice) after I discovered that simply putting in a whole spiketrain for a 2h session takes too much RAM. As I had to cut up my spiketrain anyway, it felt like it made sense to do so in trials. As such the default input for the function will be (if you agree) a list of arrays, where each array is a trial of size `n_phases`. For the default version of PPC we can simply append all these phases, and compute the consistency, and I'll leave it at this for this PR.

Then in a future PR we can easily extend this function with PPC1, by selecting indices of the input phases, such that we only average between pairs of phases between trials (i.e. PPC1).

elephant/phase_analysis.py
dizcza: According to numpy's doc, broadcast_to should use only N memory instead of N^2, as in tile.
Because it does not allocate memory but rather returns a read-only view with the given shape. And it's everywhere stated through the numpy doc that broadcast_to should be preferred for exactly this reason - memory.

dizcza: Therefore, we can use the datatype returned by `cos`, which is double float by default.

TRuikes: ah I see, since its a view it wont actually construct a new array. And would we save memory if we add `dtype=np.float32` to `np.multiply` in the following line? Because the `dot_prod` is going to be a new matrix of size N^2 (not a view like cos).

elephant/test/test_phase_analysis.py
```diff
@@ -202,5 +202,131 @@ def test_regression_269(self):
         self.assertEqual(len(phases_noint[0]), 2)
 
 
+class PairwisePhaseConsistencyTestCase(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):  # Note: using setUp makes the class call this
```
TRuikes, line 208: Do you prefer `setUp` over `setUpClass`?

elephant/phase_analysis.py
dizcza: ah, yes, so we do need to have N^2 after all.
Then yes, I'd cast everything to float32

elephant/test/test_phase_analysis.py
dizcza: They serve different use cases: `setUp` is run at each test while `setUpClass` is run only once at the test class initialization. Since you're initializing read-only values only once, `setUpClass` is preferred for the sake of runtime. If the initialization is fast, either way is fine.

dizcza:
Let me know when it's ready for review.

TRuikes:
It's ready for review! I'm allready testing it on a (large) dataset and will let you know if something comes up.

TRuikes:
If I'm supposed to do something please let me know

dizcza (approved):
I created a PR again https://github.com/TRuikes/elephant/pull/2

There was a todo "handle nans". Nans can occur iff the sum is zero and n_trials=1. In such cases, nans are the valid output.

Once the PR is merged, I merge this.

Moritz-Alexander-Kern:
Hi @TRuikes ,
To keep the whole thing neatly organized, it seems to make sense to merge this as one PR. 
So I suggest to continue on this after september, otherwise we would have to scatter the PPC1extension and additional tests across multiple PRs.
Nice work so far, looking forward to continue working on this with you.

PR #393, opened 2020-12-23
Granger causality tutorial notebook

Qkley:
I have implemented the GC tutorial notebook focusing on pairwise and conditional causality methods. Please feel free to suggest changes and additions.

### DIFF
--- a/doc/_templates/autosummary/class.rst
+++ b/doc/_templates/autosummary/class.rst
@@ -5,7 +5,7 @@
 .. autoclass:: {{ objname }}
    :members:
    :inherited-members:
-   :special-members: __call__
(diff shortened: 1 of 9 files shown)


dizcza:
There is an issue with the diagram plot (check the built doc, link will be posted later), that is circles are (partially or completely) invisible. I'm thinking about whether we should create a folder with such images for additional materials. For now, just fix the diagram.

A description that the direct and indirect non-zero causality values are expected at the end of the tutorial is missing. Probably, you wanted to show that in this case a non-zero value is expected (but the value itself does not matter unless we want to compare it with another causality value).

dizcza (approved):

PR #395, opened 2021-01-08
Improve memory efficiency of _create_sparse_matrix in BinnedSpikeTrain class

morales-gregorio:
Hi!

I was having some memory problems when running the synchrotool and @Kleinjohann helped me figure out that the highest memory usage was actually happening in the binning. 

We then together made some modifications to _check_neo_spiketrain() that improve the memory efficiency quite a bit. Using this test script for memory profiling:
```
import quantities as pq
import elephant.conversion as conv_old
import elephant.conversion_new as conv_new
from elephant.spike_train_generation import homogeneous_poisson_process as poisson
import ctypes
libc = ctypes.CDLL("libc.so.6")


@profile
def old_binning(sts):
    conv_old.BinnedSpikeTrain(sts, binsize=1*pq.ms, tolerance=1e-8)


@profile
def new_binning(sts):
    conv_new.BinnedSpikeTrain(sts, binsize=1*pq.ms, tolerance=1e-8)


if __name__ == "__main__":
    sts = []
    for i in range(2000):
        st = poisson(10*pq.Hz, t_start=0*pq.ms, t_stop=1000000*pq.ms)
        sts.append(st)

    # Run the old function
    old_binning(sts)

    # Force python to collect garbage, linux-specific
    libc.malloc_trim(0)

    # Run the new function
    new_binning(sts)
    libc.malloc_trim(0)
```

Which we run using the python memory-profiler as such:
```
mprof run --include-children --multiprocess --interval 0.01 --python python binning_linewise_memtest.py --noreload
```

The output is:
Figure_1

So not only did we reduce the highest memory usage spike, but also the runtime went down. This output is possibly exaggerated because I am using a rather large list of spiketrains (2000) as an example, but certainly also the case in smaller runs. With 100 spiketrains it looks like this:
Figure_2

So major improvements to memory efficiency occur in both larger and smaller settings.

No major IO feature was changed, only the dtypes of the arrays, which were also hard coded before (but only in the last step, hence the memory spike). Please have a look and a nice weekend!
Aitor & Alex K

### DIFF
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -1070,6 +1070,10 @@ class BinnedSpikeTrain(object):
             sparse_matrix = sps.csr_matrix(spiketrains, dtype=np.int32)
             return sparse_matrix
 
+        # Get dtype from largest index
+        shape = (len(spiketrains), self.n_bins)
+        numtype = sps.sputils.get_index_dtype(maxval=max(shape))
+
         row_ids, column_ids = [], []
         # data
         counts = []
@@ -1089,21 +1093,24 @@ class BinnedSpikeTrain(object):
             valid_bins = bins[bins < self.n_bins]
             n_discarded += len(bins) - len(valid_bins)
             f, c = np.unique(valid_bins, return_counts=True)
+            c = c.astype(numtype)
             column_ids.append(f)
             counts.append(c)
-            row_ids.append(np.repeat(idx, repeats=len(f)))
+            row_ids.append(np.repeat(idx, repeats=len(f)).astype(numtype))
+            del f, c
 
         if n_discarded > 0:
             warnings.warn("Binning discarded {} last spike(s) of the "
                           "input spiketrain".format(n_discarded))
 
-        counts = np.hstack(counts)
-        row_ids = np.hstack(row_ids)
-        column_ids = np.hstack(column_ids)
+        # Stack arrays and ensure dtype
+        counts = np.hstack(counts).astype(np.int32)
+        column_ids = np.hstack(column_ids).astype(numtype)
+        row_ids = np.hstack(row_ids).astype(numtype)
 
         sparse_matrix = sps.csr_matrix((counts, (row_ids, column_ids)),
-                                       shape=(len(spiketrains), self.n_bins),
-                                       dtype=np.int32, copy=False)
+                                       shape=shape, dtype=np.int32, copy=False)
+
         return sparse_matrix
 
 
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -1066,6 +1066,7 @@ class BinnedSpikeTrain(object):
 
         """
 
+        # The data type for numeric values
         data_dtype = np.int32
 
         if not _check_neo_spiketrain(spiketrains):
@@ -1077,7 +1078,9 @@ class BinnedSpikeTrain(object):
         # (this is the same dtype that will be used for the index arrays of the
         #  sparse matrix, so already using it here avoids array duplication)
         shape = (len(spiketrains), self.n_bins)
-        numtype = sps.sputils.get_index_dtype(maxval=max(shape))
+        numtype = np.int32
+        if max(shape) > np.iinfo(numtype).max:
+            numtype = np.int64
 
         row_ids, column_ids = [], []
         # data
@@ -1088,9 +1091,9 @@ class BinnedSpikeTrain(object):
         scale_units = 1 / self._bin_size
         for idx, st in enumerate(spiketrains):
             times = st.magnitude
-            times = times[(times >= self._t_start) & (
+            bins = times[(times >= self._t_start) & (
                 times <= self._t_stop)] - self._t_start
-            bins = times * scale_units
+            bins *= scale_units
 
             # shift spikes that are very close
             # to the right edge into the next bin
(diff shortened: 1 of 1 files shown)


apdavison:
What happens if you run the new binning first then the old binning?

morales-gregorio:
Hi!

@Kleinjohann included some consistency changes via a PR to my branch, functionality should not be changed by that.

Thanks for having a look @apdavison , that is indeed an interesting question.

I run the new and old binning 10 times one after the other on the exact same 2000 spiketrains, and the updated PR code. As you can see there is some variability in runtimes and in memory usage, but the new version is always much more memory efficient. 
Figure_3

The runtime improvement appears to be gone in this fresh runs (not because the changes by @Kleinjohann, I saw this also without his refactored code). Anyway, the most important thing is that the new code does not take much more time to run than the old code.

I hope this helps!
Best,
Aitor

dizcza:
Thanks for the perfplots, Aitor.

There are several issues with your test code and proposed improvement:

1. What is your random seed? Assumed 1.
2. Don't `del f, c` - it's not C/C++, where the memory is guaranteed to be deallocated after this call. In Python, it's a bad practice. Avoid it. The garbage collector will do the job perfectly. What `del a` actually means is
  You: "Dear PVM (python virtual machine), I don't need these objects any longer. You _MAY_ delete them."
  PVM: "Don't worry, I'll decide it on my own. (Stupid humans.)"

The rest you'll find in the code comments.

Regarding the performance improvements:
1. There is no speedup gain (2.14 sec vs 2.26 sec). Tested locally with kernprof. No MPI.
2. The memory footprint is indeed always less than the original implementation: 307.188 MB vs 383.957 MB. When the order is changed, that is old_binning is run after the new_binning, the difference becomes even larger.

I cannot push to your private branch (why didn't you push in INM-6/elephant?). Therefore, I post here the code I agree on. Please substitute my version for yours.

```python

    def _create_sparse_matrix(self, spiketrains):
        """
        Converts `neo.SpikeTrain` objects to a sparse matrix
        (`scipy.sparse.csr_matrix`), which contains the binned spike times, and
        stores it in :attr:`_sparse_mat_u`.

        Parameters
        ----------
        spiketrains : neo.SpikeTrain or list of neo.SpikeTrain
            Spike trains to bin.

        """

        # Array data type for data (how many spikes in a bin) and indices
        # (rows and columns) must be the same.
        data_dtype = np.int32

        if not _check_neo_spiketrain(spiketrains):
            # a binned numpy array
            sparse_matrix = sps.csr_matrix(spiketrains, dtype=data_dtype)
            return sparse_matrix

        row_ids, column_ids = [], []
        # data
        counts = []
        n_discarded = 0

        # all spiketrains carry the same units
        scale_units = 1 / self._bin_size
        for idx, st in enumerate(spiketrains):
            times = st.magnitude
            bins = times[(times >= self._t_start) & (
                times <= self._t_stop)] - self._t_start
            bins *= scale_units

            # shift spikes that are very close
            # to the right edge into the next bin
            bins = round_binning_errors(bins, tolerance=self.tolerance)
            valid_bins = bins[bins < self.n_bins]
            n_discarded += len(bins) - len(valid_bins)
            f, c = np.unique(valid_bins, return_counts=True)
            # f inherits the dtype np.int32 from bins, but c is created in
            # np.unique with the default int dtype (usually np.int64)
            c = c.astype(data_dtype)
            column_ids.append(f)
            counts.append(c)
            row_ids.append(np.repeat(idx, repeats=len(f)).astype(data_dtype))

        if n_discarded > 0:
            warnings.warn("Binning discarded {} last spike(s) of the "
                          "input spiketrain".format(n_discarded))

        # Stacking preserves the data type. In any case, while creating
        # the sparse matrix, a copy is performed even if we set 'copy' to False
        # explicitly (however, this might change in future scipy versions -
        # this depends on scipy csr matrix initialization implementation).
        counts = np.hstack(counts)
        column_ids = np.hstack(column_ids)
        row_ids = np.hstack(row_ids)

        sparse_matrix = sps.csr_matrix((counts, (row_ids, column_ids)),
                                       shape=(len(spiketrains), self.n_bins),
                                       dtype=data_dtype, copy=False)

        return sparse_matrix
```

Kleinjohann:
Hi Danylo!

Concerning the copying in the scipy sparse matrix creation: First, a `scipy.sparse.coo_matrix` is created since we're inputting the data in that format. In that creation we should be successfully avoiding the copying of input arrays. This matrix is then converted to a `scipy.sparse.csr_matrix` which involves the allocation of new arrays as you correctly stated.

> Not sure:
> 
> 1. scipy utils get_index_dtype returns either int32 or int64. Cannot imagine the indices being larger than int32. Have this happened to you? This could happen for 50 days recording session, stored in seconds, with 1 ms bins. Sounds possible but not probable. Anyway, if you do want to support such cases, go ahead - revert my code (only this part) to your implementation.

Assuming a binsize of 1/30 ms which is a common sampling frequency the minimal recording length requiring an int64 index array goes down to 3.85 days. I agree that it is quite unlikely that users want to bin such a huge chunk of data, I originally included it because allowing for this should not come with any noticeable overhead. However I agree that your version is a bit easier to read now.

dizcza:
Hi Alex,

> First, a scipy.sparse.coo_matrix is created since we're inputting the data in that format. In that creation we should be successfully avoiding the copying of input arrays.

Interesting, I didn't know that. Is it possible to initialize the CSR matrix directly, omitting a COO step?
Feel free to modify the comment then.

> Assuming a binsize of 1/30 ms which is a common sampling frequency the minimal recording length requiring an int64 index array goes down to 3.85 days. I agree that it is quite unlikely that users want to bin such a huge chunk of data, I originally included it because allowing for this should not come with any noticeable overhead. However I agree that your version is a bit easier to read now.

OK, you can write something like this
```python
shape = ...
numtype = np.int32
if max(shape) > np.int32:
    numtype = np.int64
```
because that's what get_index_dtype essentially doing.

morales-gregorio:
Hi there, 

I have incorporated the changes to the branch. Sorry for not having worked on a branch of INM-6/elephant, I did not realise how working on my own fork would be complicating things.

Hopefully, all the requests are addressed. I have copied the code @dizcza posted on his first comment plus the suggested code from the previous comment. 

Let me know if any further changes are needed. I re-run the memory profiling and it looks very similar, so the changes have not compromised the memory efficiency improvements. As for the seed, I have not fixed the seed, but my multiple iterations run on the same spiketrains, so the memory requirements would remain comparable.

Best,
Aitor

dizcza (approved):

dizcza:
please revert my changes back to
```python
            times = times[(times >= self._t_start) & (
                times <= self._t_stop)] - self._t_start
            bins = times * scale_units
```
one test is failing.
Sorry about that.

morales-gregorio:
No problem, must be some quantities perk, the code did run normally for profiling with the code

PR #396, opened 2021-01-21
Joint-ISI dithering: fixed a bug regarding first ISI bin

pbouss:
The first ISI bin was giving a zero probability due to an error in the combination of the cumulated sum and the padding.


dizcza:
Gosh! I spot `np.r_[]` in your code!

dizcza (approved):

PR #398, opened 2021-01-29
Account for unidirectional spiketrain->segment links in synchrofact deletion

Kleinjohann:
See https://github.com/NeuralEnsemble/python-neo/issues/939 for an in-depth description and a minimal example.

This ensures that no error is raised for neo structures with unidirectional uplinks (which can still happen even if this is changed in neo, e.g. if people use an older version or modify the structure manually). The user is warned about the unidirectional link.

This also fixes a bug which occurred when the Synchrotool was initialised with the actual spiketrain list of a segment, so `Synchrotool(segment.spiketrains, ...)`. Due to the execution order in the in-place deletion, the spiketrain was replaced in the list before its index was extracted, resulting in the same error as above.

### DIFF
--- a/elephant/spike_train_synchrony.py
+++ b/elephant/spike_train_synchrony.py
@@ -341,7 +341,6 @@ class Synchrotool(Complexity):
             if mode == 'extract':
                 mask = np.invert(mask)
             new_st = st[mask]
-            spiketrain_list[idx] = new_st
             if in_place:
                 segment = st.segment
                 if segment is None:
@@ -368,6 +367,7 @@ class Synchrotool(Complexity):
 
                     # st found in group, replace with new_st
                     group.spiketrains[idx] = new_st
+            spiketrain_list[idx] = new_st
 
         return spiketrain_list
 
--- a/elephant/spike_train_synchrony.py
+++ b/elephant/spike_train_synchrony.py
@@ -356,7 +356,11 @@ class Synchrotool(Complexity):
                                   "input list spiketrains has a "
                                   "unidirectional uplink to a segment in "
                                   "whose segment.spiketrains list it does not "
-                                  "appear.")
+                                  "appear. Only the spiketrains in the input "
+                                  "list will be replaced. You can suppress "
+                                  "this warning by setting "
+                                  "spiketrain.segment=None for the input "
+                                  "spiketrains.")
 
                 block = segment.block
                 if block is not None:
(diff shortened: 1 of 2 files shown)


elephant/spike_train_synchrony.py
```diff
@@ -341,33 +342,37 @@ def delete_synchrofacts(self, threshold, in_place=False, mode='delete'):
...
+                    new_index = self._get_spiketrain_index(
+                        segment.spiketrains, st)
+                    segment.spiketrains[new_index] = new_st
+                except ValueError:
+                    # st is not in this segment even though it points to it
+                    warnings.warn(f"The SpikeTrain at index {idx} of the "
+                                  "input list spiketrains has a "
+                                  "unidirectional uplink to a segment in "
+                                  "whose segment.spiketrains list it does not "
+                                  "appear.")
```
dizcza, line 359: a typical elephant user might not digest such a message.
How about adding a description (on top of your sentence) about how to actually _fix_ the warning? Like "update the neo to the most recent version" or do/don't do things like `obj.segment = self.segment` manually.

elephant/spike_train_synchrony.py
```diff
@@ -341,33 +342,37 @@ def delete_synchrofacts(self, threshold, in_place=False, mode='delete'):
...
+                        try:
+                            idx = self._get_spiketrain_index(
+                                group.spiketrains,
+                                st)
+                        except ValueError:
+                            # st is not in this group, move to next group
+                            continue
+
+                        # st found in group, replace with new_st
+                        group.spiketrains[idx] = new_st
```
dizcza, line 374: this change is not necessary but OK
you've just increased the number of tabs --> increased (insignificantly) the code complexity

Kleinjohann: It is necessary because the `continue` was skipping the line below the block you commented on here which replaces the spiketrain in the input list.

elephant/spike_train_synchrony.py
Kleinjohann: I have added more information on what's happening and on how to fix this. The warning is getting quite long though. What do you think?

dizcza (approved):

elephant/spike_train_synchrony.py
dizcza: It's better even though it's longer.

Issue #12, opened 2014-08-06
Metadata for results


rproepp:
I think this warrants its own issue:

@toddrjen wrote in #11:

> This also leads me to another issue I have been thinking about: what do we do about the metadata of a neo object? When, for example, we get the average spike rate of a spike train, we end up with just a quantity. Is that what we want? Might it be a good idea to have some class that stores the output of these sorts of analyses along with the metadata of the original neo object? Or is that overkill?
> 
> The problem with this is doing it in a generic manner. You can't really use a SpikeTrain, since the resulting object may not meet the rules of a SpikeTrain. On the other hand, creating a generic "results" class would make it impossible to know what metadata you should expect from an object. And having a more specific SpikeTrainResults object would be difficult since it would need to be able to handle scalars, 1D arrays, and maybe even ND arrays depending on what analyses we allow. So it is a difficult problem, but I think having some way to keep the metadata bound to the results of some manipulation is important.

I think this verges into overkill territory :-) For most results (like the average rate of a spike train), the caller knows exactly from what object the result has been calculated. The caller also knows if and what metadata is needed, while our analysis function doesn't, so I would leave the responsibility upstream.

However, there might be analysis where this information is not available to the caller. For example, an analysis that takes a number of objects, but only uses some of them based on their content. I don't know if we will have such functions - I would try to avoid it but it might be necessary for some algorithms. In that case, I would return providence information to the caller: provide which objects have actually been used. By linking results to the actual objects used in their creation, all metadata is available and we do not need to create new result types with all the complications that come with that.

Moritz-Alexander-Kern:
For provenance tracking see Alpaca (Automated Lightweight Provenance Capture):

Documentation:
https://alpaca-prov.readthedocs.io/en/latest/

Code:
https://github.com/INM-6/alpaca

![](https://alpaca-prov.readthedocs.io/en/latest/_static/images/alpaca_logo.png)

PR #406, opened 2021-03-02
[WIP] Basic provenance tracking in Elephant

kohlerca:
This pull request introduces tools to capture provenance information when using Elephant, and saving it as metadata (#12).
Two elements are implemented as the module `elephant.buffalo`:

**1. A provenance tracker to be used as a function decorator (`Provenance` class in the `elephant.buffalo.provenance` module).**
 
Tracks each function call inside an analysis script, and stores a provenance track.

Some Elephant functions already have the decorator (e.g., functions in `statistics`, `spectral` or `spike_train_generation` modules). The user can also apply the decorator to the user-defined functions in the Python script whose provenance is being tracked.

For each decorated function, specific arguments in the initialization of the decorator object define how the actual function arguments are treated during tracking:
* inputs: list of the names of the arguments that are considered input to the function
* file_input: list of the names of the arguments that represent a file read by the function
* file_output: list of the names of the arguments that represent a file written by the function

The tracking is activated by calling the `activate()` interface function within the script. The code in the scope of this function call is then tracked.

Provenance data can be saved at the end by calling `save_provenance()` interface function, to serialize the history according to the W3C PROV data model (that can be subsequently queried and visualized).

**2. Data Analysis objects (`elephant.buffalo.objects` module)**

Defines a hierarchy of objects to represent the output of Elephant functions in a standardized way. Provenance information is embedded, and the classes are supposed to support data sharing and reuse (e.g., serialization to disk, together with the provenance information).

Currently, objects for histograms and power spectrum density estimates are implemented (affects `statistics.time_histogram` and `spectral.welch_psd`). A new function was implemented for computing PSTHs (`statistics.psth`). 

The objects are designed so that the output is compatible with legacy code (i.e., they extend the Python object that a particular function was returning previously).

The behavior of Elephant functions is controlled by the `elephant.buffalo.USE_ANALYSIS_OBJECT` flag. If `True`, the functions that were modified to support this new functionality will return the new data analysis objects. If `False`, the old object is returned. It is `False` by default.

This ensures that the objects can be used independently of the provenance tracking decorator.

**Examples**

Usage is demonstrated in several scripts in the `elephant.buffalo.examples` directory. A `README.md` file is provided.

**Dependencies**

This module introduces new package dependencies: `networkx`, `pyvis`, `prov`, `dill`, `joblib`.

**Specific points**

- [ ] `statistics.psth` and time histogram data model: this functionality is partly implemented in Viziphant, where the result of `statistics.time_histogram` can be plotted and a `neo.Event` can be used to add markers to the plot later. Some questions may be raised. One possibility is to consider the PSTH as a time histogram strictly around one event (as the current implementation in `PSTHObject`, with negative pre-event times in case the event is not at the start, or zero if the event is at the start of the histogram). An alternative would be to consider PSTH as a time histogram with one or more bound events (with or without resetting the time information with respect to any of the events, that could be optional and set in the object). In the current implementation, one specific time point is used, but this could be easily replaced by `neo.Event` with the definition of the label by the annotations, as the Viziphant `add_event` function. Histogram bin times are relative to this event time (i.e., the event is the zero in the time axis). Therefore, the current data model for `PSTHObject` assumes that the histogram is bound to an event, as the peristimulus in the name suggests. In the end, the class also allows reusing, as the `TimeHistogramObject` can become a `PSTHObject` by just setting the event.

- [ ] Strategy for unit tests (i.e, how to ensure that Elephant functionality is not broken when provenance is activated and to verify the tracked output). Ideally, all the Elephant tests must pass when provenance and/or analysis objects are active. Currently, this was done in `statistics` by copying the `test_*.py` and inserting provenance tracking for each test of the affected module (but this leads to duplication of the code). The PROV information can be serialized directly to a string. The resulting provenance data can then be easily checked.

- [ ] Documentation and final module structure (i.e., source files hierarchy, object and function names, and public/private functions/classes).

- [ ] Include other computations that can be done in a given analysis object as methods? For instance, one can compute the power in a specific frequency band, which would be the integration of the resulting PSD in the interval defined by the low and high frequencies of the band (a common task in the analysis, but that must be done by hand currently). Then we would have a method like `PSDObject.band_power(low, high)`, that would return the array/matrix (and this method would also have embedded provenance tracking as other Elephant functions). This way, we can quickly provide functionality that requires NumPy/SciPy functions in the workflow, like computing the mean of the `PSDObject.psd` array calling `np.mean` with `axis=0` in the case of multichannel data, that may ease the user experience while helping with provenance (avoiding to insert the decorator into NumPy functions or creating user-defined functions). Some functions could be added to Elephant (e.g., the integration for band power as `spectral.band_power()`). But for simpler computations, the user would probably prefer using straightforward functions like `np.mean`. 

Work in progress:

- [ ] PROV serialization
- [ ] Neo container hashes for faster performance
- [ ] Flow control information (for description and can also be used to organize the visualization in blocks)
- [ ] Visualization provided by Elephant (currently, a custom version of NetworkX DiGraph). Can also think about Sankey diagrams?


dizcza:
## General issues

The approach taken here requires creating a new class for each type of analysis results. This will populate Elephant with sugarcode, empty wrappers, and stubs. Each new functionality will require this as well. For example, 

```python
class PSDObject(AnalysisObject, namedtuple('PSDObject', 'frequencies psd')):
    """
    Class to store outputs of Elephant functions that compute power spectrum
    density (PSD) estimations (e.g.: `elephant.spectral.welch_psd`).

    It's a `namedtuple` with two values, frequencies and PSD, such that one can
    call ``frequencies, psd = PSDObject(...)``. All parameters listed below
    are accessible as attributes.

    Parameters
    ----------
    frequencies : (F,) array-like
        Vector with the frequencies for which the PSD was computed.
    psd : (N, F) array-like or (F,) array-like
        Value of the PSD for each frequency. If multichannel data, `N` is the
        number of channels and `F` the number of frequencies. Dimension `F`
        must be the same as `frequencies`.
    method : str
        Name of the method that was used to compute the PSD.
    params : dict, optional
        All the parameters that were used to compute the PSD.
        Default: None

    Raises
    ------
    ValueError
        If the dimension `F` of `frequencies` and `psd` differ.
    """

    def __new__(cls, frequencies, psd, method, params=None, *args, **kwargs):
        if psd.shape[-1] != frequencies.shape[0]:
            raise ValueError("PSD array must have the same length for the"
                             "frequency dimension")
        return super().__new__(cls, frequencies, psd)

    def __init__(self, frequencies, psd, method, params=None):
        super().__init__(self, frequencies, psd)
        self.method = method
        self.params = deepcopy(params)
```

takes 40 lines of code with a very small useful payload.

First of all, `welch_psd` should *always* return a named tuple - let's be nice to the users even if they don't want to turn the provenance on.
Secondly, `AnalisisObject` can be rewritten to be a general-purpose class that does not require inheritance each time a new analysis method is added leading to a new analysis result output.

```python
@Provenance
def func(array, **params):
    return namedtuple("MyNamedTuple", ('a1', 'a2'))  # example 1
    return any_other_thing   # example 2

def Provenance(func):
    @wraps
    def wrapped(*args, **kwargs):
        res = func(*args, **kwargs)
        if buffalo.is_active and not isinstance(res, AnalysisObject):
            res = AnalysisObject(res, params=kwargs)
        return res
    return wrapped
```

If needed (if there are valuable parameters computed inside the function that are not part of the input `params`), `res` can be initialized to an AnalysisObject directly in the function body, as Cristiano currently did in `welch_psd`. The rest of the simple functions that can be easily wrapped as in the `Provenance` function above don't need to mess up with the implementation function body, which is nice. Moreover, I vote for never changing Elephant functions code and always wrapping the result in an AnalysisObject at the cost of specifying less information about the object (only input-output parameters and relationship) but, at the same time, be generic that will surely work now and in the unforeseen future, which I cannot say otherwise if the current approach of subclassing is taken.

That was, actually, my initial idea about the provenance one year ago. **Have we explored this path?**

## Other issues

1. Don't edit the existing tutorials - statistics. Instead, create a separate tutorial that shows the features of provenance.
2. The version should be `0.0.1`, not `0.0.2`, since it's the first public release - FIXED.
3. As I understand, all elephant public functions will be decorated with a provenance decorator or?
4. PSDObject must implement abstract `serialize` and `load` methods.

Moritz-Alexander-Kern:
This project was moved to a different location, if needed this PR can be re-opened.

PR #411, opened 2021-03-16
Feature WPLI

ojoenlanuca:
To further increase the number of functions in the phase_analysis module of ‘elephant’, we implemented the ‘weighted_phase_lag_index()’-function to calculate the WPLI, which is a measure of phase-synchronization and based on the imaginary part of the complex-valued cross-spectrum of two signals. This measure was first introduced in ‘An improved index of phase-synchronization for electrophysiological data in the presence of volume-conduction, noise and sample-size bias’ by Vinck et al., 2011. Correctness was tested  among other things via ‘ground-truth’-testing with:  MATLABs package FieldTrip and its function ft_connectivity_wpli() and its wrapper ft_connectivity(); as well as with python package  MNE and its function spectral_connectivity().


ojoenlanuca:
Could you have a look at this PR and share your feedback/opinion about it? @mdenker , @rjurkus , @kohlerca

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+    -----
+    This implementation is base on the formula taken from [1] (pp.1550):
+    .. math::
+        WPLI = \frac{| E( |Im(X)| * sgn(Im(X)) ) |}{E( |Im(X)| )}
+
+    with:
+    E{...} : expected value operator
+    Im{X} : imaginary component of the cross-spectrum
+    X = Z_1 * Z_2_conjugate : as cross-spectrum
+    X = R * exp(i * theta) : R = magnitude and theta = relative phase
```
mdenker, line 347: I think this is somewhat redundant -- that's just saying that X is complex. 
```suggestion
```

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+    Z = A * Y
```
mdenker, line 348: I thought Z is the Fourier spectrum?

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+
+def weighted_phase_lag_index(signal_i, signal_j, sampling_frequency,
+                             absolute_value=True):
+    r"""
+    Calculates the Weigthed Phase-Lag Index (WPLI)
+
+    This function expects the two signals (each containing multiple trials).
+    It Fourier-transforms the signals and for each trial pair, it calculates
+    the cross-spectrum. With that the WPLI is calculated.
+
```
mdenker, line 320: This reads more like an algortihmic instruction on how to calculate the WPLI -- as a user, you would expect a more useful explanation of this measure. I think one should rephrase.

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+    # obtain cross-spectrum
+    cs = fft1 * np.conjugate(fft2)
+    # calculate WPLI
+    wpli_num = np.mean(np.abs(np.imag(cs)) * np.sign(np.imag(cs)), axis=0)
+    if absolute_value:
+        wpli_num = np.abs(wpli_num)
+    wpli_den = np.mean(np.abs(np.imag(cs)), axis=0)
+    wpli = wpli_num / wpli_den
+
+    return freqs, wpli
```
mdenker, line 386: As such, the function returns the most interesting thing, `wpli`, second, whereas the first return `freqs` is the more uninteresting thing. I would strongly suggest ot flip this. Otherwise often you have to type
```
,wpli = weighted_phase_lag()...
```

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+    with:
+    E{...} : expected value operator
+    Im{X} : imaginary component of the cross-spectrum
+    X = Z_1 * Z_2_conjugate : as cross-spectrum
+    X = R * exp(i * theta) : R = magnitude and theta = relative phase
+    Z = A * Y
+    where
+    A: real-valued matrix with frequency dependent coefficients
+    Y: complex-valued vector, representing the Fourier spectra
+    of a particular frequency
```
mdenker, line 352: This is so complicated, I don't get it...

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+
+    Usage/Interpretation of the parameter 'absolute_value':
+    When set to 'False', the WPLI contains additional directionality
+    information about which signal leads/lags the other signal:
+        - wpli > 0 : first signal leads second one
+        - wpli < 0 : first signal lags second one
```
mdenker, line 358: I think this should go right to the parameter description further up.

mdenker (dismissed):

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+    r"""
+    Calculates the Weigthed Phase-Lag Index (WPLI)
+
+    This function expects the two signals (each containing multiple trials).
+    It Fourier-transforms the signals and for each trial pair, it calculates
+    the cross-spectrum. With that the WPLI is calculated.
+
+    Parameters
+    ----------
+    signal_i, signal_j: (t, n) np.array
```
kohlerca, line 323: In the tests, it appears that Quantity is passed to the function. If the function supports multiple input types, then they must be listed here, and specific unit tests should be implemented for each type to check for any potential breaking changes when packages are updated. Moreover, the docstring should be formatted according to the conventions and style guide (e.g., like it was done with `phase_locking_value`).

elephant/test/test_phase_analysis.py
```diff
@@ -370,5 +373,266 @@ def testPhaseLockingValue_raise_Error_if_trial_lengths_are_different(self):
...
+        self.lfps2_r = dataset2_r['lfp_matrix'] * pq.uV
+        # ARRTIFICIAL LFP-DATASET
+        filename1_a = os.path.sep.join(['cross_testing_scripts',
+                                        'artificial_LFPs_1.mat'])
+        dataset1_a = scipy.io.loadmat(filename1_a, squeeze_me=True)
+        filename2_a = os.path.sep.join(['cross_testing_scripts',
+                                        'artificial_LFPs_2.mat'])
+        dataset2_a = scipy.io.loadmat(filename2_a, squeeze_me=True)
+        # get the relevant values
+        self.lfps1_a = dataset1_a['lfp_matrix'] * pq.uV
```
kohlerca, line 490: The naming of the variables are using different conventions, i.e., in some `_a` or `_r` are used for naming the artificial or real datasets, while later we have `ARTIFICIAL` or `REAL`. The overall code would be more understandable if using the same naming and using full words.

elephant/test/test_phase_analysis.py
```diff
@@ -370,5 +373,266 @@ def testPhaseLockingValue_raise_Error_if_trial_lengths_are_different(self):
...
+            self.wpli_ground_truth_FieldTrip_ARTIFICIAL,
+            atol=1e-14, rtol=1e-12, equal_nan=True)
+
+    def test_WPLI_comparison_to_multitaper_approaches(self):
+        """
+        Test if WPLI values are equal to those calculated from
+        FieldTrips' ft_connectivity() and MNEs' spectral_connectivity() at
+        frequencies [16, 36, 52, 70, 100]Hz.
+        """
+        freq, wpli, = elephant.phase_analysis.weighted_phase_lag_index(
```
kohlerca, line 555: The extra comma after `wpli` is not needed, as this is already a tuple.

elephant/phase_analysis.py
```diff
@@ -307,3 +307,80 @@ def phase_difference(alpha, beta):
...
+        Time-series of the first and second signals,
+        with `t` time points and `n` trials.
+    sampling_frequency: quantity/ scaler
+        Sampling frequency of the signals in Hz.
+    absolute_value: boolean (default: True)
+        Takes the absolute value of the numerator in the WPLI-formula.
+
+    Returns
```
kohlerca, line 331: It seems that the function returns `freqs` and `wpli`. Therefore, both must be described here.

elephant/test/test_phase_analysis.py
```diff
@@ -370,5 +373,266 @@ def testPhaseLockingValue_raise_Error_if_trial_lengths_are_different(self):
...
+                                        'artificial_LFPs_2.mat'])
+
+        savemat(filename1_a, mdic_1)
+        savemat(filename2_a, mdic_2)
+
+    def setUp(self):
+        self.tolerance = 1e-15
+
+        # simple samples of different shapes to assert ErrorRaising
+        self.simple_x = np.array([[0, -np.pi, np.pi], [0, -np.pi, np.pi]])
```
kohlerca, line 465: To understand the tests later, more descriptive names can be used, like `trials2_length3`, `trials1_length3`, `trials1_length4`). This makes it easier to understand the two unit tests later, and also makes sure that `x`, `y`, or `z` are not swapped and you are indeed emulating that particular error in the test. 

Or, in order to improve simplicity and readability, you could instead implement just a single test to check the error of wrong input dimensions. Therefore, these arrays would be defined inside that test function and the two assertions for either violation would be tested at once. These variables are defined at the class set up just for those two small tests at the end.

elephant/test/test_phase_analysis.py
```diff
@@ -370,5 +373,266 @@ def testPhaseLockingValue_raise_Error_if_trial_lengths_are_different(self):
             self.simple_y, self.simple_z)
 
 
+class WeightedPhaseLagIndexTestCase(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        np.random.seed(73)
+        cls._generate_datasets_for_ground_truth_testing()
```
kohlerca, line 380: I am not sure about this flow. The two `.mat` files saved by this function are used in the other scripts, to generate the CSVs to check the results against the other packages. But every time this test is run, new `.mat` files are generated, so the precomputed ground truth will not correspond. How is everything integrated into the testing pipeline? 

If that is the case, one alternative would be to have this function returning arrays, using it outside the test to generate the `.mat` for the scripts using FieldTrip and MNE (i.e., before running the scripts for MNE or Fieldtrip, call a script that computes the source `.mat` files used).
 
Then, when instantiating this test, call the function again and check if the arrays are the same as the ones stored in the `.mat` files that were used for the ground truth CSVs.

elephant/phase_analysis.py
ojoenlanuca: yes i agree.

elephant/phase_analysis.py
ojoenlanuca: Yes it is. This was another mathematical representation from the Vinck-paper. 
To be more explicit, i suggest this:
    X = Z_i * Z_j_conjugate : as cross-spectrum
    where
    Z_i,j: complex-valued vector, representing the Fourier spectra
    of a particular frequency of the signals i and j

elephant/phase_analysis.py
```diff
@@ -307,3 +308,92 @@ def phase_difference(alpha, beta):
...
+    References
+    ----------
+    ..  [1] Martin Vinck, Robert Oostenveld, Marijn van Wingerden,
+        Franscesco Battaglia, Cyriel M.A. Pennartz; "An improved index of
+        phase-synchronization for electrophysiological data in the presence
+        of volume-conduction, noise and sample-size bias"; NeuroImage, vol. 55,
+        pp. 1548-1565, 2011
+
+    """
+    if sampling_frequency is None:  # neo.AnalogSignal-input
```
kohlerca, line 374: It will be safer if checking the input types to be `neo.AnalogSignal` when None, otherwise this will raise an unhandled exception if the user passes a NumPy array and the rate is None. It should also raise an exception if not passing the sample rate when needed. Moreover, what if the sampling rates in `signal_i` and `signal_j` are different?

elephant/test/test_phase_analysis.py
```diff
@@ -370,5 +373,329 @@ def testPhaseLockingValue_raise_Error_if_trial_lengths_are_different(self):
...
+                wpli, self.wpli_ground_truth_FieldTrip_artificial,
+                atol=1e-14, rtol=1e-12, equal_nan=True)
+
+    def test_WPLI_comparison_to_multitaper_approaches(self):
+        """
+        Test if WPLI values are equal to those calculated from
+        FieldTrips' ft_connectivity() and MNEs' spectral_connectivity() at
+        frequencies [16, 36, 52, 70, 100]Hz.
+        """
+        for i in range(1, 4):
```
kohlerca, line 521: The only difference in each sub-test is the variables that are used as parameters for the WPLI function. You can then further improve and take advantage of `subTest`s by using a `dict` to define the parameters for the function and iterate over it. Therefore, you can get the same information as in the branching that you are using, but the code will be less redundant. Not only this reduce errors and improve readability, but also all tests will actually use the same code.

```
tests = {

	'Quantity': {
		'signal_i': self.lfps1_artificial,
		'signal_j': self.lfps2_artificial,
		'sampling_frequency': self.sf1_artificial
	},


	'np.ndarray': {
		'signal_i': self.lfps1_artificial.magnitude,
		'signal_j': self.lfps2_artificial.magnitude,
		'sampling_frequency': self.sf1_artificial
	},


	'neo.AnalogSignal' : {
		'signal_i': self.lfps1_artificial_AnalogSignal,
		'signal_j': self.lfps2_artificial_AnalogSignal
	}


}

for input_type, inputs in tests.items():
	# Compute WPLI using each input type
	freq, wpli = elephant.phase_analysis.weighted_phase_lag_index(**inputs, absolute_value=False)
 
        # mask for supposed wpli=1 frequencies,
        # freq=70Hz with supposed wpli=0 is treated separately,
        # because of noise existence in the artificial dataset
        mask = ((freq == 16) | (freq == 36) | (freq == 52) | (freq == 100))

        # comparing to FieldTrips' ft_conectivity()
        with self.subTest(msg=f"FieldTrip; wpli=1; {input_type} input"):
		...
	...
```

elephant/phase_analysis.py
ojoenlanuca: new commit 39785fe7cb60d2b5a829b5a5ae4f391ef0c8917a is regarding this now

elephant/test/test_phase_analysis.py
ojoenlanuca: commit 131f5abed9c010881d38b502c07875dce39d6b5a adopted this suggestion

ojoenlanuca (commented):
Connect commits to requested changes/suggestions.

mdenker (approved):

PR #412, opened 2021-03-24
[Enhancement] Set complexity dtypes for memory efficiency

morales-gregorio:
Hey!

@Kleinjohann and I have made minor changes to the Complexity class to use `np.uint16` type values in the complexity arrays, this should be able to represent complexities all the way to 65,535 which not even the largest recording systems can achieve right now. 

We also updated the tests to check that this dtype is kept all the way to the single spike train annotations, reducing the memory usage by a factor of 4. 

Let us know if we should change somethings

Best,
Aitor&Alex

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -1361,11 +1361,12 @@ class Complexity(object):
                           'Note that using the complexity epoch to get '
                           'precise spike times can lead to rounding errors.')
 
+        complexity = self.time_histogram.magnitude.flatten()
+        complexity = complexity.astype(np.uint16)
+
         epoch = neo.Epoch(left_edges,
                           durations=durations,
-                          array_annotations={
-                              'complexity':
-                              self.time_histogram.magnitude.flatten()})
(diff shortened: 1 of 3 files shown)


elephant/statistics.py
```diff
@@ -1429,6 +1430,8 @@ def _epoch_with_spread(self):
         # Note: all spike trains share the same t_start and t_stop.
         left_edges[0] = max(self.t_start, left_edges[0])
 
+        complexities = complexities.astype(np.uint16)
```
mdenker, line 1433: This seems redundant with line 1417. Can one not directly initialize complexities as `np.uint16`?

morales-gregorio: I do not remember exactly why included this line. I am guessing that after doing some operations in the array it turned the values to a more memory consuming type. @Kleinjohann do you remember if that was the case?

Kleinjohann: @mdenker is right, I changed it to immediately cast to `np.unit16`.

mdenker (approved):

Issue #408, opened 2021-03-11
[Feature] Move away from using deprecated nose for testing


sanjayankur31:
**Feature**
nose is deprecated, so it'll be good to move to another testing tool, perhaps py.test.

**Motivation**
nose is deprecated upstream (the last commit was in 2016.) and is also being deprecated in Fedora.

https://github.com/nose-devs/nose/commits/master

It is still in use here, though: in the CI script (and it's listed in the requirements file also).

https://github.com/NeuralEnsemble/elephant/blob/master/.travis.yml#L74

The website says:

https://nose.readthedocs.io/en/latest/

"Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2."

Deprecation of nose in Fedora: https://fedoraproject.org/wiki/Changes/DeprecateNose

**Alternatives**
py.test or nose2 (but nose2 doesn't seem to be at par with nose or py.test)

**Additional context**
On the Fedora build, I've used `py.test` to run the tests instead of the nosetests command, and that seems to work fine. However, I'm not well-versed enough with the code to know if there some nose-specific bits in there that will need to be migrated too.

mdenker:
Dear sanjayankur31,
thanks for your feedback, that's indeed a very good suggestion to take up.

PR #413, opened 2021-04-02
[Enhancement] Change test framework from Nose to pytest

kohlerca:
The unit tests are run using `pytest` ([https://pytest.org](https://pytest.org)) in replacement for the deprecated `nose`.

This PR addresses #408.

Travis CI settings were updated to use the new framework.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
         - conda install -c conda-forge pyopencl oclgrind clang=9.0.1
         - pip install -r requirements/requirements-extras.txt
         - pip install mpi4py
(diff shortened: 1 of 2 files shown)


mdenker (approved):

mdenker:
Seems all ok, and in fact, we do need pytest capabilities in fixing a bug currently in the travis testing: The package is installed using `python setup.py install`, however, the unit test actually do not use this installed version, but import from the working directory instead. This was not problemantic before, but with the inclusion of C code in an upcoming PR, this leads to problems. Therefore, this future PR will include a further option to pytest
```
pytest --cov=elephant --import-mode=importlib
```
which makes sure that the installed version of elephant is used, not local version.
This is just a heads-up, there is, I think,  no need to change this in the current PR.

Issue #409, opened 2021-03-11
[Bug] Failing test: elephant/test/test_conversion.py::BinnedSpikeTrainTestCase::test_repr


sanjayankur31:
**Describe the bug**
While building 0.10.0 on Fedora, we're seeing a test fail: `elephant/test/test_conversion.py::BinnedSpikeTrainTestCase::test_repr`

**To Reproduce**
1. Build elephant
2. Run `pytest` to run tests

We get this:

```
______________________ BinnedSpikeTrainTestCase.test_repr ______________________                                                                                             
                                                                                                                                                                             
self = <elephant.test.test_conversion.BinnedSpikeTrainTestCase testMethod=test_repr>                                                                                         
                                                                                                                                                                             
    def test_repr(self):                                                                                                                                                     
        train = neo.SpikeTrain(times=np.array([1.001, 1.002, 1.005]) * pq.s,                                                                                                 
                               t_start=1 * pq.s, t_stop=1.01 * pq.s)                                                                                                         
        bst = cv.BinnedSpikeTrain(train, t_start=1 * pq.s,                                                                                                                   
                                  t_stop=1.01 * pq.s,                                                                                                                        
                                  bin_size=1 * pq.ms)                                                                                                                        
>       self.assertEqual(repr(bst), "BinnedSpikeTrain(t_start=1.0 s, "                                                                                                       
                                    "t_stop=1.01 s, bin_size=0.001 s; "                                                                                                      
                                    "shape=(1, 10), format=csr_matrix)")                                                                                                     
E       AssertionError: 'Binn[19 chars]t=1.0, t_stop=1.01, bin_size=0.001; shape=(1, [19 chars]rix)' != 'Binn[19 chars]t=1.0 s, t_stop=1.01 s, bin_size=0.001 s; shap[25 char
s]rix)'                                                                                                                                                                      
E       - BinnedSpikeTrain(t_start=1.0, t_stop=1.01, bin_size=0.001; shape=(1, 10), format=csr_matrix)                                                                       
E       + BinnedSpikeTrain(t_start=1.0 s, t_stop=1.01 s, bin_size=0.001 s; shape=(1, 10), format=csr_matrix)
E       ?                             ++             ++                ++

elephant/test/test_conversion.py:684: AssertionError
```

**Expected behavior**
All tests should pass

**Environment**
 - OS (e.g., Linux): Fedora rawhide (development/35)
 - How you installed elephant (`conda`, `pip`, source): From source
[...]

PR #418, opened 2021-05-10
Changed __repr__ in BinnedSpikeTrain to support quantities<0.12.4

kohlerca:
The unit test for method `__repr__` in `conversion.BinnedSpikeTrain` fails if `quantities` is `0.12.3` or below.

The old method was using `Quantity` scalars inside an f-string. This required the `__format__` method implemented in `quantities==0.12.4` to generate the correct string representation of `BinnedSpikeTrain` with magnitudes and units.

This PR changes the string generation by casting the `Quantity` scalars to string, which uses the `__str__` method in `Quantity` and displays the magnitudes and units as expected. This does not rely on any formatting method and thereby it is compatible with older versions of `quantities`.

This addresses issue #409.

### DIFF
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -405,8 +405,8 @@ class BinnedSpikeTrain(object):
         return self.n_bins
 
     def __repr__(self):
-        return f"{type(self).__name__}(t_start={self.t_start}, " \
-               f"t_stop={self.t_stop}, bin_size={self.bin_size}; " \
+        return f"{type(self).__name__}(t_start={str(self.t_start)}, " \
+               f"t_stop={str(self.t_stop)}, bin_size={str(self.bin_size)}; " \
                f"shape={self.shape}, " \
                f"format={self.sparse_matrix.__class__.__name__})"
 


ojoenlanuca (approved):
I can approve these changes.
Little side-note: it is maybe IDE dependent, but when i run test_conversion.test_repr in Pycharm to recreate the bug #409, the self.assertEqual() function might expect as first parameter the desired value and secondly the actual value. Because in the respective debug-message, the order is switched to
Expected: string generated by repr()
Actual: hardcoded reference string

kohlerca:
This is a PyCharm issue, as it is known that its diff tool provides these `expected/actual` reports in an order different than the one used in several of our tests (https://youtrack.jetbrains.com/issue/PY-26471), i.e., assuming that the expected value is the first. There is a recommendation by Guido van Rossum against forcing this, and instead, refer to the values in a generic form, such as `first/second` (https://mail.python.org/pipermail/python-dev/2010-December/106954.html). This is actually what the assert function documentation uses.  

If running the tests using the command line, the standard `!=` and the points with the differences are shown without any semantics attached:

```
E       AssertionError: 'Binn[19 chars]t=1.0, t_stop=1.01, bin_size=0.001; shape=(1, [19 chars]rix)' != 'Binn[19 chars]t=1.0 s, t_stop=1.01 s, bin_size=0.001 s; shap[25 chars]rix)'
E       - BinnedSpikeTrain(t_start=1.0, t_stop=1.01, bin_size=0.001; shape=(1, 10), format=csr_matrix)
E       + BinnedSpikeTrain(t_start=1.0 s, t_stop=1.01 s, bin_size=0.001 s; shape=(1, 10), format=csr_matrix)
E       ?                             ++             ++                ++
```

ojoenlanuca:
Ok, so then i can approve the pull-request.

ojoenlanuca (approved):
These changes prevent test_converstion.test_repr()-failure for quantities version <= 0.12.3 and should be adopted.

mdenker (approved):

PR #419, opened 2021-05-14
Optimized SPADE analysis

mdenker:
This PR includes a new version of the FP-growth algorithm for the SPADE analysis based on the work of Florian Porrmann (https://github.com/fporrmann/FPG). The new library features a number of SPADE specific optimizations that improve run-time significantly.

In order to include the C++ module, the install procedure was altered to compile the corresponding `fim.so` from scratch, rather than downloading a precompiled module post-install. Also, recipes for building pip wheels for Linux and Windows were included as github actions, building in `cibuildwheels`. Currently, these are configured to run on every push, and after a testing period, this may be altered to build only after a PR or even only after a version-tagged PR.

Also, a change from `nosetest` to `pytest` was required in order to have available the `--import-lib` option that makes sure the compiled version of the library is used for unit testing. This includes most changes of #413 .

@stellalessandra There is one unit test in line 275 of `test_spade.py` that you had commented out, and which fails. Does this need fixing?


stellalessandra:
@mdenker I commented out the test as a filter for the maximum number of occurrences is not implemented in Florian's fpgrowth. It is a filtering not to mine patterns with more occurrences of the one fixed. I think we can just remove it from SPADE, as it is an unintuitive parameter given the purposes of the method. I don't see why one should decide to mine patterns with fewer occurrences that the ones that can be found...

mdenker:
@stellalessandra Ah, thanks for the clarification. Then I suggest to leave the unit test commented for now, and discuss removing this parameter in a new issue/PR, in order not to blow up this PR.

elephant/test/test_spade.py
```diff
@@ -698,25 +701,25 @@ def test_signature_significance_fdr_bh_corr(self):
...
+    #                 'joint_isi_dithering': [[2, 2, 0.8]],
+    #                 'bin_shuffling': [[2, 2, 1.0], [2, 3, 0.2]],
+    #                 'dither_spikes_with_refractory_period':
+    #                     [[2, 2, 0.8]]}
+    #     for surr_method in surr_methods:
+    #         pv_spec = spade.pvalue_spectrum(
+    #             spiketrains, bin_size=self.bin_size,
+    #             winlen=self.winlen, dither=15*pq.ms,
+    #             n_surr=5, surr_method=surr_method)
+    #         self.assertEqual(pv_spec, pv_specs[surr_method])
```
Kleinjohann, line 722: What's up with this test?

Kleinjohann (commented):
I found a few typos and I'm not sure about a second test that's also commented out. I was able to verify locally that the build works on Ubuntu using [act](https://github.com/nektos/act), but Windows is not supported there, how did you test that @mdenker ?

elephant/test/test_spade.py
mdenker: Not sure -- it does pass for me, also appear to be functional. @stellalessandra Do you know why this one is commented out?

mdenker: Ok, I see the point, I'd agree this would be somewhat redundant, and may break in case of changes to the implementations of the surrogate methods, even if those are correct (i.e.: maybe after an update of the algorithm, a surrogate method produces a different realization with a given random seed -- in that case this test would start to fail, but there would be no bug.

PR #432, opened 2021-09-08
BUG-Fix #424: t_start information was lost while transposing lfp

### LINKED ISSUE #424: [Bug] estimate csd loses t_start information for iCSD methods
**Description**
`estimate_csd` always returns an `AnalogSignal` object with `t_start=0` for iCSD methods. 

**To Reproduce**
```
>>> import numpy as np
>>> import quantities as pq
>>> from neo import AnalogSignal
>>> from elephant.current_source_density import estimate_csd
>>> lfp = AnalogSignal(np.random.rand(10, 2)*pq.V, sampling_rate=1*pq.Hz, t_start=3.5*pq.s)
>>> csd = estimate_csd(lfp, np.array([[1],[2]])*pq.um, method='StandardCSD')
>>> csd.t_start
array(0.) * s
>>> csd_kern = estimate_csd(lfp, np.array([[1],[2]])*pq.um, method='KCSD1D')
>>> csd_kern.t_start
array(3.5) * s
```

**Proposed fix**
The root seems to be in l.188 of `current_source_density.py`:
```
lfp = neo.AnalogSignal(np.asarray(lfp).T, units=lfp.units, sampling_rate=lfp.sampling_rate)
```
which, does not use `t_start` from the input (kernel methods do not have the issue because they do not redefine the input lfp).
Adding `t_start=lfp.t_start` in l.188 should fix the problem. Alternatively, one could perhaps replace the redefinition with `lfp=lfp.T` since that seems to be all it does but I am not sure I am correct in that thought.

ojoenlanuca:
This bug appeared in line 188 of 'current_source_density.py',  because t_start was not set in the new neo.AnalogSignal object, which should be the transpose of the neo.AnalogSignal object 'lfp'. 
Therefore  the default value for t_start = 0s was used. 
It was fixed by transposing the neo.AnalogSignal 'lfp' directly with lfp = lfp.T, which preserves all set attributes of the neo.AnalogSignal.

### DIFF
--- a/elephant/current_source_density.py
+++ b/elephant/current_source_density.py
@@ -185,8 +185,7 @@ def estimate_csd(lfp, coordinates=None, method=None,
                 raise ValueError("The order of {} filter must be \
                                   specified".format(kwargs['f_type']))
 
-        lfp = neo.AnalogSignal(np.asarray(lfp).T, units=lfp.units,
-                               sampling_rate=lfp.sampling_rate)
+        lfp = lfp.T
         csd_method = getattr(icsd, method)  # fetch class from icsd.py file
         csd_estimator = csd_method(lfp=lfp.magnitude * lfp.units,
                                    coord_electrode=coordinates.flatten(),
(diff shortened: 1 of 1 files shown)


rgutzen (approved):

Moritz-Alexander-Kern (approved):

Moritz-Alexander-Kern (commented):
good modification, @rgutzen  @ojoenlanuca

Issue #384, opened 2020-11-18
[Bug] zscore(..., inplace=True) does not change the units of the original signal to dimensionless


jpgill86:
If `inplace=True`, the returned signal has dimensionless units, but the modified in-place signal has the original units.

rgutzen:
The problem seems to arise in this line:
```
sig_normalized = sig.magnitude.astype(mean.dtype, copy=not inplace)
```
which is just a link to the asig object when `inplace` is True. However, the removal of the units is only done on the newly created asig object `sig_dimless`.
So, just adding the line
```
sig_normalized.units = pq.dimensionles
```
would fix the behavior.

dizcza:
At the first glance, I also thought about this. But we cannot change the units of an analog signal:
```
ValueError: can not modify units of a view of a Quantity
```

jpgill86:
This seems to be a way to forcibly overwrite a Quantity's units, but I haven't checked whether it breaks anything:
```python
sig_normalized._dimensionality = pq.dimensionless.dimensionality
```

dizcza:
Let's ask @apdavison what he thinks about modifying the units of an analog signal in-place.

dizcza:
@samuelgarcia maybe Samuel has an opinion?

PR #440, opened 2021-12-12
Updated zscore for in-place operations

kohlerca:
The `zscore` function has a parameter `inplace`, to allow storing the z-scored values in the original `AnalogSignal` object(s).
The function always returns the `AnalogSignal` objects, whether or not an in-place operation is performed.

However, the in-place operation only operated at the level of the data array, i.e., whenever the NumPy array from the original `AnalogSignal` was modified with the z-scored values, a new `AnalogSignal` object was created pointing to that NumPy array. Therefore, this has never been a true in-place operation, as the returned objects differed from the inputs.

This PR fixes that by returning the original `AnalogSignal` when `inplace` is True, and a new object (also using a different data array) when `inplace` is False.

Moreover, the units in the original `AnalogSignal` object were not modified when `inplace` is True and only the returned object was dimensionless (see #384). Therefore, the original `AnalogSignal` ended up with z-scored values with a unit such as `uV`, which is misleading. The code in the PR now sets the unit to `dimensionless` in the original `AnalogSignal`.

Finally, the docstring mentioned that the in-place operation was not attempted if the `AnalogSignal` data type did not allow it (i.e., whenever the type was not float). The PR code now checks, for every object in the input list, whether the type allows the in-place operation. If `inplace` is True and this is not allowed, the parameter is ignored for that `AnalogSignal` (i.e., a copy will be returned along with a new `AnalogSignal` object), and a warning to the user is shown. Alternatively, this could be done before the main loop, checking if there is at least one object in the input list that does not allow the operation, and then setting `inplace` to False in that case (i.e., preventing any in-place operation at all).

To test the behavior of the updated function:

```
import numpy as np
import quantities as pq
import neo
from elephant.signal_processing import zscore

for inplace, dtype in ((True, np.int64), (False, np.int64),
                       (True, np.float64), (False, np.float64)):

    print("=======================")
    print(f"Inplace: {inplace}; dtype: {dtype}")

    test = neo.AnalogSignal(np.random.normal(10, 2, (60000, 96)),
                   units='uV', dtype=dtype, sampling_rate=30000*pq.Hz)

    zscored = zscore(test, inplace=inplace)

    print("IDs", id(test), id(zscored))
    print("Same object?", test is zscored)

    print(test[:1, :5])
    print(zscored[:1, :5])
    print("Equal data?", np.array_equal(test, zscored))
```
which produces

```
=======================
Inplace: True; dtype: <class 'numpy.int64'>
/home/koehler/PycharmProjects/elephant_fix/elephant/signal_processing.py:172: UserWarning: Could not perform inplace operation as the original signal dtype is not float. Source: None
  warnings.warn("Could not perform inplace operation as the "
IDs 139982366671472 139982321011504
Same object? False
[[11 10  8  8 10]] uV
[[ 0.73385822  0.25246159 -0.74701266 -0.74680096  0.24532251]] dimensionless
Equal data? False
=======================
Inplace: False; dtype: <class 'numpy.int64'>
IDs 139982321011840 139982321011728
Same object? False
[[13 12  8  5 10]] uV
[[ 1.74665356  1.23204397 -0.73947224 -2.22643084  0.24853166]] dimensionless
Equal data? False
=======================
Inplace: True; dtype: <class 'numpy.float64'>
IDs 139982321011616 139982321011616
Same object? True
[[ 0.4987234  -0.25468458 -1.32321378  0.41974894  0.32817535]] dimensionless
[[ 0.4987234  -0.25468458 -1.32321378  0.41974894  0.32817535]] dimensionless
Equal data? True
=======================
Inplace: False; dtype: <class 'numpy.float64'>
IDs 139982333189472 139982321011840
Same object? False
[[11.18127483  8.5491104  11.8618302  13.54534356 11.8022081 ]] uV
[[ 0.58965413 -0.73046345  0.93888403  1.77151628  0.89908117]] dimensionless
Equal data? False
```

### DIFF
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -68,8 +68,7 @@ def zscore(signal, inplace=True):
     inplace : bool, optional
         If True, the contents of the input `signal` is replaced by the
         z-transformed signal, if possible, i.e when the signal type is float.
-        If the signal type is not float, the parameter is ignored and a
-        warning is raised whenever trying the in-place operation.
+        If the signal type is not float, an error is raised.
         If False, a copy of the original `signal` is returned.
         Default: True
 
@@ -80,10 +79,10 @@ def zscore(signal, inplace=True):
         `neo.AnalogSignal`, a corresponding `neo.AnalogSignal` is returned,
         containing the z-transformed signal with dimensionless unit.
 
-    Warns
-    -----
-    UserWarning
-        If `inplace` is True and the signal dtype is not float.
+    Raises
+    ------
+    ValueError
+        If `inplace` is True and the type of `signal` is not float.
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


kohlerca:
As discussed, the function was modified to raise an error when the in-place operation is not valid. Also, the duplicate is made using the Neo function.

All unit tests were modified to check if the objects are the same/different according to the `inplace` parameter. Moreover, the unit test using an `AnalogSignal` of type int as input with `inplace=True` was modified to check for the exception.

Moritz-Alexander-Kern:
after modification to raise an error, check behavior with added exception handling:
```python
import numpy as np
import quantities as pq
import neo
from elephant.signal_processing import zscore

for inplace, dtype in ((True, np.int64), (False, np.int64),
                       (True, np.float64), (False, np.float64)):

    print("=======================")
    print(f"Inplace: {inplace}; dtype: {dtype}")

    test = neo.AnalogSignal(np.random.normal(10, 2, (60000, 96)),
                   units='uV', dtype=dtype, sampling_rate=30000*pq.Hz)
    try:
        zscored = zscore(test, inplace=inplace)

        print("IDs", id(test), id(zscored))
        print("Same object?", test is zscored)

        print(test[:1, :5])
        print(zscored[:1, :5])
        print("Equal data?", np.array_equal(test, zscored))
    except:
        print("error was raised")
```

which produces the same output as in the first comment by @kohlerca, except now an error is raised with `Inplace: True; dtype: <class 'numpy.int64'>`, as expected.

```
=======================
Inplace: True; dtype: <class 'numpy.int64'>
error was raised
=======================
Inplace: False; dtype: <class 'numpy.int64'>
IDs 139652275362384 139652275359920
Same object? False
[[ 8  6  9  7 12]] uV
[[-0.74279302 -1.73908209 -0.24362649 -1.23294208  1.24192959]] dimensionless
Equal data? False
=======================
Inplace: True; dtype: <class 'numpy.float64'>
IDs 139652275363168 139652275363168
Same object? True
[[ 1.30684587  0.06441688 -1.34509361 -0.47104841  0.33951849]] dimensionless
[[ 1.30684587  0.06441688 -1.34509361 -0.47104841  0.33951849]] dimensionless
Equal data? True
=======================
Inplace: False; dtype: <class 'numpy.float64'>
IDs 139652275362160 139652275362384
Same object? False
[[ 6.62071703  8.79765984 11.69532766 14.2552753  13.66768817]] uV
[[-1.68625204 -0.59526308  0.8431924   2.12242394  1.83686467]] dimensionless
Equal data? False
```

Moritz-Alexander-Kern (approved):

Issue #442, opened 2021-12-24
Elephant GPFA not running


Sary789:
**Questions and Help**

Hello, I am trying to run the tutorial on GPFA on my jupyter notebook, but it seems I cannot import GPFA because when I type "from elephant.gpfa import GPFA" I keep receiving the following error: "cannot import name 'float_factorial' from 'scipy._lib._util' (../opt/anaconda3/lib/python3.7/site-packages/scipy/_lib/_util.py). Does anyone know how I can solve this? Any suggestions will be greatly appreciated!

Moritz-Alexander-Kern:
Hello Sary789,
thank you for reporting this.

Looking at the error message, this seems to be related to scipy.
Unfortunately I wasn't able to reproduce this behavior with the following setup:

**Environment**

-  OS: Linux
-   Python version: 3.8.10
-   neo  version: 0.10.0
-   elephant  version: 0.10.0
-  scipy version: 1.6.3

**Expected behavior**
The following error: 
_"cannot import name 'float_factorial' from 'scipy._lib._util' (../opt/anaconda3/lib/python3.7/site-packages/scipy/_lib/_util.py)_


Further information on the environment you are running could be helpful:
Which version of elephant and scipy are you using? (elephant requires scipy<1.7.0)

Moritz-Alexander-Kern:
We are closing this now, but feel free to contact if the problem persists.

PR #465, opened 2022-03-23
fix: extended error message on import errors, elephant[extras] missing

Moritz-Alexander-Kern:
This PR is partly a response to Issue #442 .

When installing elephant without extras requirements, e.g. `pip install elephant`, imports that rely on extra requirements fail. 
The corresponding `__init__.py` files contain helpful comments, but no information is given to the user. 

This PR extends the error messages displayed upon `ImportError` and adds a note for modules, which require elephant extras.


Moritz-Alexander-Kern:
We agreed, that this solution is to messy. Closed for now.

Issue #464, opened 2022-03-21
discretise_spiketimes behavior and documentation


kohlerca:
Hi everyone,

I noticed that the new `discretise_spiketimes` function (PR #454) replaces the generated `SpikeTrain` annotations `dict` with a reference to the `dict` in the original `SpikeTrain`, after the duplication with new data (line 1310). This will make any changes to the annotations in the original `SpikeTrain` to also change the annotations in the generated `SpikeTrain`, and vice-versa.

As this behavior is not clear, the code could make a brief explanation of the motivation as comments, and this could also be mentioned in the documentation so that it is apparent to the user. 

Alternatively, the line could be omitted from the code, as the Neo function already copies the annotations, and the benefit of having a reference to the original annotations is not clear.

Finally, the docstring of the function (or the module) would benefit from having a quick example showing what discretization does, similar to what is done for `BinnedSpikeTrain` in the module docstring.

Kleinjohann:
Hi @kohlerca, thanks for noticing this! The referenced annotations are a bug, that was not intended.

Should the usage example go into the docstring of the module or into the docstring of the function itself?

Moritz-Alexander-Kern:
Hi,
thanks for taking care of this @kohlerca and @Kleinjohann .

I would prefer to have the usage example in **the docstring of the function itself.** 
This will avoid clutter and overloaded docstrings for modules containing many functions in the future. Further I assume this is the place one would expect a usage example.

PR #468, opened 2022-04-01
added example to docstring, removed reference to annotations

Moritz-Alexander-Kern:
This PR addresses Issue #464.

Changes:
- added example to docstring
- removed reference to annotations

### DIFF
--- a/elephant/conversion.py
+++ b/elephant/conversion.py
@@ -1267,6 +1267,26 @@ def discretise_spiketimes(spiketrains, sampling_rate):
     -------
     neo.SpikeTrain or list of neo.SpikeTrain
         The discretised spiketrain(s)
+
+    Examples
+    --------
+    >>> import neo
+    >>> import numpy as np
+    >>> import quantities as pq
+    >>> from elephant import conversion
+    >>>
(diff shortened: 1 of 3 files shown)


kohlerca (approved):
The changes look good.

elephant/conversion.py
```diff
@@ -1267,6 +1267,26 @@ def discretise_spiketimes(spiketrains, sampling_rate):
...
         The discretised spiketrain(s)
+
+    Examples
+    --------
+    >>> import neo
+    >>> import numpy as np
+    >>> import quantities as pq
+    >>> from elephant import conversion
+    >>>
+    >>> times = (np.arange(10) + np.random.uniform(size=10)) * pq.ms
```
Kleinjohann, line 1278: Without a fixed seed this will fail when being run as a doctest since the result will change every time.
What's our policy for these examples? Should they work as doctests? 
I guess that would help to keep the examples up to date, when running this I noticed that a few other examples in this file currently fail as well.

Moritz-Alexander-Kern: fixed, see https://github.com/NeuralEnsemble/elephant/pull/468/commits/232dcb0558b171a083e88a88dd0639b68d5a30ed

Issue #469, opened 2022-04-07
pip install fails on macOS 12.3


heplesser:
**Describe the bug**
Pip-installing Elephant fails with a compiler error under macOS 12.3

**To Reproduce**
1.  Create conda environment from https://github.com/nest/nest-simulator/blob/master/environment.yml
2. Activate environment
3. `pip install quantities neo`
4. `pip install elephant`

This leads to a compiler failure, see attached log [elephant_pip_error.txt](https://github.com/NeuralEnsemble/elephant/files/8442475/elephant_pip_error.txt)

**Expected behavior**
Pip installs elephant.

**Environment**
 - OS (e.g., Linux): macOS 12.3.1
 - Python version: 3.9.12
 - `neo` python package version: 0.10.0

mdenker:
Thanks, @heplesser, indeed the mac build seems problematic due to compiler incompatibilities, We will likely deprecate the building of this module for the current elephant release, such that only the Python-fallback is available for now. The installation breaking for mac is an unintentional bug,  @Moritz-Alexander-Kern I suggest to disable mac installs for now in `setup.py` and consider setting up a new CI environment for darwin in github actions to ensure this from now surfacing again.

PR #472, opened 2022-04-12
disable mac installs for now in setup.py

Moritz-Alexander-Kern:
This PR addresses Issue #469 . 
Thanks @heplesser for reporting this.

**Bug:**
 Pip-installing Elephant fails with a compiler error under macOS 12.3.

**Fix:**
Disable mac installs for now in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,6 +84,6 @@ setup_kwargs = {
 }
 # do not compile external modules on darwin
 if platform.system() in ["Windows", "Linux"]:
-    setup_kwargs["ext_modules"] = fim_module
+    setup_kwargs["ext_modules"] = [fim_module]
 
 setup(**setup_kwargs)


setup.py
```diff
@@ -53,23 +53,23 @@
...
-    license="BSD",
-    url='http://python-elephant.org',
-    python_requires=">=3.7",
-    classifiers=[
+setup_kwargs = {
+    "name": "elephant",
+    "version": version,
+    "packages": ['elephant', 'elephant.test'],
+    "include_package_data": True,
+    "ext_modules": [fim_module],
```
mdenker, line 61: Remove here...

setup.py
```diff
@@ -82,4 +82,9 @@
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3 :: Only',
         'Topic :: Scientific/Engineering']
-)
+}
+# do not compile external modules on darwin
+if platform.system() == "Darwin":
```
mdenker, line 87: ...and add the extension module here for `Windows` and `Linux` platforms.

mdenker (changes requested):
Better than removing for mac builds, I suggest to only add the fim module for `Windows` and `Linux` platforms.

Also, in the if-then-else clause on line 43, I suggest to specifically list `elif platform.system()=="Linux":`

mdenker (approved):

PR #474, opened 2022-04-13
Fix/asset on mac os, add CI workflow for macOS

Moritz-Alexander-Kern:
This PR is a response to #469 .

**Changes:**
- added 2 GitHub actions runners with macOS 10.15 and macOS-latest
- changed precision in asset to double


setup.py
```diff
@@ -4,6 +4,7 @@
 
 from setuptools import setup, Extension
 
+print(platform.system())
```
mdenker, line 7:
```suggestion
```

mdenker (approved):

PR #483, opened 2022-04-25
Fixed MANIFEST.in to include CUDA/OpenCL sources when installing

kohlerca:
Currently, when installing Elephant, the CUDA/OpenCL sources in `elephant/asset` are not copied. This leads to failure when trying to run ASSET using GPU acceleration.

This PR fixes the manifest file so that all required files are copied.


Moritz-Alexander-Kern:
Thanks for pointing this out @kohlerca 

I checked elephant release 0.10.0, here the relevant `*cl` and `*cu`are still included in the build package. 
In 0.11.0 and 0.11.1, they are missing, this PR  fixes this.

Moritz-Alexander-Kern (approved):

PR #497, opened 2022-06-23
Fix/doctests

Moritz-Alexander-Kern:
### doctests
The doctest module looks for such sequences of prompts in a docstring, re-executes the extracted command and checks the output against the output of the command given in the docstrings test example. 

Many doctrings in elephant contain code examples using the following syntax:

- When using the Python shell, the primary prompt: >>> , is followed by new commands. 
- The secondary prompt: ... , is used when continuing commands on multiple lines; and the result of executing the command is expected on following lines. 
- A blank line, or another line starting with the primary prompt is seen as the end of the output from the command.

So far elephant has no standardized form for doctests, which means, that some examples e.g. contain figures or don't produce any result. In those cases running doctests executes those  examples and checks for errors, but does not check the results.

With this PR errors occurring during doctests are fixed and further steps aim to make doctests part of the Cx workflow.

This PR could be the starting point to establish doctests as a standard in elephant.

### run _only_ doctests 
`pytest elephant --doctest-modules --ignore=elephant/test/`

todo:
- [x] create CI workflow to execute doctests, done, see #503 
- [x] split this PR into multiple parts, part 1: fixes that only require reformatting, done, see #502


Moritz-Alexander-Kern:
This Pr was split in 2 parts and is done

Issue #455, opened 2022-02-21
Error in instantaneous rate when passing a 1-dimensional kernel sigma [Bug] 


Kleinjohann:
**Describe the bug**
When passing a more than 0-dimensional `sigma` in the generation of a kernel and using said kernel to calculate the instantaneous firing rate an error is raised somewhere in `fftconvolve`.

**To Reproduce**
```
import neo
import elephant
import quantities as pq

st = neo.SpikeTrain([1]*pq.s, t_stop=2*pq.s)
kernel = elephant.kernels.GaussianKernel(sigma=[20]*pq.ms)

rate = elephant.statistics.instantaneous_rate(st,
                                              sampling_period=1*pq.ms,
                                              kernel=kernel)
```
returns
```
Traceback (most recent call last):
  File "reproduce_sigma_error.py", line 9, in <module>
    rate = elephant.statistics.instantaneous_rate(st,
  File "/home/kleinjohann/software/elephant/elephant/utils.py", line 79, in wrapper
    return func(*args, **kwargs)
  File "/home/kleinjohann/software/elephant/elephant/statistics.py", line 868, in instantaneous_rate
    rate = scipy.signal.fftconvolve(time_vectors,
  File "/home/kleinjohann/software/miniconda3/envs/elephant/lib/python3.8/site-packages/scipy/signal/signaltools.py", line 525, in fftconvolve
    raise ValueError("in1 and in2 should have the same dimensionality")
ValueError: in1 and in2 should have the same dimensionality
```
I tracked this down to https://github.com/NeuralEnsemble/elephant/blob/d0d39206a5a2bff0d57d365a63a30e402dbc597d/elephant/statistics.py#L852
where `t_arr` gains an extra dimension since `cutoff_sigma` keeps the extra dimension of `sigma`.

**Expected behavior**
Either a more helpful error should be raised which tells me to fix the dimensionality of `sigma`, or different dimensionalities should be accounted for internally and no error should be raised.

**Environment**
 - OS: Linux
 - How I installed elephant: from source
 - Python version: 3.8.1
 - `neo` python package version: 0.9.0
 - `elephant` version: d0d3920
 - `scipy` version: 1.5.4

Moritz-Alexander-Kern:
Thanks for reporting this. 
I get the same error. 

We will look into this, currently `elephant.statistics.instantaneous_rate` is undergoing changes with PR #458 in relation to #374.

PR #499, opened 2022-06-28
Fix/onedimensional sigma

Moritz-Alexander-Kern:
This PR addresses Issue #455 .

### Changes
- check dimensions of sigma, if dimension > 0 , raise `typeError`
- handle unambiguous case `ndim=1` , `size = 1` with `sigma = sigma[0]`
- added tests for both above cases

### Test
The following code snippet should no longer throw an error

```python
import neo
import elephant
import quantities as pq

st = neo.SpikeTrain([1]*pq.s, t_stop=2*pq.s)
kernel = elephant.kernels.GaussianKernel(sigma=[20]*pq.ms)

rate = elephant.statistics.instantaneous_rate(st,
                                              sampling_period=1*pq.ms,
                                              kernel=kernel)
```

### DIFF
--- a/elephant/kernels.py
+++ b/elephant/kernels.py
@@ -147,8 +147,8 @@ class Kernel(object):
             raise TypeError("'sigma' must be a quantity")
 
         if sigma.ndim > 0:
-            # handle the one-dimensional case
-            if sigma.size == 1:
+            # handle the one-dimensional case of size 1
+            if sigma.ndim == 1 & sigma.size == 1:
                 sigma = sigma[0]
             else:
                 raise TypeError("'sigma' cannot be multidimensional")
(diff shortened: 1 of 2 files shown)


elephant/kernels.py
```diff
@@ -142,6 +146,13 @@ def __init__(self, sigma, invert=False):
         if not isinstance(sigma, pq.Quantity):
             raise TypeError("'sigma' must be a quantity")
 
+        if sigma.ndim > 0:
+            # handle the one-dimensional case of size 1
+            if sigma.ndim == 1 & sigma.size == 1:
```
Kleinjohann, line 151: Just a side remark: 
This will also catch something like `sigma = [[5]] * pq.s`
Seems fine to me though, since a meaningful error is raised in that case now.

Kleinjohann (approved):
Looks good, thanks for the fix, I can confirm that my use case that originally caused this error is covered by this.

mdenker (approved):
I assume that in the long run -- besides this not being Python -- we may want to have a more standard way to check types, and in particular the case of 0, 1 and 2 dimensional `quantities` arrays.

Issue #410, opened 2021-03-11
[Bug] Test failure on armv7hl: UETestCase.test__UE_surrogate


sanjayankur31:
**Describe the bug**
While updating to 0.10.0 in Fedora, on the 32 bit builders (armv7hl), this test fails:

**To Reproduce**
1.  run tests on 32 bit (armv7hl) system
2.

This is the error we see:

```
________________________ UETestCase.test__UE_surrogate_________________________
self = <elephant.test.test_unitary_event_analysis.UETestCase testMethod=test__UE_surrogate>
    def test__UE_surrogate(self):
        mat = self.binary_sts
        pattern_hash = np.array([4])
        _, rate_avg_surr, _, n_emp_surr, indices_surr =\
>           ue._UE(
                mat,
                pattern_hash,
                method='surrogate_TrialByTrial',
                n_surrogates=100)
elephant/test/test_unitary_event_analysis.py:305: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
elephant/unitary_event_analysis.py:687: in _UE
    pval = dist_exp(n_emp)
elephant/unitary_event_analysis.py:585: in pval
    hist = np.bincount(np.int64(n_exp))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
args = (array([1, 2, 1, 1, 1, 1, 0, 0, 2, 2, 0, 2, 1, 1, 1, 0, 0, 2, 0, 1, 1, 0,
       1, 1, 1, 1, 1, 1, 1, 0, 2, 2, 1, 1, 2... 1, 2, 0, 1, 0, 0, 0, 0, 2, 1, 1, 2, 1, 2, 1, 2, 0, 0, 1, 2,
       2, 1, 1, 2, 1, 2, 0, 0, 0, 0, 2, 1], dtype=int64),)
kwargs = {}
relevant_args = (array([1, 2, 1, 1, 1, 1, 0, 0, 2, 2, 0, 2, 1, 1, 1, 0, 0, 2, 0, 1, 1, 0,
[...]

PR #500, opened 2022-06-28
fix/#410, UE cast to int in 'unsafe' mode

Moritz-Alexander-Kern:
This PR addresses Issue #410  

The following error occurred in https://github.com/NeuralEnsemble/elephant/blob/0df45812b93229b3186c061cb84ec29fd380e09c/elephant/unitary_event_analysis.py#L585

```
TypeError: Cannot cast array data from dtype('int64') to dtype('int32') according to the rule 'safe'
```

This might be fixed by using np.astype(int) instead of np.in64() to cast to int. 
`np.astype` allows 'unsafe' casting by default, see:
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html

### DIFF
--- a/elephant/unitary_event_analysis.py
+++ b/elephant/unitary_event_analysis.py
@@ -582,7 +582,7 @@ def gen_pval_anal(mat, pattern_hash, method='analytic_TrialByTrial',
             mat, pattern_hash, method=method, n_surrogates=n_surrogates)
 
         def pval(n_emp):
-            hist = np.bincount(np.int64(n_exp))
+            hist = np.bincount(n_exp.astype(int))
             exp_dist = hist / float(np.sum(hist))
             if len(n_emp) > 1:
                 raise ValueError('In surrogate method the p_value can be'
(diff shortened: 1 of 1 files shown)


Moritz-Alexander-Kern:
The numpy documentation suggest the following for casting:

_To convert the type of an array, use the .astype() method (preferred) or the type itself as a function. For example:_

```python
z.astype(float)                 
array([0.,  1.,  2.])

np.int8(z)
array([0, 1, 2], dtype=int8)
```

See: https://numpy.org/doc/stable/user/basics.types.html

I suggest to follow numpys preferred method, but I don't have a strong opinion on this.

Moritz-Alexander-Kern:
Hey @sanjayankur31 ,

This should fix in PR #410 , please let me know if this solves the issue.

sanjayankur31:
Thanks, it's on my list, just not managed to get to it yet. Will hopefully find some time in the next couple of days.

sanjayankur31:
Seems to fix the issue on arm, all tests pass with this set of changes:

https://koji.fedoraproject.org/koji/taskinfo?taskID=91845461

Fedora's also dropped armv7hl support from the next release (F37), so we shouldn't have armv7hl related issues in the future:

https://www.fedoraproject.org/wiki/Changes/RetireARMv7

Moritz-Alexander-Kern:
Thanks @sanjayankur31 for having a look. 

Build report looks good: https://koji.fedoraproject.org/koji/taskinfo?taskID=91845461
arm runner

PR #509, opened 2022-07-27
macOS 10.15 deprecated, update github actions

Moritz-Alexander-Kern:
github actions macOS 10.15 runner is deprecated, updated to macOS 11 and 12.

see:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,7 +73,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -124,7 +124,7 @@ jobs:
       fail-fast: false
       matrix:
         # OS [ubuntu-latest, macos-latest, windows-latest]
-        os: [macos-10.15,macos-latest]
```
Moritz-Alexander-Kern, line 127: latest refers to macOS 11

PR #535, opened 2023-01-10
Fix/numpy >= 1.24.0

Moritz-Alexander-Kern:
## Deprecations
With release of numpy v1.24 several deprecations expired, see: 
https://numpy.org/devdocs/release/1.24.0-notes.html#deprecations

The following are relevant for this PR:
- Ragged array creation will now always raise a `ValueError` unless `dtype=object` is passed. This includes nested sequences.
- As of numpy v1.20, `numpy.float` as well as similar aliases (including `numpy.int`) were deprecated, with numpy 1.24 those have been completely removed

## Fix
- replaced the numpy aliases for builtin types with the buitlin types
- rewritten UE unit tests (replaced `np.vstack()`)
- In accordance with NEP 29 + 1 year support for numpy 1.18 was dropped.


Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

PR #536, opened 2023-01-10
Fix/remove python 3.7

Moritz-Alexander-Kern:
In accordance with NEP 29 support for Python 3.7 and numpy 1.20 is dropped. 
(https://numpy.org/neps/nep-0029-deprecation_policy.html)


Moritz-Alexander-Kern (commented):
build wheels

PR #540, opened 2023-02-17
Fix/changes in documentation

Moritz-Alexander-Kern:
This PR changes:
- change heading from "Spike-triggered LFP phase" to "Phase Analysis"
- removed statistics tutorial from function reference


Moritz-Alexander-Kern (commented):
build wheels

PR #554, opened 2023-03-23
Fix complexity for binary=False and spread=0

Kleinjohann:
I noticed that the complexity class actually does not work reliably for `spread=0` and `binary=False`, because the length of the complexity histogram is hard-coded to the number of input spiketrains, which limits the maximum complexity that can be detected. Firstly, this is inconsistent with the `spread>0` case in which the histogram always has the minimal length which accommodates the highest entry. Secondly, this may lead to errors, because for `binary=False` the complexity can take arbitrarily high values, so the required size of the histogram cannot be known in advance.

I set the length of the histogram to the minimal value which accommodates the maximal complexity, made the documentation of the `binary` parameter clearer and added tests for `binary=False`.

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -1152,8 +1152,9 @@ class Complexity(object):
 
         Default: None
     binary : bool, optional
-        *  If True then the time histograms will be binary.
-        *  If False the total number of synchronous spikes is counted in the
+        *  If True then the time histograms will only count the number of
+           neurons which spike in each bin.
+        *  If False the total number of spikes per bin is counted in the
            time histogram.
 
         Default: True
@@ -1360,10 +1361,13 @@ class Complexity(object):
                                    self.bin_size,
                                    binary=self.binary)
 
+        time_hist_magnitude = time_hist.magnitude.flatten()
+        max_hist_entry = max(time_hist_magnitude)
+
         # Computing the histogram of the entries of pophist
         complexity_hist = np.histogram(
-            time_hist.magnitude,
-            bins=range(0, len(self.input_spiketrains) + 2))[0]
+            time_hist_magnitude,
+            bins=range(0, max_hist_entry + 2))[0]
 
         return time_hist, complexity_hist
 
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -1362,12 +1362,9 @@ class Complexity(object):
                                    binary=self.binary)
 
         time_hist_magnitude = time_hist.magnitude.flatten()
-        max_hist_entry = max(time_hist_magnitude)
 
         # Computing the histogram of the entries of pophist
-        complexity_hist = np.histogram(
-            time_hist_magnitude,
-            bins=range(0, max_hist_entry + 2))[0]
+        complexity_hist = np.bincount(time_hist_magnitude)
 
         return time_hist, complexity_hist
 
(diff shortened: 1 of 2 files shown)


elephant/statistics.py
```diff
@@ -1360,10 +1361,13 @@ def _histogram_no_spread(self):
...
 
+        time_hist_magnitude = time_hist.magnitude.flatten()
+        max_hist_entry = max(time_hist_magnitude)
+
         # Computing the histogram of the entries of pophist
         complexity_hist = np.histogram(
-            time_hist.magnitude,
-            bins=range(0, len(self.input_spiketrains) + 2))[0]
+            time_hist_magnitude,
+            bins=range(0, max_hist_entry + 2))[0]
```
Moritz-Alexander-Kern, line 1370:
```suggestion
```

see: possible simplification

elephant/statistics.py
```diff
@@ -1360,10 +1361,13 @@ def _histogram_no_spread(self):
                                    self.bin_size,
                                    binary=self.binary)
 
+        time_hist_magnitude = time_hist.magnitude.flatten()
+        max_hist_entry = max(time_hist_magnitude)
+
         # Computing the histogram of the entries of pophist
         complexity_hist = np.histogram(
```
Moritz-Alexander-Kern, line 1368: **[optional] possible simplification ?**
Could this also be achieved by using np.bincount?, see: https://numpy.org/doc/stable/reference/generated/numpy.bincount.html


```suggestion
        complexity_hist = np.bincount(time_hist_magnitude)
```

If I understand correctly in the following example:
```python
time_histogram= np.array([0, 2, 0, 0, 1, 1, 0, 0, 1, 1,
                          0, 1, 1, 0, 0, 0, 2, 0, 1, 1])
```
we count the number of zeros, ones, and twos, so the desired result is:
```python
correct_complexity_histogram = np.array([10, 8, 2])
```

elephant/statistics.py
```diff
@@ -1360,10 +1361,13 @@ def _histogram_no_spread(self):
...
-            time_hist.magnitude,
-            bins=range(0, len(self.input_spiketrains) + 2))[0]
+            time_hist_magnitude,
```
Moritz-Alexander-Kern, line 1369:
```suggestion
```
see: possible simplification

elephant/statistics.py
```diff
@@ -1360,10 +1361,13 @@ def _histogram_no_spread(self):
                                    self.bin_size,
                                    binary=self.binary)
 
+        time_hist_magnitude = time_hist.magnitude.flatten()
+        max_hist_entry = max(time_hist_magnitude)
```
Moritz-Alexander-Kern, line 1365:
```suggestion
```

see: possible simplification

Moritz-Alexander-Kern (approved):
Nice fix with regression tests addressing the described issue.

Found one possible (?) simplification, feel free to ignore.

elephant/statistics.py
mdenker: After some experimenting, I think this simplification is in order, also for the border case of a zero-filled array

mdenker (approved):
This PR seems ok for me, but I would consider implementing the suggestion by @Moritz-Alexander-Kern -- looks for for me, I think both statements produce identical code. What do you think @Kleinjohann ?

Kleinjohann:
Thanks for your reviews! Good suggestion for the simplification, I tested it and it works fine!

PR #561, opened 2023-05-05
Update codemeta.json to fix download URL

apdavison:


Moritz-Alexander-Kern:
Hey Andrew, thanks for catching this. :1st_place_medal:

PR #568, opened 2023-05-30
[Fix] deprecation warnings numpy 1.25.0

Moritz-Alexander-Kern:
## Deprecation
1. The default behavior of `numpy.linalg.lstsq` will be changed, to fix this, set a parameter for `rcond`
  - set to `rcond = None` in granger
2. Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation.
 - fixed in gpfa, icsd, phase analysis 
3. changed import from `numpy.testing.utils` to `numpy.testing`
4. Comment out  example in doc string of icsd unittest, pytest is trying to run this as a docstest
5. replaced np.matrix with ndarray in KCSD

### DIFF
--- a/elephant/causality/granger.py
+++ b/elephant/causality/granger.py
@@ -284,8 +284,9 @@ def _vector_arm(signals, dimension, order):
     positive_lag_covariances = np.reshape(lag_covariances[1:],
                                           (dimension*order, dimension))
 
-    lstsq_coeffs = \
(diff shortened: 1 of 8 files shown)


Moritz-Alexander-Kern:
Thanks @rjurkus for your help approval on point 1, `rcond = None` .

Moritz-Alexander-Kern (commented):
build wheels

mdenker (approved):

Issue #574, opened 2023-06-27
[Bug] SPADE param min_neu does not work


pbouss:
**Describe the bug**
The parameter min_neu in SPADE should fix the minimal number of neurons in a pattern. This is typically used to avoid auto-patterns. In older versions of SPADE this was properly running but with the latest version. I get clearly patterns, that I don't want to get as output.  
My suspection is that the generated `fim.so` is not working properly. Especially, it doesn't work the same way as the locally stored `fim.so` that we were using before. 
@Kleinjohann @mdenker : do you have any idea where this problem might come from? 

**To Reproduce**
```
import numpy as np
from elephant.spade import spade
from elephant.spike_train_generation import StationaryPoissonProcess
import quantities as pq

np.random.seed(0)
poisson_process = StationaryPoissonProcess(rate=20*pq.Hz, t_stop=5*pq.s)

n_spiketrains = 3
spiketrains = poisson_process.generate_n_spiketrains(
    n_spiketrains=n_spiketrains)
bin_size = 5 * pq.ms

spade_output = spade(
    spiketrains,
    bin_size=bin_size,

    # +1 to pattern length to be able to find it again
    winlen=2,
    min_spikes=n_spiketrains,
    max_spikes=n_spiketrains,

    # This should avoid to get auto-patterns.
    # But apparently it doesn't work.
    min_neu=n_spiketrains,
    spectrum='3d#')

patterns = spade_output['patterns']
print(patterns)

for pattern in patterns:
    # Check that all spiketrains contain three different neurons
    # assert len(np.unique(pattern['neurons']))==n_spiketrains
    if len(np.unique(pattern['neurons']))!=n_spiketrains:
        print(f'Found autopattern with neurons {pattern["neurons"]}')
```

**Expected behavior**
The expected output of the code snippet is:
```
[{'itemset': (2, 5, 1), 'windows_ids': (36, 134, 856), 'neurons': [1, 2, 0], 'lags': array([5., 5.]) * ms, 'times': array([ 180.,  670., 4280.]) * ms, 'signature': (3, 3, 1), 'pvalue': -1}]
```

The actual outcome now is:
```
[...]

PR #575, opened 2023-06-28
Fixed bug in spade, linked to missing call of min_neu, added test

pbouss:
Resolving issue: https://github.com/NeuralEnsemble/elephant/issues/574

### DIFF
--- a/elephant/spade.py
+++ b/elephant/spade.py
@@ -110,11 +110,13 @@ from collections import defaultdict
 from functools import reduce
 from itertools import chain, combinations
 
+import numpy as np
+from scipy import sparse
+
+import quantities as pq
+
 import neo
 from neo.core.spiketrainlist import SpikeTrainList
-import numpy as np
-import quantities as pq
-from scipy import sparse
 
 import elephant.conversion as conv
 import elephant.spike_train_surrogates as surr
@@ -355,7 +357,7 @@ def spade(spiketrains, bin_size, winlen, min_spikes=2, min_occ=2,
             min_occ=min_occ, max_spikes=max_spikes, max_occ=max_occ,
             min_neu=min_neu, report='a')
         time_mining = time.time() - time_mining
-        print("Time for data mining: {}".format(time_mining))
+        print(f"Time for data mining: {time_mining}")
 
     # Decide if compute the approximated stability
     if compute_stability:
@@ -368,7 +370,7 @@ def spade(spiketrains, bin_size, winlen, min_spikes=2, min_occ=2,
         concepts = approximate_stability(
             concepts, rel_matrix, **approx_stab_pars)
         time_stability = time.time() - time_stability
-        print("Time for stability computation: {}".format(time_stability))
+        print(f"Time for stability computation: {time_stability}")
         # Filtering the concepts using stability thresholds
         if stability_thresh is not None:
             concepts = [concept for concept in concepts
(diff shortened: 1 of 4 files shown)


elephant/spade.py
```diff
@@ -912,6 +913,7 @@ def _fpgrowth(transactions, min_c=2, min_z=2, max_z=None,
                 report='a',
                 algo='s',
                 winlen=winlen,
+                min_neu=min_neu,
```
Moritz-Alexander-Kern, line 916: This line appears to be the relevant change if I understand correctly. It seems to address the core issue and provide a solution. Thanks!

Please correct me if I'm wrong about this.

elephant/test/test_spade.py
```diff
@@ -348,6 +348,41 @@ def test_fpgrowth_fca(self):
...
+        bin_size = 5 * pq.ms
+
+        spade_output = spade.spade(
+            spiketrains,
+            bin_size=bin_size,
+
+            # +1 to pattern length to be able to find it again
+            winlen=2,
+            min_spikes=n_spiketrains,
+            max_spikes=n_spiketrains,
```
Moritz-Alexander-Kern, line 370: Is the usage of `n_spiketrains` in this context specifically due to its value being 3? Or does the number of spiketrains have a direct impact on `min_spikes` and `max_spikes`?

elephant/test/test_spade.py
```diff
@@ -348,6 +348,41 @@ def test_fpgrowth_fca(self):
...
+            spectrum='3d#')
+
+        patterns = spade_output['patterns']
+        assert len(patterns) == 1
+        pattern = patterns[0]
+        assert len(np.unique(pattern['neurons'])) == n_spiketrains
+        assert_array_equal(pattern['itemset'], [2, 5, 1])
+        assert_array_equal(pattern['windows_ids'], [36, 134, 856])
+        assert_array_equal(pattern['neurons'], [1, 2, 0])
+        assert_array_equal(pattern['signature'], [3, 3, 1])
```
Moritz-Alexander-Kern, line 384: I'm curious to know why exactly this pattern is expected or considered the correct one. Are there any previous results or a rationale behind this choice? 
I noticed that the spiketrains were generated using the `StationaryPoissonProcess` class, so perhaps there is an obvious reason that I'm not aware of. 
I'd appreciate any insights or explanations to further educate myself on this matter.

Moritz-Alexander-Kern (commented):
Thanks for taking care of this bug and the code cleanup @pbouss .
**Cleanup**
- replaced old string formatting with f-strings
- sorted imports
- removed `_fpgrowth_filter` function (was already commented out)

**Bugfix**
- see Line [916](https://github.com/NeuralEnsemble/elephant/pull/575/files#diff-f086abec9c91a6cf36d2b8009c4d662000c81736514adc5e9758f1e99411d6dbR916): hand over `min_neu` as argument to `fim.fpgrowth` - function.
- implemented unit-test to check for expected pattern

In conclusion the issue was not with the `fim.so` but with the way arguments were passed to it? Please correct me if I get this wrong.

pbouss:
The old `fim.so`, i.e. the Borgelt version didn't have the argument. Therefore, there was this extra cleaning function that was outcommented when including the Porrman FIM implementation, which works correctly. - It was just forgotten to pass the argument.

elephant/test/test_spade.py
pbouss: I chose arbitrarily the 3 here. I just wanted to have a case, where I can have an unwanted autopatterns, that is with a repeating neuron. - This I get with this Poisson spike trains and choice of parameters.

elephant/test/test_spade.py
pbouss: There is no specific reason. Originally I only had this line: 
```
for pattern in patterns:
    # Check that all spiketrains contain three different neurons
    assert len(np.unique(pattern['neurons']))==n_spiketrains
```

Then, I thought for later code changes it might help to fix here a specific reproducible outcome. But there is no specific reason that the pattern looks like this. It's only the outcome given the random seed.

Moritz-Alexander-Kern (commented):
build wheels

PR #533, opened 2022-12-02
Fix/cad validation matlab

Moritz-Alexander-Kern:
This is a follow up PR to #576 , adding a validation test.
# Bug
This was discovered in issue #443 .

Summary: CAD (cell assembly detection) output is not the same as the original MATLAB version provided by the article's author. 

The MATLAB implementation by Eleonora Russo can be found here: 
https://github.com/DurstewitzLab/Cell-Assembly-Detection . 
This implementation was used in: 
Eleonora Russo and Daniel Durstewitz. Cell assemblies at multiple time scales with arbitrary lag constellations. Elife, 6:e19428, 2017.

## MATLAB
The MATLAB implementation defines `alph` as the input parameter of the function: 
[Line 1 in FindAssemblies_recursive.m](https://github.com/DurstewitzLab/Cell-Assembly-Detection/blob/master/Programs_and_data/FindAssemblies_recursive.m#L1)
```matlab
function [assembly_output]=FindAssemblies_recursive(binM,maxlag,alph,gg,Dc,No_th,O_th,bytelimit,reference_lag)
```

In later steps, the original input `alph` is  kept and derived values  are stored as a separate variable  `alpha` .
[Line 30 in FindAssemblies_recursive.m](https://github.com/DurstewitzLab/Cell-Assembly-Detection/blob/465109df4226c6b2996398baabeddd3f69c61844/Programs_and_data/FindAssemblies_recursive.m#L30)
```matlab
alpha=alph*2/(nu*(nu-1)*(2*maxlag+1));      %% bonferroni correction 
```
[Line 106 in FindAssemblies_recursive.m](https://github.com/DurstewitzLab/Cell-Assembly-Detection/blob/465109df4226c6b2996398baabeddd3f69c61844/Programs_and_data/FindAssemblies_recursive.m#L106)
```matlab
alpha=alph/(length(w2_to_test)*n_as*(2*maxlag+1));  %% bonferroni correction only for the test that I actually perform
```

## Elephant
The Elephant implementation defines `alpha` as the input parameter of the function: 
https://github.com/NeuralEnsemble/elephant/blob/b9d0d54e8560074fc34ec333cf93d8ab17d34b47/elephant/cell_assembly_detection.py#L90-L95

In subsequent steps, the original input `alpha` is changed:
https://github.com/NeuralEnsemble/elephant/blob/b9d0d54e8560074fc34ec333cf93d8ab17d34b47/elephant/cell_assembly_detection.py#L255

[...]

Issue #443, opened 2022-01-15
[Bug] CAD output is not the same as original MATLAB Version


Peyman7:
**Describe the bug**
CAD cell assembly detection method output is not the same as the original MATLAB version provided by the article's author. I have checked the source codes, and found the parameter alpha in the "    # testing for higher order assemblies" part gives the wrong value as it should take the original alpha=0.05 in the right side. 
alpha = alpha / float(len(w2_to_test) * n_as * (2 * max_lag + 1))
As alpha in previous lines recalled and its value is changed because of same alpha values in both side of equation, it should be changed in a way that the value of alpha in the right side always is be the same is the alpha input parameter.

**To Reproduce**
As alpha parameter name in input parameter may be changed to alph (as well as the same MATLAB version) and alpha parameter in the right side of alpha = _alph_/ float(len(w2_to_test) * n_as * (2 * max_lag + 1)) changes to alph.

**Environment**
 - OS: Windows
 - How you installed elephant (`pip):
 - Python version: 3.9.6
 - `elephant` python package version: 0.10.0 

**Environment**
 - OS (e.g., Linux):
 - How you installed elephant (`conda`, `pip`, source):
 - Python version:
 - `neo` python package version: 
 - `elephant` python package version:
 - (Any additional python package you want to include here)

Moritz-Alexander-Kern:
you've pointed to line:
https://github.com/NeuralEnsemble/elephant/blob/6ed3cb3089783035885e2fbec1fff535df7369cd/elephant/cell_assembly_detection.py#L363

another change to alpha seems to occur here:
https://github.com/NeuralEnsemble/elephant/blob/6ed3cb3089783035885e2fbec1fff535df7369cd/elephant/cell_assembly_detection.py#L255

I suggest creating a validation test based on the original MATLAB implementation as a regression test.
Is this  the MATLAB-implementation provided by the author you are using:  https://github.com/DurstewitzLab/Cell-Assembly-Detection ?

Peyman7:
Hi, 

That’s a good idea! 
Regarding the original MATLAB code link, you have pointed the correct link.

Cheers,
Peyman

Moritz-Alexander-Kern:
Fixed with #576

PR #576, opened 2023-07-12
[Fix] cell assembly detection

Moritz-Alexander-Kern:
## Issue
This PR is to address Issue #443 .

## Validation test (not part of this PR)
A second PR adding a validation test against the original MATLAB implementation by E. Russo is open here: #533 . 

## Changes
- improve code-style (PEP8)
- have the initial value of `alpha` under a separate name `alph`, this was also done in the original MTALB implementation.


Moritz-Alexander-Kern (commented):
build wheels

PR #578, opened 2023-07-12
[MAIN] Drop support for numpy 1.22 and Python 3.8 following NEP 29

Moritz-Alexander-Kern:
In accordance with NEP 29 + 1 year support for Python 3.8 and numpy 1.22 is dropped. 
(https://numpy.org/neps/nep-0029-deprecation_policy.html)


Moritz-Alexander-Kern:
this PR is one year ahead of schedule.

PR #585, opened 2023-07-25
[ENH] Added ASSET class initialization parameter to define the binning rounding error tolerance

kohlerca:
Currently, the `ASSET` class performs the binning of the spiketrains using the default behavior of the `BinnedSpikeTrain` class in `conversion`. 

Therefore, when spike times are closer to the right bin edge than a defined tolerance value, the `BinnedSpikeTrain` class shifts the spike to the next bin. This behavior can be controlled by the `tolerance` parameter of `BinnedSpikeTrain`.

This PR adds the new parameter `bin_tolerance` to the `ASSET` class, in order to allow the user to define the desired tolerance (and hence `BinnedSpikeTrain` behavior).

The parameter can take a string `'default'` to use the default behavior of `BinnedSpikeTrain` (and which corresponds to the current status of `ASSET` before this PR), or a float value or None. In the latter two cases, these values are forwarded to the binning class, and will result in either shifting or not shifting the spikes, respectively.

### DIFF
--- a/elephant/asset/asset.py
+++ b/elephant/asset/asset.py
@@ -158,7 +158,6 @@ except ImportError:
     size = 1
     rank = 0
 
-
 __all__ = [
     "ASSET",
     "synchronous_events_intersection",
@@ -528,7 +527,7 @@ def _stretched_metric_2d(x, y, stretch, ref_angle, working_memory=None,
             chunk_size = D_chunk.shape[0]
 
             assert (chunk_size == estimated_chunk or
-                    chunk_size == last_chunk)           # Safety check
+                    chunk_size == last_chunk)  # Safety check
 
             dX = x_array[:, start: start + chunk_size].T - x_array
             dY = y_array[:, start: start + chunk_size].T - y_array
@@ -605,6 +604,7 @@ class _GPUBackend:
        Python objects, PyOpenCL and PyCUDA clean up and free allocated memory
        automatically when garbage collection is executed.
     """
+
     def __init__(self, max_chunk_size=None):
         self.max_chunk_size = max_chunk_size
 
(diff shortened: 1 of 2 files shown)


elephant/asset/asset.py
```diff
@@ -1964,6 +1964,12 @@ class ASSET(object):
         If None, the attribute `t_stop` of the spike trains is used
         (if the same for all spike trains).
         Default: None
+    bin_tolerance : float or 'default' or None, optional
+        Defines the tolerance value for rounding errors when binning the
+        spike trains. If 'default', the value is the default as defined in
+        :class:`conv.BinnedSpikeTrain`. If None, no correction for binning
+        errors is performed. If a number, the binning will consider this value.
```
Moritz-Alexander-Kern, line 1971:
```suggestion
        :class:`~.conversion.BinnedSpikeTrain`. If None, no correction for
        binning errors is performed. If a number, the binning will consider
        this value.
```

Get working link to `BinnedSpikeTrain` on readthedocs.

elephant/asset/asset.py
```diff
@@ -1977,11 +1983,15 @@ class ASSET(object):
 
           fully disjoint.
 
+    See Also
+    --------
+    conv.BinnedSpikeTrain
```
Moritz-Alexander-Kern, line 1988:
```suggestion
    :class:`elephant.conversion.BinnedSpikeTrain`
```

Get working link to `BinnedSpikeTrain` on readthedocs.

Moritz-Alexander-Kern (changes requested):
Thanks for this nice enhancement of ASSET @kohlerca :tada:

elephant/asset/asset.py
kohlerca: Fixed in commit [8a090e8](https://github.com/NeuralEnsemble/elephant/pull/585/commits/8a090e88466b547206ae495e10c6e6f51dbe0dfe)

elephant/asset/asset.py
kohlerca: Fixed in commit [8a090e8](https://github.com/NeuralEnsemble/elephant/pull/585/commits/8a090e88466b547206ae495e10c6e6f51dbe0dfe)

kohlerca:
Thanks for the review, @Moritz-Alexander-Kern. I made the requested changes and also modified the unit tests slightly so that the binning of each of the two input lists is checked (same spiketrains, but reverse order).

Moritz-Alexander-Kern (approved):
Thanks implementing the requested changes so promptly @kohlerca . 
Approved from my side. :heavy_check_mark:

mdenker (approved):
Looks all good to me, unit tests are also understandable.

PR #588, opened 2023-08-01
[FIX] add skip test_multitaper_cohere_perfect_cohere for numpy == 1.25.2

Moritz-Alexander-Kern:
Added `numpy == 1.25.2` to the following code snippet, since the related numpy issue was not fixed with the latest release.

https://github.com/NeuralEnsemble/elephant/blob/2bd871aec145d897031aed327a7a4af0102c47cb/elephant/test/test_spectral.py#L957-L960


Moritz-Alexander-Kern (commented):
build wheels

PR #592, opened 2023-08-30
[Fix] z score inplace

### LINKED ISSUE #591: [Bug] signal_processing.zscore: Cannot perform inplace operation for np.float 32 and np.float64
**Describe the bug**
This originally occurred in the [Elephant_Tutorial_-_LFP_analysis.ipynb](https://drive.ebrains.eu/f/848183b51a5247c38994/), when running the notebook with the latest versions of the dependencies.

**To Reproduce**
Run the above mentioned notebook or the following code snippet:

```python
import elephant
from neo.core import AnalogSignal
import quantities as pq
import numpy as np

signal = np.array([1.2, 2, 3, 4.56, 5, 6], dtype=np.float32) # same for np.float64
# create a neo.AnalogSignal
analog_signal = AnalogSignal(signal, units='V', sampling_rate=1*pq.Hz)

elephant.signal_processing.zscore(analog_signal, inplace=True)
```

which yields:
```
  in <module>
    elephant.signal_processing.zscore(analog_signal, inplace=True)
  elephant/elephant/signal_processing.py", line 171, in zscore
    raise ValueError(f"Cannot perform inplace operation as the "
ValueError: Cannot perform inplace operation as the signal dtype is not float. Source: None
```

**Expected behavior**
no error for np.float32 and np.float64

**Environment**
| Tool        | Version |
|-------------|---------|
| Python | 3.10.12 |
| elephant    | 0.13.0  |
| neo         | 0.12.0  |
| pip         | 23.2.1  |
| quantities  | 0.14.1  |
| scipy       | 1.11.0  |
| numpy       | 1.25.1  |

**Possible fix (?)**

In Lines:

https://github.com/NeuralEnsemble/elephant/blob/2bd871aec145d897031aed327a7a4af0102c47cb/elephant/signal_processing.py#L170-L172

the `numpy.issubdtype()` function is used to determine whether the first argument is a typecode lower/equal in type hierarchy from second argument. If the first argument is lower or equal it returns true, otherwise it returns false.

Changing this to:

```python
np.issubdtype(sig.dtype, np.floating): 
```
could be a possible fix here?

**Update 28. August 2023:**
Discussed in Elephant meeting.

Moritz-Alexander-Kern:
This PR closes #591 .

## Changes:

The check for the signal type when using z_score inplace was changed from:
```python
 if inplace and not np.issubdtype(float, sig.dtype): 
```

to:
```python 
if inplace and not np.issubdtype(sig.dtype, np.floating):
```

Additionally a regression unittest was added.

This was originally discovered in the `Elephant_Tutorial_-_LFP_analysis.ipynb` notebook, see also https://tutorials.python-elephant.org .

### DIFF
--- a/elephant/signal_processing.py
+++ b/elephant/signal_processing.py
@@ -167,7 +167,7 @@ def zscore(signal, inplace=True):
     for sig in signal:
         # Perform inplace operation only if array is of dtype float.
         # Otherwise, raise an error.
-        if inplace and not np.issubdtype(float, sig.dtype):
+        if inplace and not np.issubdtype(sig.dtype, np.floating):
             raise ValueError(f"Cannot perform inplace operation as the "
                              f"signal dtype is not float. Source: {sig.name}")
 
--- a/elephant/test/test_signal_processing.py
+++ b/elephant/test/test_signal_processing.py
@@ -415,7 +415,7 @@ class ZscoreTestCase(unittest.TestCase):
         """
         Regression test: Inplace operations for z_score failed when using
         np.float32 or np.float64 types.
-        See Issue #591
+        See Issue #591.
         https://github.com/NeuralEnsemble/elephant/issues/591
         """
         test_types = (np.float32, np.float64)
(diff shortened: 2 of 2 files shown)


Moritz-Alexander-Kern (commented):
build wheels

kohlerca (approved):

PR #595, opened 2023-09-18
[Fix] skip multitaper unittest for new numpy version 1.26.0

Moritz-Alexander-Kern:
This PR is in Line with e.g. PR #588 .

Skipping unittest for multitaper perfect cohere as the underlying issue was not resolved with release of numpy 1.26.0.


Moritz-Alexander-Kern (commented):
build wheels

PR #494, opened 2022-06-10
Fix/add option: do not compile fim c-extension

Moritz-Alexander-Kern:
This PR adds an extra to elephant, which allows installation of Elephant without compiling the c++ extension `fim.cpp` used in SPADE.

Do not compile any extension:
- development  `pip install -e . --install-option='--no-compile' `
- install `pip install elephant --install-option='--no-compile' `

Do not compile spade extension:
- development  `pip install -e . --install-option='--no-compile-spade' `
- install `pip install elephant --install-option='--no-compile-spade' `

This option allows to have a workaround in cases were there is no suitable c-compiler available on the target system.

Additionally extensions are now optional, according to [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html?highlight=extension#building-extension-modules):
```python
 class setuptools.Extension(name, sources, *args, **kw)
```
- optional (bool) – specifies that a build failure in the extension should not abort the build process, but simply not install the failing extension.

See elephant documentation:

https://elephant--494.org.readthedocs.build/en/494/install.html

PR #608, opened 2023-11-22
[ENH] Move to pyproject.toml

Moritz-Alexander-Kern:
This is a long one!

# Original PR Body by Moritz
This PR moves Elephant from setup.py to pyproject.toml, aligning with PEP 518.

Key changes include:

- Introduction of `pyproject.toml`
- `setup.py` is still needed to compile the C-code for the fim module.

The reason is that we need OS dependend compile args which are dynamically defined in `setup.py`.
The `pyproject.toml` does not natively support conditional logic like os_name == 'windows' directly within the configuration. The TOML format is static and doesn't allow for dynamic evaluation of conditions based on the operating system.

To achieve platform-specific compile arguments, we have to use a Python script, such as a `setup.py`, which can programmatically determine the OS and apply the correct compile arguments.

In this context, as a configuration file for setuptools, `setup.py` is not deprecated, see also: https://packaging.python.org/en/latest/discussions/setup-py-deprecated/

---
# Changes

## pyproject.toml
- Created `pyproject.toml` with full project metadata (name, version, description, authors, license,
  URLs, classifiers)
- Moved to the SPDX licence expression format per PEP 639; removed the legacy `License ::` classifier string
- Switched to auto-discovery of packages via `setuptools.find_packages()`, this correctly picks up
subpackages such as `elephant.asset`
- Added `include-package-data = true` so non-Python data files are included in the wheel install
- Added classifiers for Python 3.13 and 3.14
- Removed the long-unused `six` dependency
- Added `viziphant` to the `tutorials` optional-dependency group
- Dependencies are now declared via [dynamic metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata), pointing back to the `requirements/*.txt` files. This keeps a single source of truth for dependencies

## CI modernisation (`.github/workflows/CI.yml`)
- **Merged pip test jobs**: Three separate pip-test jobs (Ubuntu, macOS, Windows) merged into a single job with `matrix.include`, eliminating duplication
- **Composite action**: Added a composite action(`.github/actions/restore-elephant-data`) to avoid code duplication of restore action that restored cached elephant-dataset and is shared by the pip, MPI and conda jobs
- Enabled `enableCrossOsArchive: true` on the data cache so a cache built on one OS can be restored on
another, this was added to build to both restore action and cache build action (`.github/workflows/cache_elephant_data.yml`)
- Bumped `actions/cache/`, `actions/setup-python` and `actions/checkout` to newer versions to fix issue with Node.js 20 actions being deprecated
- Made python version common accross jobs
- **Per-job coveralls uploads + finish job**: every job now uploads its own partial coverage report to coveralls, and a new `finish` job combines them at the end. This is to fix the earlier race condition where last coverall job was being usede to add PR comment.
- Added `pytest-cov` to `requirements/requirements-tests.txt` to avoid duplicate install steps for each job
- **Fast CI mode**: added a `fast` workflow-dispatch boolean input that limits the matrix to a subset of jobs
- Cleaned up workflow triggers 
- Cleaned up the DOC BUILD and DOC TEST jobs

## Environment files
- Refactored `requirements/environment*.yml` conda environment files to import from the corresponding `requirements/*.txt` files instead of duplicating the dependency list (system-level conda-only packages are still listed directly) 

## Bug fix for Windows path
- `elephant/datasets.py`: switched to `local_file.as_posix()` when constructing the `ValueError` message in `download_datasets`. On Windows, `pathlib` rendered backslashes, causing `test_download_with_cache_dir_with_failed_integrity_check` to fail because the assertion looks for the POSIX-style `repo_path` substring.


Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern (commented):
build wheels

Moritz-Alexander-Kern:
Editable install with `setup.py` will be deprecated with pip 25.0

> DEPRECATION: Legacy editable install of elephant==1.2.0b1 from file:/// (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

Moritz-Alexander-Kern (commented):
build wheels

CozySocksAlways:
- [ ] Update requirements file to remove upper bound matplot lib (notebooks were okay as well as sphinx docs).
- [x] Add viziphant to tutorials requirements.

setup.py
```diff
@@ -30,7 +15,7 @@
         extra_compile_args=[
             '-DMODULE_NAME=fim', '-DUSE_OPENMP', '-DWITH_SIG_TERM',
             '-Dfim_EXPORTS', '-fopenmp', '/std:c++17'],
-        optional=True
+        #optional=True
```
CozySocksAlways: Why was this commented out? 
- Was added in  #494 to overcome the compile fail issue, but didn't work at the time with development (editable) install?
- Commenting it out will cause `optional` defaulting to `None` which is Falsy. So install fails on build?

CozySocksAlways (commented):
trigger build wheels ci

CozySocksAlways:
Postponing addition of GPU tests with group runner available on jugit.fz-juelich.de to a different PR.

CozySocksAlways:
Once PR approved, full CI should be triggered. (Run once for each PR approval)

pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+        "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
+        "Programming Language :: Python :: 3.13",
+        "Programming Language :: Python :: 3.14",
+        "Programming Language :: Python :: 3 :: Only",
+        "Topic :: Scientific/Engineering"]
+keywords = ["neuroscience", "neurophysiology", "electrophysiology", "statistics", "data-analysis"]
+readme = {file = "README.md", content-type = "text/markdown"}
+requires-python = ">=3.8"
```
mdenker: I feel like this line is giving me problems when doing the following:
- Create a new environment using `uv venv`
- Doing a `uv pip install .` in the elephant directory
- Calling `uv run python`m which gives me an error about dependencies not being satisfyable:uv responds that while I am currently using Python 3.14, the `reuires-python` indicates >=3.8. This is however incompatible with numpy 2. Therefore, uv fails, as it cannot resolve for any of the support python versions. I do not fully understand this logic, however, I guess we should drop 3.8 despite this being a hot fix. Or maybe I am missing something?

mdenker (changes requested):
Thank you for this extensive PR, besides the comment about Python versions, it looks good to me so far -- I guess we need to see how the remaining CI jobs work out. I tested installs on Linux and Windows, which seem to work (including the fim module unter linux). However, the Python 3.8 issue that prevents installation of the package in a uv managed pip environment is something we should address.

pyproject.toml
CozySocksAlways: Yup I could recreate this issue, it's an error from my side. The setup.py on our latest release has a requirement on `python>=3.9`, so I'll fix this now on the toml file. 

As a side note, the latest release of numpy (2.4.6) has a requirement of `python>=3.11`, though lower version of numpy 2.0 still support python 3.9. So we should definitely fix this on our next minor release.

CozySocksAlways:
Test duplication in case of pull_request_review trigger (i.e. both pull_request_review triggered tests pollutes the existing PR checks summary made with the checks triggered on PR synchronize) is the right behavior on GitHub and cannot be avoided as GitHub treats these two as separate events. Refer https://github.com/orgs/community/discussions/26940

As some form of relief, an if clause has been added on the setup job that makes it only run when pull_request_review is approved and not when a review comment is added.

kohlerca (approved):
The PR looks good to me. I tested locally in Ubuntu, using `mamba` and `uv`, and the installation succeeds.

CozySocksAlways:
- [x] Update names of tests for consistency (i.e with MPI and conda)

mdenker (approved):

PR #610, opened 2023-12-11
[Feature] Integrate trials object with GPFA

Moritz-Alexander-Kern:
This pull request introduces an enhancement to GPFA using the the `elephant.trials` module to represent trials. This change is part of a greater effort to provide a more structured and unified approach for handling trial data in Elephant.

## Trial representations
-  **trial object**: The new trial_object is now used as the primary input structure for representing trials. This object handles trial data, see <https://elephant.readthedocs.io/en/latest/reference/trials.html> .

- **list of lists**: The previous implementation relied on a nested structure of lists to represent trials.


elephant/utils.py
```diff
@@ -372,3 +394,54 @@ def get_opencl_capability():
...
+    ValueError: If the number of trials in the Trials object is 0.
+
+    Example:
+        @trials_to_list_of_spiketrainlist
+        def process_data(self, spiketrains):
+            ...
+    """
+
+    @wraps(method)
+    def wrapper(self, *args, **kwargs):
```
kohlerca, line 423: In my opinion, the decorator could be made generic by just having `*args`, `**kwargs` here. Then it would work for both pure functions or methods that have the object instance as the first argument, and be more reusable throughout Elephant. Any reason to have it just for methods?

elephant/utils.py
```diff
@@ -372,3 +394,54 @@ def get_opencl_capability():
         return True
     except ImportError:
         return False
+
+
+def trials_to_list_of_spiketrainlist(method):
+    """
+    Decorator to convert Trials object to a list of spiketrainlists before
```
kohlerca, line 401: Just small fixes in the docstrings: backticks in Trials, spiketrainlists is not the actual Python classs

elephant/utils.py
```diff
@@ -372,3 +394,54 @@ def get_opencl_capability():
...
+    calling the wrapped method.
+
+    Parameters
+    ----------
+    method: callable, The method to be decorated.
```
kohlerca, line 406: Here we put the description in a new line

elephant/utils.py
```diff
@@ -372,3 +394,54 @@ def get_opencl_capability():
...
+
+    Returns
+    -------
+    callable: The decorated method.
```
kohlerca, line 410: idem

elephant/utils.py
```diff
@@ -372,3 +394,54 @@ def get_opencl_capability():
...
+
+    Raises
+    ------
+    ValueError: If the number of trials in the Trials object is 0.
```
kohlerca, line 414: idem

elephant/utils.py
```diff
@@ -372,3 +394,54 @@ def get_opencl_capability():
...
+
+    Example:
```
kohlerca, line 416: The example section also has a specific header

kohlerca (changes requested):
The functionality to make the conversion looks good to me. I made some small suggestions to improve the code and docstrings.

elephant/utils.py
Moritz-Alexander-Kern: Thanks, good catch. This was a leftover the previous way of implementing this.
Fix: `self`- removed.

elephant/utils.py
Moritz-Alexander-Kern: Fixed.

elephant/utils.py
Moritz-Alexander-Kern: Fixed.

elephant/utils.py
Moritz-Alexander-Kern: Fixed.

elephant/utils.py
Moritz-Alexander-Kern: Fixed.

elephant/utils.py
Moritz-Alexander-Kern: Fixed.

Issue #493, opened 2022-06-08
[Bug]  IndexError in spike_train_synchrony.py, annotate_synchrofacts


Moritz-Alexander-Kern:
This Bug was originally discovered by @skrausse , thank you for reporting.

**Describe the bug**
When trying to use the `delete_synchrofacts` method from `Synchrotool` class in  `spike_train_synchrony.py`, in specific cases an IndexError is raised.

**To Reproduce**
1.  Install elephant and neo
2. Run the following minimal example:
```python
# fix synchrotool
import elephant, neo
import quantities as pq
from elephant import spike_train_synchrony


import elephant.spike_train_generation as stg
import numpy as np

np.random.seed(1)  # set random seed 2 for no error

SPP = stg.StationaryPoissonProcess(50 * pq.Hz, t_start=0 * pq.ms,
                                   t_stop=1000*pq.ms)
test_sts = SPP.generate_n_spiketrains(2)

sampling_rate = 30000*pq.Hz
test_obj = spike_train_synchrony.Synchrotool(
    test_sts, sampling_rate=sampling_rate, spread=2)
test_obj.delete_synchrofacts(threshold=2, in_place=True)
```

Gives the following error messages: (if `np.random.seed(1)`, if the random seed is set to 2 , no error is raised)
```
Traceback (most recent call last):
    test_obj.delete_synchrofacts(threshold=2, in_place=True)
  elephant/spike_train_synchrony.py", line 325, in delete_synchrofacts
    self.annotate_synchrofacts()
  elephant/spike_train_synchrony.py", line 407, in annotate_synchrofacts
    complexity_per_spike = epoch_complexities[spike_to_epoch_idx]
IndexError: index 91 is out of bounds for axis 0 with size 91
```

**Expected behavior**
Index should not be out of bounds

**Environment**
 - OS (e.g., Linux):Ubuntu 20.04.4 LTS (64-bit)
 - How you installed elephant (`conda`, `pip`, source): pip, source
 - Python version: 3.8
 - `neo` python package version: 0.10.2
 - `elephant` python package version: 0.11.0
 - `quantities`: 0.13.0
 - `numpy`: 1.22.3

Moritz-Alexander-Kern:
Hot-fix:
Added an if-clause, which checks if index is out of bounds, see commit: 786194c2378ad866776f74c6c510512d39177fd7
https://github.com/INM-6/elephant/tree/fix/synchrotool

[...]

PR #612, opened 2023-12-26
[Bug fix] Fix bug of synchrotool index out of bounds

morales-gregorio:
Hi,

The bug was a known issue, when applying the synchrotool if there was a synchrofact in the last bin the indices went out of bounds. Here a minimal code to reproduce the error:
```python
import neo
import numpy as np
import quantities as pq
from elephant.spike_train_synchrony import Synchrotool

sampling_rate = 1/pq.ms
st = neo.SpikeTrain(np.arange(0, 11)*pq.ms, t_start=0*pq.ms, t_stop=10*pq.ms)

synchrotool_instance = Synchrotool([st, st], sampling_rate, spread=0)
synchrotool_instance.annotate_synchrofacts()
```
which (with the master branch elephant, python  3.9.7) produces the output:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[2], line 5
      2 st = neo.SpikeTrain(np.arange(0, 11)*pq.ms, t_start=0*pq.ms, t_stop=10*pq.ms)
      4 synchrotool_instance = Synchrotool([st, st], sampling_rate, spread=0)
----> 5 synchrotool_instance.annotate_synchrofacts()

File elephant/elephant/spike_train_synchrony.py:414, in Synchrotool.annotate_synchrofacts(self)
    406 for idx, st in enumerate(self.input_spiketrains):
    407 
    408     # all indices of spikes that are within the half-open intervals
    409     # defined by the boundaries
    410     # note that every second entry in boundaries is an upper boundary
    411     spike_to_epoch_idx = np.searchsorted(
    412         right_edges,
    413         st.times.rescale(self.epoch.times.units).magnitude.flatten())
--> 414     complexity_per_spike = epoch_complexities[spike_to_epoch_idx]
    416     st.array_annotate(complexity=complexity_per_spike)
    418 self.annotated = True

IndexError: index 10 is out of bounds for axis 0 with size 10
```
_**The following is outdated**_
_To fix this issue I suggest a very simple change in the definition of the bins. To avoid spikes at bin edges (when binning at sampling resolution) the bins were shifted by half a bin width to the left. This leads to one additional (half)-bin at the end of the time series, which only raises the issue when a synchrofact is found in this last bin, which is only rarely in real data. Instead of shifting to the left, I propose shifting to the right, this leads to two initial spikes being in the first bin, but no extra bin at the end._

Let me know what you think,
Aitor

PS: I see some of the tests fail with this change due to small differences, let me know if this change is ok and I will fix the tests


Moritz-Alexander-Kern:
Hey Aitor, 
thanks for suggesting those changes and creating a pull request. :tada:

This seems to be similar to the discussion we had on this issue, see https://github.com/NeuralEnsemble/elephant/issues/493 .

The question seems to be whether we agree on those changes with respect to the definition of the bins, the code looks good to me.

Regarding the small differences in the results which lead to failing unit tests: we need to discuss whether those differences are significant and pose a problem with respect to reproducibility of previous results calculated with the "old" way of defining the bins.

morales-gregorio:
As discussed in the meeting on 2024.02.05, BinnedSpiketrain is not the correct place to fix this since it can have unexpected ramifications. Instead the Fix by @skrausse and @jo460464 was transferred to this PR

elephant/statistics.py
```diff
@@ -1550,11 +1550,12 @@ def _epoch_no_spread(self):
...
-            # Note: all spike trains share the same t_start and t_stop.
-            if left_edges[0] < self.t_start:
-                left_edges[0] = self.t_start
-                durations[0] -= bin_shift
+        # Ensure that an epoch does not start before the minimum t_start.
+        # Note: all spike trains share the same t_start and t_stop.
+        if left_edges[0] < self.t_start:
+            left_edges[0] = self.t_start
+            durations[0] -= bin_shift
+
```
Moritz-Alexander-Kern, line 1558: Was this change of the indentation intentional?  Consider the first if-clause in connection with the else statement:
```python
if self.sampling_rate:
    ....
else:
            warnings.warn('No sampling rate specified. ' ...
```

Now we have 
```python
if left_edges[0] < self.t_start:
    ...
else:
            warnings.warn('No sampling rate specified. ' ...
```

mdenker: I agree with the assessment. The variable `bin_shift` is set in the first `if`-clause that checks for the sampling rate. Thus, if this clause  is not executed, `bin_shift` is not defined. However, the second `if`-clause that test for `left_edges` uses the variable `bin_shift`. This will through an exception if `self.sampling_rate` is `False`. Therefore, I also think that the if clause needs to be either shifted back to the original state -- or one must do something else if there are other issues beyond the syntactic analysis that initiated this change.

morales-gregorio: My bad, the indentation was accidental

Moritz-Alexander-Kern:
Hey @morales-gregorio , 
thanks for the contribution.

This is still open?
> PS: I see some of the tests fail with this change due to small differences, let me know if this change is ok and I will fix the tests

morales-gregorio:
Hi, yes I am looking at the tests now

morales-gregorio:
Turns out there is a very good reason why those tests were failing. Shifting the bins rightward causes unpredictable behavior when there are coincident spikes in the first bin, which messed with the output of the complexity histogram, annotation etc.

I propose a new simplified method, by extending the t_stop by one bin (this does not modify the t_stop of the original spiketrains!). This solves the problem without all the additional weirdness around the first bin.

morales-gregorio:
Hi @Moritz-Alexander-Kern the tests are now all passing, something missing for merge?

Moritz-Alexander-Kern:
Hi @morales-gregorio,

> Hi @Moritz-Alexander-Kern the tests are now all passing, something missing for merge?

Thank you for the update. With regards to your question about the merge, everything looks good on the testing front – all tests are passing. However, before proceeding with the merge, there are a couple of points:

**Approach Change:**
We need to re-review, since the last review, the approach was changed:

> I propose a new simplified method, by extending the t_stop by one bin (this does not modify the t_stop of the original > spiketrains!). This solves the problem without all the additional weirdness around the first bin.

The previous understanding regarding the approach no longer stands, as the fix by @skrausse and @jo460464 is no longer utilized here. i.e. this no longer holds true:

>Instead the Fix by @skrausse and @jo460464 was transferred to this PR

What was discussed regarding this PR is therefore also outdated, it seems only fair to update the involved people on the new status and allow them to update their comments. (The PR description is also outdated)

Thanks again for your attention to this matter and for the progress made thus far.

doc/authors.rst
```diff
@@ -51,6 +51,8 @@ contribution, and may not be the current affiliation of a contributor.
 * Florian Porrmann [13]
 * Sarah Pilz [13]
 * Oliver Kloß [1]
+* Jonas Oberste-Frielinghaus [1]
+* Sven Krausse [1]
```
Moritz-Alexander-Kern, line 55: Are Jonas and Sven aware and in agreement of the new approach to change `t_stop`?
Sorry, I might have missed or forgotten this part.
I do renember they have been involved with the first version of the suggested fix.

The affilaition of Jonas and you chnaged to IAS-6, refer to `authors.rst` in v1.1.0, reference [14].

https://github.com/NeuralEnsemble/elephant/blob/15e8ef8e2a3b9853f9d7ca6bffd158c56cb545ba/doc/authors.rst?plain=1#L69

elephant/test/test_spike_train_synchrony.py
```diff
@@ -408,6 +413,28 @@ def test_binning_for_input_with_rounding_errors(self):
...
+        spiketrains = [st, st]
+
+        annotations = 2*np.ones(10)
+
+        correct_annotations = np.array([annotations, annotations])
+
+        self._test_template(spiketrains, correct_annotations, sampling_rate,
+                            spread=0, mode='delete', in_place=True,
+                            deletion_threshold=2)
+
```
Moritz-Alexander-Kern, line 437: See review comment regarding this unit test.

Moritz-Alexander-Kern (changes requested):
Hey @morales-gregorio,

thanks again for your work.

I added some `print` statements to the code example supplied in the PR description:

```python
import neo
import numpy as np
import quantities as pq
from elephant.spike_train_synchrony import Synchrotool

sampling_rate = 1/pq.ms
st = neo.SpikeTrain(np.arange(0, 11)*pq.ms, t_start=0*pq.ms, t_stop=10*pq.ms)
print(f"Spike train before Synchrotool: {st}")
print(f"Spike train t_stop before Synchrotool: {st.t_stop}")

synchrotool_instance = Synchrotool([st, st], sampling_rate, spread=0)
synchrotool_instance.annotate_synchrofacts()

print(f"Spike train after Synchrotool: {st}")
print(f"Spike train t_stop after Synchrotool: {st.t_stop}")
```

I get the following output:
```
Spike train before Synchrotool: [ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10.] ms
Spike train t_stop before Synchrotool: 10.0 ms
Spike train after Synchrotool: [ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10.] ms
Spike train t_stop after Synchrotool: 11.0 ms
```

Is this intentional? I anticipated that the input spike trains would remain unchanged, the spike times are, but `t_stop` seems to be changed.

**Note**
I did not take the general approach to change the behavior of the `Complexity`- class into account with this review.

requirements/environment.yml
```diff
@@ -13,6 +13,6 @@ dependencies:
   - statsmodels
   - jinja2
   - pip:
-    - neo>=0.10.0
+    - neo>=0.13.0
```
Moritz-Alexander-Kern, line 16: Could you please also update this in `requirements/requirements.txt`?
Thank you :)

elephant/statistics.py
```diff
@@ -1555,6 +1569,7 @@ def _epoch_no_spread(self):
             if left_edges[0] < self.t_start:
                 left_edges[0] = self.t_start
                 durations[0] -= bin_shift
+
```
Moritz-Alexander-Kern, line 1572:
```suggestion
```

requirements/environment.yml
morales-gregorio: done

morales-gregorio:
> Hey @morales-gregorio,
> 
> thanks again for your work.
> 
> I added some `print` statements to the code example supplied in the PR description:
> 
> ```python
> import neo
> import numpy as np
> import quantities as pq
> from elephant.spike_train_synchrony import Synchrotool
> 
> sampling_rate = 1/pq.ms
> st = neo.SpikeTrain(np.arange(0, 11)*pq.ms, t_start=0*pq.ms, t_stop=10*pq.ms)
> print(f"Spike train before Synchrotool: {st}")
> print(f"Spike train t_stop before Synchrotool: {st.t_stop}")
> 
> synchrotool_instance = Synchrotool([st, st], sampling_rate, spread=0)
> synchrotool_instance.annotate_synchrofacts()
> 
> print(f"Spike train after Synchrotool: {st}")
> print(f"Spike train t_stop after Synchrotool: {st.t_stop}")
> ```
> 
> I get the following output:
> 
> ```
> Spike train before Synchrotool: [ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10.] ms
> Spike train t_stop before Synchrotool: 10.0 ms
> Spike train after Synchrotool: [ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10.] ms
> Spike train t_stop after Synchrotool: 11.0 ms
> ```
> 
> Is this intentional? I anticipated that the input spike trains would remain unchanged, the spike times are, but `t_stop` seems to be changed.
> 
> **Note** I did not take the general approach to change the behavior of the `Complexity`- class into account with this review.

Hi, I thought this did not happen but I also now reproduced it in my implementation. I agree this could lead to some confusion. 

The options I suggest are either:
1. Work exclusively on a copy of the spiketrains, this will double the memory usage, which was a concern of this implementation.
2. Keep it as is but raise a warning when t_stop is modified, so the users know that it has happened.

I lean towards option 2, since the extension of t_stop by 1 time bin is minimal when working with long recordings.

pbouss:
Hi @morales-gregorio  and @Moritz-Alexander-Kern, 
can this fix be merged to elephant? 
It would help me for one of me projects. 
Thanks, Peter

Moritz-Alexander-Kern:
Hi @pbouss,

Could you please clarify which of the proposed fixes (A, B, or C) you believe would be most beneficial for your project?

A) The approach suggested here to modify t_stop of the spike train (This PR).

B) or C) Proposed fix: Condition on binned spiketrain – if the sampling rate is provided and the shift parameter is allowed, then shift the spiketrains; otherwise, warn about spikes being removed if they coincide with t_stop.

_Originally posted by @skrausse in https://github.com/NeuralEnsemble/elephant/issues/493#issuecomment-1774894355_

It's important to note that the underlying issue is not a bug with the code per se, but rather a decision regarding how to handle situations where the spike trains do not conform to the assumptions made by the author of the synchrotool regarding the input spike trains. Workarounds could include checking spike trains in advance and possibly changing t_stop before using synchrotool or removing/masking spikes.

Please also read #493, where this was discussed previously.

The "fix" might be different depending on your use-case; please understand that this is why there is no straightforward solution.

Moritz-Alexander-Kern:
Consider the following example:

```python
import quantities as pq
from elephant import spike_train_synchrony

import elephant.spike_train_generation as stg
import numpy as np

np.random.seed(1)  # set random seed 2 for no error

SPP = stg.StationaryPoissonProcess(50 * pq.Hz, t_start=0 * pq.ms,
                                   t_stop=1000*pq.ms)
test_sts = SPP.generate_n_spiketrains(2)

sampling_rate = 1/pq.ms

test_obj = spike_train_synchrony.Synchrotool(
    test_sts, sampling_rate=sampling_rate, spread=2)
test_obj.annotate_synchrofacts()
```

With the proposed fix I still get:

```
  File "git/INM-6/elephant/build/test2_612_snchrotool_index_out_of_bounds.py", line 20, in <module>
    test_obj.annotate_synchrofacts()
  File "git/INM-6/elephant/elephant/spike_train_synchrony.py", line 410, in annotate_synchrofacts
    complexity_per_spike = epoch_complexities[spike_to_epoch_idx]
IndexError: index 72 is out of bounds for axis 0 with size 72
```

So does this PR fix your issue @pbouss ? Maybe you can test this approach by installing from this branch with:

```bash
pip install git+https://github.com/morales-gregorio/elephant.git@synchrotool_fix
```

pbouss:
> Consider the following example:
> 
> ```python
> import quantities as pq
> from elephant import spike_train_synchrony
> 
> import elephant.spike_train_generation as stg
> import numpy as np
> 
> np.random.seed(1)  # set random seed 2 for no error
> 
> SPP = stg.StationaryPoissonProcess(50 * pq.Hz, t_start=0 * pq.ms,
>                                    t_stop=1000*pq.ms)
> test_sts = SPP.generate_n_spiketrains(2)
> 
> sampling_rate = 1/pq.ms
> 
> test_obj = spike_train_synchrony.Synchrotool(
>     test_sts, sampling_rate=sampling_rate, spread=2)
> test_obj.annotate_synchrofacts()
> ```
> 
> With the proposed fix I still get:
> 
> ```
> Traceback (most recent call last):
>   File "git/INM-6/elephant/build/test2_612_snchrotool_index_out_of_bounds.py", line 20, in <module>
>     test_obj.annotate_synchrofacts()
>   File "git/INM-6/elephant/elephant/spike_train_synchrony.py", line 410, in annotate_synchrofacts
>     complexity_per_spike = epoch_complexities[spike_to_epoch_idx]
> IndexError: index 72 is out of bounds for axis 0 with size 72
> ```
> 
> So does this PR fix your issue @pbouss ? Maybe you can test this approach by installing from this branch with:
> 
> ```shell
> pip install git+https://github.com/morales-gregorio/elephant.git@synchrotool_fix
> ```

Hi @Moritz-Alexander-Kern, thanks for your message. I understand a bit more why there are so many options. 
I'm already happy with what is done here in this branch. I used this now for the SPADE analysis work flow (as you proposed) and there it was running without any error. 
It's strange that it still does not work for the example you show here. Do you know why the fix a) does not work in this case here?

Moritz-Alexander-Kern:
Hey,

Great to hear this fixes your issue! 🎉

Regarding the example I provided, @Kleinjohann pointed out that the spiketrains lack a sampling rate. Synchrotool assumes that spiketrains are discrete and that all spike times are multiples of the provided sampling rate. The suggested fix is to discretize the spiketrains before using Synchrotool:

```python
from elephant.conversion import discretise_spiketimes
test_sts = discretise_spiketimes(test_sts, sampling_rate)
```

But you can also "fix" the example I provided by changing the sampling rate to `sampling_rate = 10/pq.ms` or by setting `spread=2`.

There seem to be multiple ways this issue can arise and several potential solutions, depending on which assumptions about the input spiketrains are not met.

So far, the proposed fixes involve modifying the spiketrains, such as discretizing them or changing t_stop, to align with Synchrotool's assumptions.

This raises a question: should Synchrotool handle these cases, or is it the user's responsibility to ensure the spiketrains comply with Synchrotool's assumptions? 

The fix proposed here could also be implemented manually, or is there a reason why this must be implemented in Synchrotool ?
Pseudo Code:
```bash
# Check if spikes happen in the last bin
for st in input_spiketrains:
     # Extend t_stop to avoid indexing problems
     if last spike is close to t_stop
         st.t_stop += st.bin_size
```

Now you tell me:  Is there a bug in Synchrotool or are the input spike trains "wrong"?

pbouss:
> Hey,
> 
> Great to hear this fixes your issue! 🎉
> 
> Regarding the example I provided, @Kleinjohann pointed out that the spiketrains lack a sampling rate. Synchrotool assumes that spiketrains are discrete and that all spike times are multiples of the provided sampling rate. The suggested fix is to discretize the spiketrains before using Synchrotool:
> 
> ```python
> from elephant.conversion import discretise_spiketimes
> test_sts = discretise_spiketimes(test_sts, sampling_rate)
> ```
> 
> But you can also "fix" the example I provided by changing the sampling rate to `sampling_rate = 10/pq.ms` or by setting `spread=2`.
> 
> There seem to be multiple ways this issue can arise and several potential solutions, depending on which assumptions about the input spiketrains are not met.
> 
> So far, the proposed fixes involve modifying the spiketrains, such as discretizing them or changing t_stop, to align with Synchrotool's assumptions.
> 
> This raises a question: should Synchrotool handle these cases, or is it the user's responsibility to ensure the spiketrains comply with Synchrotool's assumptions?
> 
> The fix proposed here could also be implemented manually, or is there a reason why this must be implemented in Synchrotool ? Pseudo Code:
> 
> ```shell
> # Check if spikes happen in the last bin
> for st in input_spiketrains:
>      # Extend t_stop to avoid indexing problems
>      if last spike is close to t_stop
>          st.t_stop += st.bin_size
> ```
> 
> Now, let me know: Is there a bug in Synchrotool, or are the input spike trains incorrect?

In my case, I'm pretty sure the input spike trains are correct and they should also have a sampling_rate, as it's based on the R2G-experiment which has a sampling rate. 
In general, I should go for Synchrotool should always give you a reasonable output or at least it gives the user an error message which is solvable. But in my case, it was giving in 3 of 20 sessions the IndexError, where there was the same code for all the sessions that is then much better to have the slight change in t_stop and everything runs smoothly.

Moritz-Alexander-Kern:
ahh ok I see.
Which of the suggested options from morales-gregorio do you prefer?

> Hi, I thought this did not happen but I also now reproduced it in my implementation. I agree this could lead to some confusion.
> 
> The options I suggest are either:
> 
> 1. Work exclusively on a copy of the spiketrains, this will double the memory usage, which was a concern of this implementation.
> 
> 2. Keep it as is but raise a warning when t_stop is modified, so the users know that it has happened.
> 
> 
> I lean towards option 2, since the extension of t_stop by 1 time bin is minimal when working with long recordings.

pbouss:
> ahh ok I see. Which of the suggested options from morales-gregorio do you prefer? (I guess Option 2?)
> 
> > Hi, I thought this did not happen but I also now reproduced it in my implementation. I agree this could lead to some confusion.
> > The options I suggest are either:
> > 
> > 1. Work exclusively on a copy of the spiketrains, this will double the memory usage, which was a concern of this implementation.
> > 2. Keep it as is but raise a warning when t_stop is modified, so the users know that it has happened.
> > 
> > I lean towards option 2, since the extension of t_stop by 1 time bin is minimal when working with long recordings.

I think the second option is better, the change is minimal, especially for the bins considered in the Synchrotool and the warning gives you the chance to make your own fix if you like.

Moritz-Alexander-Kern:
I see, thanks for the feedback.

Further, I suggest to also consider the fix suggested by @jo460464 and @skrausse , see: https://github.com/jo460464/elephant/tree/enh/iss101

Install with: (make sure you use scipy<1.14.0)
```bash
pip install git+https://github.com/jo460464/elephant.git@enh/iss101
```
This approach does not change the spike trains but might also work for you?

Moritz-Alexander-Kern:
> As discussed in the meeting on 2024.02.05, BinnedSpiketrain is not the correct place to fix this since it can have unexpected ramifications. Instead the Fix by @skrausse and @jo460464 was transferred to this PR

I believe this was the fix originally intended to be implemented with this PR.

pbouss:
> I see, thanks for the feedback.
> 
> Further, I suggest to also consider the fix suggested by @jo460464 and @skrausse , see: https://github.com/jo460464/elephant/tree/enh/iss101
> 
> Install with: (make sure you use scipy<1.14.0)
> 
> ```shell
> pip install git+https://github.com/jo460464/elephant.git@enh/iss101
> ```
> 
> This approach does not change the spike trains but might also work for you?

Thank you. I think for now I'm fine with what I have working. For the next iteration, I will also include this.

Moritz-Alexander-Kern:
Closed in favorite of  #637

PR #446, opened 2022-02-04
Fixes neo_tools unit tests to work with Neo 0.10.0+

mdenker:
In Neo 0.10.0, several helper functions from the test suite, such as  `fake_neo()`, have been removed and replaced by a cleaner and more light-weight implementation. Elephant tests had been using this functionality in the `neo_tools` module. This PR changes all tests to use the new implementations from Neo 0.10., and bumps minimum Neo requirements accordingly. Several deprecated objects, such as `Unit` and `ChannelGroup` have also been removed from the tests.

PR #617, opened 2024-02-07
[FIX] failing unitests for neo_tools with Neo 0.13.0

Moritz-Alexander-Kern:
## Changes

- fixed failing unitests for `neo_tools`.
- fixed failing tests in `spike_train_synchrony`

## Background
With Neo v0.13.0  child lists (such as Segment.spiketrains, Segment.analogsignals, Block.segments, etc.) are now represented by pseudo-list objects that manage the necessary bookkeeping related to child-parent relations transparently.

See also: https://neo.readthedocs.io/en/latest/releases/0.13.0.html

This leads to  issues with unit tests in `neo_tools`, similar to the difficulties encountered with the pseudo list `SpikeTrainList` introduced with Neo version 0.10.0.

For Neo 0.10.0 this was fixed here: https://github.com/NeuralEnsemble/elephant/pull/446

**Update**

Neo Issue [Neo #1405](https://github.com/NeuralEnsemble/python-neo/issues/1405) was fixed. Now it is no longer possible to add the same object to one of the collections (pseudo-lists) twice. Therefore those parts of the unit-tests that creating data in this way have been removed.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -57,8 +57,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # python versions for elephant: [3.8, 3.9, "3.10", 3.11]
-        python-version: [3.8, 3.9, "3.10", 3.11]
+        # python versions for elephant: [3.8, 3.9, "3.10", 3.11, 3.12]
+        python-version: [3.9, "3.10", 3.11, 3.12]
         # OS [ubuntu-latest, macos-latest, windows-latest]
         os: [ubuntu-latest]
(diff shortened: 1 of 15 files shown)


Moritz-Alexander-Kern:
Update: You can no longer add the same object twice to a Neo ObjectList: https://github.com/NeuralEnsemble/python-neo/pull/1414 .

So unit tests no longer need to check for duplicates.

kohlerca (approved):
The changes look good to me.

mdenker (approved):
Looks good to me. I agree that we no longer need the "duplication test", as this is not allowed by Neo any longer.

PR #630, opened 2024-04-16
[MAIN] add ruff for autoformatting

Moritz-Alexander-Kern:
This PR adds autoformatting with ruff


Moritz-Alexander-Kern:
Check resulting PR for Elephant here: https://github.com/Moritz-Alexander-Kern/elephant-1/pull/9

PR #632, opened 2024-04-17
Fix date in codemeta.json

apdavison:


codemeta.json
```diff
@@ -6,7 +6,7 @@
     "contIntegration": "https://github.com/NeuralEnsemble/elephant/actions",
     "dateCreated": "2013-17-15",
```
apdavison, line 7: There's also something wrong with this date

Moritz-Alexander-Kern: Yes indeed, thanks for pointing this out.

PR #634, opened 2024-05-13
[Fix] Tests for Neo 0.13.1, add same object to list

Moritz-Alexander-Kern:
This is related to Neo PR https://github.com/NeuralEnsemble/python-neo/pull/1414 .

With this Neo PR  the same object can no longer be added more than once to a given list. 

Some tests in Elephant created test data by adding the same object multiple times to a list. Those tests have been fixed without changing the functionality of Elephant functions.

The tests ensuring that functions return only unique objects within a collection, even if the original list contains duplicates, have been revised. This change was made because the test is now conducted in Neo.


Moritz-Alexander-Kern (commented):
build wheels

Issue #415, opened 2021-04-13
[Doc] Spike Extraction parameters


myagmur01:
Hi @mdenker 

I am using elephant.spike_train_generation.spike_extraction function for my MEA 96-channel signal. 
I have following questions:

1. Why default values are selected as [-2ms, 4ms]

2. What is exactly the `time_interval` here? 
When I set it to [-10 ms,10 ms] or [0 ms, 10 ms] or [5 ms,15 ms] I get the same results of spike times as below. Could you explain me if I am doing something wrong?

```
spikes for [-10 ms,10 ms] =  [0.4288     5.9382     6.91823333 6.91963333 6.9386     7.71603333
 7.73576667 8.18213333 8.29553333 8.3024     8.30493333 8.31963333] s 

spikes for [0 ms, 10 ms] = [0.4288     5.9382     6.91823333 6.91963333 6.9386     7.71603333
 7.73576667 8.18213333 8.29553333 8.3024     8.30493333 8.31963333] s 

spikes for [5 ms,15 ms] = [0.4288     5.9382     6.91823333 6.91963333 6.9386     7.71603333
 7.73576667 8.18213333 8.29553333 8.3024     8.30493333 8.31963333] s 
```

Thank you

mdenker:
Hi @mytrjp ,

both questions relate to the spike waveforms which are extracted from the signal, i,e., the cutout of the signal around each spike.

The `time_interval` specifies the cut-out interval around each spike. -2 and 4 is of course somewhat arbitrary, however, many spike sorting tools choose values in this range: the spike is about 1-2 ms in length, and then you take a bit before the spike peak, and a bit more after the spike peak (to catch the dynamics after the spike).

If you set the `time_stamps` to `None`, peaks are extracted using numpy's peak finder. Therefore, the spike times will always be the same. However, the waveforms stored in the spike train (in `spikes.waveforms`, an NxM matrix for the N spikes, each row containing a waveform) should be different: In the first example, they should range from -10ms to 10ms, in the last example they should be from 5 to 15 ms. Thus, as an example, if the sampling frequency is 10KHz, and 123 spikes where detected, then spikes.waveforms should be size 123x200 for the first example.

I hope this helps. Thanks for asking this, I will keep this issue open as a reminder to augment the documentation to better explain this.

PR #641, opened 2024-09-23
[Doc] improve spike_train_generation.spike_extraction parameter documentation

Moritz-Alexander-Kern:
Enhanced description for the `spike_interval` parameter to clarify the bahavior and effects on the resulting spike train aswell as the choice of the chosen default values.

Added notes section to clarify the behavior when `time_stamps` is set to None and what the expected results are.
 
This PR adresses #415 .

### DIFF
--- a/elephant/spike_train_generation.py
+++ b/elephant/spike_train_generation.py
@@ -187,8 +187,8 @@ def spike_extraction(
     interval : tuple of :class:`pq.Quantity`
         Specifies the time interval around the `time_stamps` where the waveform
         is extracted. The default time interval [-2ms, 4ms] are based on experience,
-        and many spike sorting tools choose values in this range:
-        Spikes are typically about 1-2 ms in length. This choice of default values includes a small interval before
+        and many spike sorting tools choose values in this range as the
+        spikes are typically about 1-2 ms in length. This choice of default values includes a small interval before
         the spike peak and a slightly larger interval after the spike peak to capture the dynamics following
         the spike. Adjusting this interval will modify the waveforms stored in the neo :class:`neo.core.SpikeTrain`,
         but it will not affect the spike times.
--- a/elephant/spike_train_generation.py
+++ b/elephant/spike_train_generation.py
@@ -186,14 +186,19 @@ def spike_extraction(
         Default: None
     interval : tuple of :class:`pq.Quantity`
         Specifies the time interval around the `time_stamps` where the waveform
-        is extracted.
+        is extracted. The default values [-2ms, 4ms] are based on experience,
+        and many spike sorting tools choose values in this range:
+        Spikes are typically about 1-2 ms in length. This is to include a small interval before
+        the spike peak and a slightly larger interval after the spike peak to capture the dynamics following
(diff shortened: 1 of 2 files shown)


elephant/spike_train_generation.py
```diff
@@ -186,21 +186,34 @@ def spike_extraction(
         Default: None
     interval : tuple of :class:`pq.Quantity`
         Specifies the time interval around the `time_stamps` where the waveform
-        is extracted.
+        is extracted. The default values [-2ms, 4ms] are based on experience,
+        and many spike sorting tools choose values in this range:
+        Spikes are typically about 1-2 ms in length. This is to include a small interval before
```
kohlerca, line 191: I think it would read better by making a continuous sentence, e.g., "choose values in this range as the spikes are typically ..."

kohlerca, line 191: This choice of default values includes a small ...

elephant/spike_train_generation.py
```diff
@@ -186,21 +186,34 @@ def spike_extraction(
         Default: None
     interval : tuple of :class:`pq.Quantity`
         Specifies the time interval around the `time_stamps` where the waveform
-        is extracted.
+        is extracted. The default values [-2ms, 4ms] are based on experience,
```
kohlerca, line 189: If this is a tuple, it is better to use `(`, `)` instead of square brackets to be consistent

kohlerca: However, if the brackets are referring to the mathematical interval notation to the edges being inclusive, then it is clearer if "default values" is changed to the "default time interval"

elephant/spike_train_generation.py
```diff
@@ -186,21 +186,34 @@ def spike_extraction(
...
         Contains the time_stamps of each of the spikes and the waveforms in
         `result_st.waveforms`.
 
     See Also
     --------
     :func:`elephant.spike_train_generation.peak_detection`
+
+    Notes
+    -----
+    If `time_stamps` is set to None, peaks are extracted using numpy's peak finder. Therefore, the spike times will
```
kohlerca, line 212: numpy and neo are in lower case letters instead of NumPy and Neo

elephant/spike_train_generation.py
```diff
@@ -186,21 +186,34 @@ def spike_extraction(
...
+    not be affected by the `interval` parameter. However, the waveforms stored in the neo :class:`neo.core.SpikeTrain`
```
kohlerca, line 213: However, the dimensions of the array containing the waveforms, which is stored in the Neo ... object in `result_st.waveforms`, will differ depending on the `interval` parameter. The extracted waveforms are stored as an NxM matrix ...., each row containing a waveform. For example ....

elephant/spike_train_generation.py
kohlerca: I would favor the latter as in line 216, the values between brackets refer to the word interval.

kohlerca (changes requested):
I made some suggestions for clarity in the text.

PR #649, opened 2024-11-15
[Fix] instantaneous rate with list[neo.SpikeTrain]

Moritz-Alexander-Kern:
The pull request addresses inconsistent behavior with different options in combination with various inputs, specifically:

**Issue:**
- `pool_spiketrains = True` with `list[spiketrains]` does not pool spiketrains as expected. The output should be a single `neo.AnalogSignal`.

**Fix**
- check for other types, previously only `neo.spiketrainlist` was regarded 
- add tests

### DIFF
--- a/elephant/statistics.py
+++ b/elephant/statistics.py
@@ -715,7 +715,7 @@ def instantaneous_rate(spiketrains, sampling_period, kernel='auto',
         `sampling_period`.
 
         If `spiketrains` is :class:`neo.core.SpikeTrain` then output is a
-        single :class:`neo.core.AnalogSignal` with `n_estimates=1`.
+        single :class:`neo.core.AnalogSignal` with `n_estimates = 1`.
 
         If `spiketrains` is a list of :class:`neo.core.SpikeTrain` then output
         is a single :class:`neo.core.AnalogSignal`. `n_estimates` depends on the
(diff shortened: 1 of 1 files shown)


elephant/statistics.py
```diff
@@ -1031,7 +1031,7 @@ def optimal_kernel(st):
                              sigma=str(kernel.sigma),
                              invert=kernel.invert)
 
-    if isinstance(spiketrains, neo.core.spiketrainlist.SpikeTrainList) and (
+    if isinstance(spiketrains, (neo.core.spiketrainlist.SpikeTrainList, list, tuple)) and (
```
Moritz-Alexander-Kern, line 1034: check if is iterable instead of individual classes

elephant/utils.py
```diff
@@ -446,3 +449,32 @@ def wrapper(*args, **kwargs):
         return method(*new_args, **new_kwargs)
 
     return wrapper
+
+
+def is_list_neo_spiketrains(obj: object) -> bool:
```
Moritz-Alexander-Kern, line 454: `is_list_spiketrains`

elephant/utils.py
```diff
@@ -446,3 +449,32 @@ def wrapper(*args, **kwargs):
...
+        True if obj is an iterable containing only neo.SpikeTrain objects.
+    Raises
+    ------
+    TypeError
+        If obj is not an iterable, or if any element is not a neo.SpikeTrain.
+
+    """
+
+    if not isinstance(obj, collections.abc.Iterable):
+        raise TypeError("Input must be an iterable (list, tuple, etc.)")
```
Moritz-Alexander-Kern, line 475: no error raised, return false instead

elephant/test/test_utils.py
```diff
@@ -136,5 +136,42 @@ def test_decorator_return_with_list_of_lists_input_as_kwarg(self):
...
+
+    def test_mixed_types_input(self):
+        invalid_mixed = [self.spiketrain1, 42, self.spiketrain2]
+        with self.assertRaises(TypeError):
+            utils.is_list_neo_spiketrains(invalid_mixed)
+
+    def test_none_input(self):
+        with self.assertRaises(TypeError):
+            utils.is_list_neo_spiketrains(None)
+
```
Moritz-Alexander-Kern, line 174: add test with input: spiketrain

elephant/utils.py
```diff
@@ -446,3 +449,32 @@ def wrapper(*args, **kwargs):
...
+
+    Parameters
+    ----------
+    obj : object
+        The object to check. Can be a neo.spiketrainlist, list, tuple or any other iterable.
+
+    Returns
+    -------
+    bool
+        True if obj is an iterable containing only neo.SpikeTrain objects.
```
Moritz-Alexander-Kern, line 466: make return for single spiketrain explicit

elephant/utils.py
```diff
@@ -446,3 +449,32 @@ def wrapper(*args, **kwargs):
...
+
+
+def is_list_neo_spiketrains(obj: object) -> bool:
+    """
+    Check if input is an iterable containing only neo.SpikeTrain objects.
+
+    Parameters
+    ----------
+    obj : object
+        The object to check. Can be a neo.spiketrainlist, list, tuple or any other iterable.
```
Moritz-Alexander-Kern, line 461:
```suggestion
        The object to check.
```

elephant/utils.py
```diff
@@ -446,3 +449,32 @@ def wrapper(*args, **kwargs):
...
+
+    Returns
+    -------
+    bool
+        True if obj is an iterable containing only neo.SpikeTrain objects.
+    Raises
```
Moritz-Alexander-Kern, line 467:
```suggestion

```

elephant/statistics.py
```diff
@@ -1031,8 +1031,7 @@ def optimal_kernel(st):
                              sigma=str(kernel.sigma),
                              invert=kernel.invert)
 
-    if isinstance(spiketrains, neo.core.spiketrainlist.SpikeTrainList) and (
-                  pool_spike_trains):
+    if is_list_spiketrains(spiketrains) and (pool_spike_trains):
```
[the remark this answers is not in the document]
kohlerca: However, analyzing the function flow, the new and old checks for a list of `neo.SpikeTrain` objects may not be needed at all:

- in lines 908-911 it is ensured that there will be an Iterable with at least one spike train
- in lines 913-916 an error will be generated if the input is not an iterable that contains only spike trains

Therefore, at this point, it would be possible to only check if pooling or not the spike trains.

elephant/test/test_statistics.py
```diff
@@ -988,8 +988,44 @@ def test_instantaneous_rate_trials_pool_trials(self):
                                              pool_spike_trains=False,
                                              pool_trials=True)
         self.assertIsInstance(rate, neo.core.AnalogSignal)
+        self.assertEqual(rate.shape[1], self.trial_object.n_spiketrains_trial_by_trial[0])
```
kohlerca, line 991: This test checks for consistency between the dimensions of the computed rate and the data in the Trials objects. However, an additional test against the integer values would protect against errors in implementing the object attributes. These hard expected values are supposed to be known when designing the test data in line 493.

elephant/test/test_statistics.py
```diff
@@ -988,8 +988,44 @@ def test_instantaneous_rate_trials_pool_trials(self):
...
+    def test_instantaneous_rate_trials_pool_spiketrains(self):
+        kernel = kernels.GaussianKernel(sigma=500 * pq.ms)
+
+        rate = statistics.instantaneous_rate(self.trial_object,
+                                             sampling_period=0.1 * pq.ms,
+                                             kernel=kernel,
+                                             pool_spike_trains=True,
+                                             pool_trials=False)
+        self.assertIsInstance(rate, list)
+        self.assertEqual(len(rate), self.trial_object.n_trials)
```
kohlerca, line 1002: Same comment as above, where the test does not compare directly to the expected values, but retrieves them dynamically from the objects.

elephant/test/test_statistics.py
```diff
@@ -988,8 +988,44 @@ def test_instantaneous_rate_trials_pool_trials(self):
...
+    def test_instantaneous_rate_trials_pool_spiketrains_false_pool_trials_false(self):
+        kernel = kernels.GaussianKernel(sigma=500 * pq.ms)
+
+        rate = statistics.instantaneous_rate(self.trial_object,
+                                             sampling_period=0.1 * pq.ms,
+                                             kernel=kernel,
+                                             pool_spike_trains=False,
+                                             pool_trials=False)
+        self.assertIsInstance(rate, list)
+        self.assertEqual(len(rate), self.trial_object.n_trials)
```
kohlerca, line 1025: Same comment as above, where the test does not compare directly to the expected values, but retrieves them dynamically from the objects.

elephant/test/test_statistics.py
```diff
@@ -1010,7 +1058,7 @@ def test_instantaneous_rate_list_of_spike_trains(self):
             pool_spike_trains=False,
             pool_trials=False)
         self.assertIsInstance(rate, neo.core.AnalogSignal)
-        self.assertEqual(rate.magnitude.shape[1], 2)
+        self.assertEqual(rate.magnitude.shape[1], self.trial_object.n_spiketrains_trial_by_trial[0])
```
kohlerca, line 1061: Same comment as above, where the test does not compare directly to the expected values, but retrieves them dynamically from the objects. The integer in the old test could be updated, and the new one just added as an additional check.

elephant/test/test_statistics.py
```diff
@@ -988,8 +988,44 @@ def test_instantaneous_rate_trials_pool_trials(self):
                                              pool_spike_trains=False,
                                              pool_trials=True)
         self.assertIsInstance(rate, neo.core.AnalogSignal)
+        self.assertEqual(rate.shape[1], self.trial_object.n_spiketrains_trial_by_trial[0])
 
-    def test_instantaneous_rate_list_pool_spike_trains(self):
+    def test_instantaneous_rate_trials_pool_spiketrains(self):
```
kohlerca, line 993: A suggestion to improve understanding of these unit tests, where the dimensions of the outputs change depending on the pooling, is to include a comment block at the beginning to explicitly state the input dimensions --> expected output dimensions. It would be easier to assess the behavior of the object and what is being tested in each test case.

kohlerca (changes requested):
I checked the code and made suggestions. Regarding the proposed solution, this needs to be investigated further to ensure the problem was only due to a wrong class membership check. According to the comment on line 1034, the check that originated the error might be misplaced and redundant.

elephant/test/test_statistics.py
```diff
@@ -988,9 +999,19 @@ def test_instantaneous_rate_trials_pool_trials(self):
                                              pool_spike_trains=False,
                                              pool_trials=True)
         self.assertIsInstance(rate, neo.core.AnalogSignal)
-        self.assertEqual(rate.shape[1], self.trial_object.n_spiketrains_trial_by_trial[0])
+        self.assertEqual(rate.shape[1], self.n_spiketrains)
 
     def test_instantaneous_rate_trials_pool_spiketrains(self):
+        # Input:
+        #   Trails obj with self.n_trials, self.n_spiketrains
```
kohlerca: Typo

elephant/statistics.py
kohlerca: Typo: True instead of true

elephant/statistics.py
```diff
@@ -623,7 +623,7 @@ def instantaneous_rate(spiketrains, sampling_period, kernel='auto',
         Time stamp resolution of the spike times. The same resolution will
```
mdenker: I think this is a bit confusing, I would rather say time resolution of the resulting rate estimate. If this is set to the time resolution of the spike trains, then there will be no aliasing effect in the rate estimate (at the cose of a rather densely sampled output).

For example: 
Time resolution of the resulting rate estimate. The same resolution will be used for the kernel. To avoid effects of binning, set this value to the time stamp resolution of the spike train.

mdenker:
I have added a merge of the two docstring texts and smallified the table. Please have a look. For some reason, the doc string still prints bad on RTD, "Parameters" and "Returns" look squished on the side. I am not so sure what the reason is, perhaps it's the table that takes too much space...

mdenker:
As expected, the table breaks the formatting of Parameters and Returns sections. Suggestion: Put the table into the Notes (see current doc build). But maybe there are better ideas... Shorten the table? Secret Sphinx commands?

kohlerca:
I like the idea of the table. If the RTD is breaking within the Returns section, one option is to use the numbered notes format. Therefore, one can put a sentence such as `For a detailed summary of the possible outputs, see the table in Notes [3]` in Returns. If the notes are numbered instead of a bullet list, this will create a direct link within the documentation, and the table down the page can be easily accessed.

PR #666, opened 2025-06-16
[Fix] Failed ASSET GPU backend selection test when running in an environment with CUDA

kohlerca:
The unit test `test_asset_choose_backend_opencl` failed when running in a system with a GPU and CUDA libraries available, generating the error

`AssertionError: 'cuda' != 'cpu'`.

Interestingly, the test failed with or without PyCUDA installed in the environment, which is not the expected behavior. In a system without PyCUDA or OpenCL, it should always use the CPU backend.

It was detected that the utility function for CUDA capabilities detection `utils.get_cuda_capability_major` used the CUDA libraries in the system directly, which returns valid capabilities regardless of the availability of PyCUDA. The latter is required by the functions in ASSET when running CUDA GPU code. Therefore, the first fix in this PR modifies the function to return zero if PyCUDA is not available. It then uses PyCUDA instead of the libraries to detect if a GPU is available and accessible to use with the ASSET function.

This was the behavior enforced when setting flags for the tests in `test_asset.py`. Therefore, the flags are now set to use the utility functions.

Finally, the failed `test_asset_choose_backend_opencl` was implemented as a regression test to check the conditions where no GPU was available and no environmental variable flags were set, which in the past resulted in trying to use the OpenCL backend instead of the CPU. Therefore, in a CI runner with GPU/CUDA, the test is not intended to run, as the default behavior in Elephant is to use the CUDA GPU acceleration and, hence, the backend selection returned `'cuda'` correctly. Therefore, the second fix in the PR conditions the test execution on not having PyCUDA and a GPU.

### DIFF
--- a/elephant/test/test_asset.py
+++ b/elephant/test/test_asset.py
@@ -505,6 +505,7 @@ class AssetTestCase(unittest.TestCase):
 
     #  regression test Issue #481
     #  see: https://github.com/NeuralEnsemble/elephant/issues/481
+    @unittest.skipIf(HAVE_CUDA, "CUDA available, will be used instead of CPU")
     def test_asset_choose_backend_opencl(self):
         class TestClassBackend(asset._GPUBackend):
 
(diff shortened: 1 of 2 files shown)


CozySocksAlways (approved):
Looks good to me. I tested the updated test_asset.py on a GPU server and confirmed that the unit tests pass successfully for both PyCUDA and PyOpenCL.

PR #678, opened 2025-11-10
Feature/partial First Pydantic Models for parameter validation

JanNolten:
1. Creation of schema structure
2. Creation of base scripts
3. Creation of Pydantic Models for elephant.statistics, elephant.spike_train_correlation and elephant.spike_train_synchrony functions and classes
4. Creation of tests
5. Bug fixes
7. Implemented validation for elephant.statistics, elephant.spike_train_correlation and elephant.spike_train_synchrony functions and classes


elephant/schemas/function_validator.py
```diff
@@ -21,7 +21,7 @@ def wrapper(*args, **kwargs):
             validated = model_class(**data)
 
             # Call function with validated data unpacked
-            return func(**validated.model_dump())
+            return func(*args, **kwargs)
```
JanNolten: Because no normalization happens in the Pydantic models. It is safer to just pass the original arguments

elephant/statistics.py
```diff
@@ -102,9 +105,12 @@
     "optimal_kernel_bandwidth"
 ]
 
-cv = scipy.stats.variation
+@validate_with(PydanticCv)
+def cv(*args, **kwargs):
+    return scipy.stats.variation(*args, **kwargs)
```
JanNolten: This is the simplest way I found, to use a decorator for cv, but maybe it could be done without changing the previous code.

elephant/schemas/function_validator.py
```diff
@@ -29,4 +31,10 @@ def wrapper(*args, **kwargs):
...
         return wrapper
-    return decorator
\ No newline at end of file
+    return decorator
+
+def activate_validation():
+    skip_validation = False
+
+def deactivate_validation():
+    skip_validation = True
```
JanNolten: Should also be able to disable the validation globally, so it does not get annoying, if you never want to use it.

Issue #679, opened 2025-11-19
[Bug] Underflow: Negative value in van_rossum_distance function


xxcStella:
**Describe the bug**

When uses the **van_rossum_distance funtion** in **elephant.spike_train_dissimilarity**, it may generate "nan" result and give error like "RuntimeWarning: invalid value encountered in sqrt return np.sqrt(XXX"). This is because there is no zero-checking function before squaring root, making it possible to square root an extremely small number (e.g. 1e-16).

**To Reproduce**
1. Give a spike train with small values:
```python
  import neo, quantities as pq
  from elephant.spike_train_dissimilarity import van_rossum_distance

  tau = 0.1 * pq.s
  a = neo.SpikeTrain([0.1782, 0.2286, 0.2804, 0.4972, 0.5504]*pq.s, t_start=0*pq.s, t_stop=4*pq.s)
  print(van_rossum_distance([a, a], tau)[0, 1])
```

```
nan
RuntimeWarning: invalid value encountered in sqrt return np.sqrt(vr_dist)
```

**Expected behavior**

In theory, the result should be 0. But the value is `nan` and the code gives error: `RuntimeWarning: invalid value encountered in sqrt return np.sqrt(vr_dist)`

**Solution**
Before the **return** in **van_rossum_distance** function, add `vr_dist[vr_dist < 0] = 0.0`. This can clip any value lower than 0.

**Environment**
 - OS: Windows 11
 - How you installed elephant (`conda`, `pip`, source): pip
 - Python version: 3.8.20
 - `neo` python package version: 0.13.3
 - `elephant` python package version: 1.1.1 (latest)

PR #680, opened 2025-11-27
Clip small negative values in van_rossum_distance before np.sqrt

CozySocksAlways:
This PR fixes Issue #679 . Currently `van_rossum_distance` could return `nan` due to extremely small negative values because of  floating point errors before applying `np.sqrt`.

**Fix**
Added check before `np.sqrt` to find small negative values, then replaced them using `np.maximum(vr_dist,0)`. Also outputs runtime warning for transparancy.
This ensures that spike trains that should result in a distance of 0 no longer produce `nan`.

**Test**
Added unit test that reproduces the spike train scenario from the issue, it asserts that the computed distance is 0 and not `nan`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -514,6 +514,7 @@ jobs:
           # create an environment and install everything
           python -m venv ~/test_env
           source ~/test_env/bin/activate
+          sudo apt-get update
           sudo apt install -y libopenmpi-dev openmpi-bin
 
           python -m pip install --upgrade pip
@@ -530,4 +531,5 @@ jobs:
       - name: Run doctests
         run: |
           source ~/test_env/bin/activate
+          printf "def pytest_configure(config):\n    import numpy as np\n    import builtins\n    np.set_printoptions(legacy='1.25')\n    builtins.np = np\n" > conftest.py
           pytest elephant --doctest-modules --ignore=elephant/test/
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ spike trains, LFP, analog signals. The input-output data format is either
 
 * Documentation: https://elephant.readthedocs.io/en/latest/
 * Mailing list: https://groups.google.com/group/neuralensemble
+* Maintainer: contact[at]python-elephant[dot]org
 
 
 #### Visualization of Elephant analysis objects
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -31,7 +31,6 @@ sys.path.insert(0, '..')
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.autosummary',
-    'sphinx.ext.doctest',
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx.ext.viewcode',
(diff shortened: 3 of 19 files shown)


elephant/test/test_spike_train_dissimilarity.py
```diff
@@ -599,7 +600,15 @@ def test_van_rossum_distance(self):
         self.assertEqual(stds.van_rossum_distance(
             [self.st21], self.tau3)[0, 0], 0)
         self.assertEqual(len(stds.van_rossum_distance([], self.tau3)), 0)
-
+        
+        # Check small negative values edge case
+        with warnings.catch_warnings(record='True') as w:
```
kohlerca: The expected type for argument `record` is `bool`, not `str`

elephant/test/test_spike_train_dissimilarity.py
```diff
@@ -47,6 +47,7 @@ def setUp(self):
         self.st23 = StationaryPoissonProcess(rate=30 * Hz, t_start=0 * ms,
                                              t_stop=1000 * ms
                                              ).generate_spiketrain()
+        self.st24 = SpikeTrain([0.1782, 0.2286, 0.2804, 0.4972, 0.5504], units='s',t_stop=4.0)
```
kohlerca: As suggested by the dedicated regression test, the spike train can be initialized locally in the function, which facilitates code reading.

elephant/spike_train_dissimilarity.py
```diff
@@ -363,6 +363,15 @@ def van_rossum_distance(spiketrains, time_constant=1.0 * pq.s, sort=True):
     for i, j in np.ndindex(k_dist.shape):
         vr_dist[i, j] = (
             k_dist[i, i] + k_dist[j, j] - k_dist[i, j] - k_dist[j, i])
+    
+    # Clip small negative values
+    if np.any(vr_dist < 0):
+        warnings.warn(
+                "van_rossum_distance: very small negative values encountered "
```
kohlerca: Here, it would also be helpful to hint at possible causes. As we tested, this does not happen if the units are in ms, and the user could change the representation to prevent the warning if noticed. Maybe something like "... very small negative values encountered, setting them to zero. This is likely due to floating point error, which can happen if the spike times are small floating point values. In this case, a possible solution to prevent the warning is to change the time unit of the input spike trains for better precision (e.g., from seconds to milliseconds).".

elephant/spike_train_dissimilarity.py
```diff
@@ -363,6 +363,18 @@ def van_rossum_distance(spiketrains, time_constant=1.0 * pq.s, sort=True):
...
+
+    # Clip small negative values
+    if np.any(vr_dist < 0):
+        warnings.warn(
+            "van_rossum_distance: very small negative values encountered; "
+            "setting them to zero. Potentially due to floating point error, "
+            "which can occur if spike times are represented as small floating "
+            "point values (e.g., in seconds). A possible way to prevent this "
+            "warning is to use a time unit with better numerical precision, "
+            "e.g. seconds to milliseconds.", RuntimeWarning)
```
kohlerca: The revised warning looks good. I have a small suggestion for improved grammar `e.g., from seconds` instead of `e.g. seconds`

Issue #526, opened 2022-11-11
[Feature] Create module that loads example datasets


mdenker:
**Feature**
Based on the `datasets` module, which handles downloading the datasets from `datasets.python-elephant.org`, one could image to have a function that is more intuitive for a user to download example datasets to play around with. For example, one could image something like:
```
elephant.datasets.download_tutorial_data(13)
```
or
```
elephant.datasets.download_tutorial_data("some_telling_name_for_the_dataset")
```

PR #687, opened 2026-02-07
[Enh] Function to load example or test datasets

kohlerca:
Currently, many tutorials contain boilerplate code to fetch data from the `elephant-data` repository at http://datasets.python-elephant.org, using the `download_datasets` function in module `elephant.datasets`. This also requires additional code to instantiate Neo IO objects and retrieve the required data (e.g., load `neo.Block` and access a specific `neo.Segment`). Altogether, this makes the initial parts of the notebooks verbose. Two examples are the ASSET and Unitary Events tutorial notebooks.

In addition, some tutorials may reuse data generation functions that are shared with unit tests (for example, the Granger Causality tutorial). This might result in code redundancy.

This PR implements a new `load_data` function into `elephant.datasets` that allows directly retrieving or generating the needed data using a single keyword (e.g., `load_data('asset')`). 

The function works with a mapping `ELEPHANT_DATA` dictionary defined in the module, where two options for a given keyword are possible:
- a dictionary, whose keys correspond to arguments for the `download_datasets` function (`repo_path` and `checksum`, allowing the download from `elephant-data`), and an additional `loader` key that can store an optional callable that is used to retrieve the correct data object (e.g., the first segment in the `neo.Block` stored in the dataset)
- a callable, that directly generates the needed data.

Unit tests were expanded for `download_datasets` to verify the expected behavior, and new tests were added to validate `load_data`. The corresponding tutorial notebooks or unit test modules that share data with tutorials were also changed. A new section `Example datasets` is added to the `Miscellaneous` part of the `Function Reference by Module` section of the Elephant documentation page.

This PR did not change any code that downloads datasets for the sole purpose of unit tests, such as the tests for spectral Granger causality functions. These can be expanded later, perhaps coupled to additional tutorial notebooks or improving the testing framework with Pytest fixtures.

This PR addresses #526.

### DIFF
--- a/elephant/datasets.py
+++ b/elephant/datasets.py
@@ -310,9 +310,10 @@ def download_datasets(repo_path, filepath=None, checksum=None,
         temporary directory with the same name as in `elephant-data`.
         Default: None
     checksum : str, optional
-        MD5 hash of the file to use as checksum to verify data integrity after
-        download. If None, no integrity check is performed. If defined and the
-        check fails, an exception is raised.
+        MD5 hash of the file to use as checksum to verify data integrity of a
+        file in a local folder or after its download. If None, no integrity
+        check is performed. If defined and the check fails, an exception is
+        raised.
         Default: None
     verbose : bool, optional
         If set to False, disable the progress bar.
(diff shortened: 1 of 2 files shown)


mdenker (changes requested):
Looks very good, found only a few very minor suggestions. Tests run through perfectly, and I think everything important is covered. The only thing I found missing was a checksum check for files taken from a local directory rather than a download.

PR #688, opened 2026-02-17
Add runtime kernel length check in `trim=True` instantaneous_rate

CozySocksAlways:
## Issue
When kernel length is larger than binned spiketrain length in `trim=True` case, it leads to ambiguous and incorrect offset in rate values, the following minimal code can trigger this case:
```
from elephant import statistics
from quantities import ms, s, Hz
from elephant.spike_train_generation import homogeneous_poisson_process
import numpy as np

np.random.seed(22)
sts = homogeneous_poisson_process(rate=10*Hz, t_start=0.*ms, t_stop=10000.*ms)
rate_estimate = statistics.instantaneous_rate( sts, sampling_period=50*ms, kernel="auto",
    t_start=sts.t_start, t_stop=sts.t_stop, trim=True,
)
``` 
## Fix
Enforces a kernel length check for `trim=True` case when computing instantaneous rates from binned spiketrains. Raises a Value error if the kernel is longer than the binned spiketrain.

## Cleanup
Removed  `t_stop` parameter passed to the ` neo.AnalogSignal ` constructor since it is not part of the initialization API and was being used as just an annotation.

## Tests
- Added decorator to skip `test_instantaneous_rate_spikes_on_edges` test, which incorrectly assumed rate magnitude is zero for all cases with edge spikes
- Added new tests:
1. `test_instantaneous_rate_trim_output_length`: checks that valid convolution produces the expected output length.
2. `test_instantaneous_rate_trim_rate_estimation`: checks that area under the curve for `trim=True` (valid convolution) is less than or equal to `trim=False` (same convolution) in order to verify edge trimming behavior.
3. `test_instantaneous_rate_trim_edges`: checks if t_stop for rate is calculated correctly with `trim=True`


elephant/statistics.py
```diff
@@ -1010,7 +1011,21 @@ def optimal_kernel(st):
...
     if trim:
+        # PR 688 Adding assertion on length of kernel
+        # A 'valid' convolution is only performed when:
+        # len(kernel) <= len(binned_spiketrain)
+        #
+        # This prevents ambiguous output lengths
+        # that occur when the kernel exceeds the binned spike train length.
+        if t_arr_kernel_length > n_bins:
+            raise ValueError(
+            f"Kernel length ({t_arr_kernel_length}) is longer than binned spiketrain "
```
CozySocksAlways: two words

elephant/test/test_statistics.py
```diff
@@ -690,34 +697,107 @@ def test_instantaneous_rate_small_kernel_sigma(self):
...
+                    # sampling_period=sampling_period,
+                    # kernel=kernel,
+                    # cutoff=cutoff, trim=True,
+                    # center_kernel=center_kernel)
+                # assert_array_almost_equal(rate.magnitude, 0, decimal=2)
+
+    def test_instantaneous_rate_trim_output_length(self):
+        # This test checks the output shape of rate with trim=True
+        # Based on known convolution result
+        # output shape = N - K + 1 (for valid conv)
```
CozySocksAlways: length instead of shape

elephant/test/test_statistics.py
```diff
@@ -690,34 +697,107 @@ def test_instantaneous_rate_small_kernel_sigma(self):
...
+                        t_stop=self.st_tr[1] * pq.s,
+                        trim=trim,
+                        center_kernel=center_kernel,
+                    )
+                    area_under_curve = spint.cumulative_trapezoid(
+                        y=rate_estimate.magnitude[:, 0],
+                        x=rate_estimate.times.rescale('s').magnitude)[-1]
+                    areas.append(area_under_curve)
+
+                self.assertTrue( areas[1] <= areas[0] or
```
CozySocksAlways: with the specific spiketrain, ti can be just less than

PR #695, opened 2026-04-23
Fix Spade compile issue

### LINKED ISSUE #689: MacOS wheel build failure for spade component
### Summary

Adding option to build wheels for macos using `build_wheels.yml' failing due to CI runner and header compatibility issues on GitHub.

Some experiments were performed using macOS CI environments, including trials with compiler path exports.

---

### Experiment Configuration Note

During trying out various options the following configuration was intentionally used because the wheel repair step was failing on macOS:
`CIBW_REPAIR_WHEEL_COMMAND_MACOS: ""`

This was done to bypass failing wheel repair execution during investigation.

---

### Plausible Root Cause

- Default macOS CI runner  lacks reliable support for the experimental C++ vector library used in `elephant/spade_src/include/FPGrowth.h>
- Attempts to install  a suitable compiler that support this header with Homebrew did not work.

---

### Archived Artifacts

CI logs and the workflow yaml file are archived under repository branch:
https://github.com/INM-6/elephant/tree/ci/archive-issue-macos-wheel-build

### ToDo

- [ ] Look into macOS C++ compilation strategies.
- [ ] Reduce dependency on manual CI environment configuration.

CozySocksAlways:
### Desc
Current Spade source code uses an experimental method `std::experimental::erase_if` which has been added as a standard library method in c++20. However spade src code is currently compiled with only c++17 causing newer runners such as macos-latest and windows-latest on github to fail as they deprecated the c++ library that contained the experimental methods.

Several other compiler warnings also came up with clang during this fix.

This will resolve issue #689 

### Changes

**C++ fixes in `spade_src`:**
- Replaced `std::experimental::erase_if` with an erase-remove idiom in `FPGrowth.h`, keeping the code compliant with C++17
- Replaced `high_resolution_clock` with `system_clock` uniformly across all platforms in `Timer.h`, `high_resolution_clock`  is just an alias that resolves to `steady_clock` or `system_clock`, unfortunately on macOS it resolves to `steady_clock` which lacks `to_time_t` method, causing a compilation error
- Added a `return 0` fallback in `Utils.h` as a no-op for the macOS branch of `GetCurrentRSS` to fix a `-Wreturn-type` clang warning
- Extended the `-Wmissing-field-initializers` pragma suppression in `fim.cpp` to cover clang as well
- Added pragma suppression for `-Wunused-private-field` warning around `m_size` attribute in `ClosedDetect.h`

**`setup.py` fixes:**
- Fixed OpenMP flag for Windows: `-fopenmp` to `-openmp` (MSVC syntax)
- Added `"Darwin"` to the platforms that include the `fim` extension module, so macOS builds now compile the C extension

**CI (`build_wheels.yml`):**
- Added `macos-15-intel` runner to the matrix to cover x86 builds alongside the arm64 runner (`macos-latest`)
- Fixed OpenMP discovery on macOS by using brew to dynamically get omp library paths during compile
- Set `MACOSX_DEPLOYMENT_TARGET` conditionally per architecture: `15.0` for ARM64, `14.0` for Intel
- Removed separate macOS and Linux+Windows split, now using a single build wheel task
- Added a post-build verification step that installs the built wheel and checks if compiled fim*.so is available.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -50,6 +50,7 @@ jobs:
       - name: Build wheels (macOS)
         if: runner.os == 'macOS'
         run: |
+          echo ${{env.LDFLAGS}}
           python -m pip install --upgrade pip setuptools wheel build
           python -m build --wheel --outdir wheelhouse
         env:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-15-intel]
 
     steps:
       - uses: actions/checkout@v4
@@ -47,19 +47,24 @@ jobs:
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9"
           CIBW_ARCHS: "auto64"
 
+      - name: Set macOS env vars
+        if: runner.os == 'macOS'
+        run: |
+          OMP_PATH=$(brew --prefix libomp)
+          echo "LDFLAGS=-L${OMP_PATH}/lib" >> $GITHUB_ENV
+          echo "CPPFLAGS=-I${OMP_PATH}/include" >> $GITHUB_ENV
+          if [ "$RUNNER_ARCH" = "ARM64" ]; then
+            echo "MACOSX_DEPLOYMENT_TARGET=15.0" >> $GITHUB_ENV
+          else
+            echo "MACOSX_DEPLOYMENT_TARGET=14.0" >> $GITHUB_ENV
+          fi
+
       - name: Build wheels (macOS)
         if: runner.os == 'macOS'
         run: |
-          echo ${{env.LDFLAGS}}
-          echo ${{ env.CPPFLAGS }}
-          echo ${{ env.CPLUS_INCLUDE_PATH }}
           python -m pip install --upgrade pip setuptools wheel build
           python -m cibuildwheel --output-dir wheelhouse
         env:
-          LDFLAGS: "-L/opt/homebrew/opt/libomp/lib ${{ env.LDFLAGS }}"
-          CPPFLAGS: "-I/opt/homebrew/opt/libomp/include ${{ env.CPPFLAGS }}"
-          CPLUS_INCLUDE_PATH: "/opt/homebrew/opt/libomp/include:${{ env.CPLUS_INCLUDE_PATH }}"
-          MACOSX_DEPLOYMENT_TARGET: "15.0"
           CIBW_ARCHS: "auto"
 
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -54,7 +54,7 @@ jobs:
           echo ${{ env.CPPFLAGS }}
           echo ${{ env.CPLUS_INCLUDE_PATH }}
           python -m pip install --upgrade pip setuptools wheel build
-          python -m build --wheel --outdir wheelhouse
+          python -m cibuildwheel --output-dir wheelhouse
         env:
           LDFLAGS: "-L/opt/homebrew/opt/libomp/lib ${{ env.LDFLAGS }}"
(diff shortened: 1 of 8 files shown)


kohlerca:
I will have a look.

elephant/spade_src/include/FPGrowth.h
```diff
@@ -44,7 +44,7 @@
 #include <mpi.h>
 #endif
 
-#include <experimental/vector>
+// #include <experimental/vector>
```
kohlerca: Maybe this can be left out to keep it clean, and just let the old command below for quick reference

CozySocksAlways:
Ready for review. Once review completed, the trigger for build_wheels.yml can be reverted to only on PR review submission and merge to master.

.github/workflows/build_wheels.yml
```diff
@@ -15,43 +17,57 @@ jobs:
...
-        if: runner.os == 'macOS'
+      - name: Verify wheels contain compiled code
         run: |
-          python -m pip install --upgrade pip setuptools wheel build
-          python -m build --wheel --outdir wheelhouse
+          # Download elephant and deps wheels to a DIR
+          mkdir -p localwheels
+          pip wheel --wheel-dir localwheels .
+
+          # Remove downloaded elephant and copy wheeels from cibuildwheel
```
kohlerca: Typo in `wheels`

kohlerca (changes requested):
I checked all the built wheels for each target system, and all binaries for FIM were included.  I also tested all Windows wheels in a 64-bit system, and the module worked as expected. Overall, the PR looks good. There is only a small typo in a comment to be corrected. With that, it can be merged.

kohlerca (approved):
The latest changes look good to me.

.github/workflows/build_wheels.yml
```diff
@@ -1,12 +1,14 @@
...
 on:
   push:
-    branches:
-      - master
+    branches: [ master ]
   pull_request_review:
     types: [ submitted ]
+  pull_request:
+    branches: [ master ]
+    types: [ synchronize ]
```
mdenker: If I understand correctly, this will build the wheels for each update to the PR branch. If that is the case, is this what we want, or is it enough to build the wheels after a successful review?

.github/workflows/build_wheels.yml
```diff
@@ -15,43 +17,56 @@ jobs:
...
 
-      - name: Build wheels (Linux / Windows)
-        if: runner.os != 'macOS'
+      - name: Build wheels
         run: python -m cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_SKIP: "cp38-*"
-          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9"
-          CIBW_ARCHS: "auto64"
+          CIBW_ARCHS: "auto"
```
mdenker, line 45: Should we synchronized wheel building with the minimum Python version from the unit testing? In that case we could perhaps keep the CIBW_PROJECT_REQUIRES_PYTHON? But I assume there were reasons for loosening constraints here.

mdenker: Also, I assume we are building for 32 bit now as well. Looking at the wheels, it seems this applied to Windows only, however, and I guess that sort of makes sense given these windows may still be around. So I think that's ok to switch from auto64 to auto.

mdenker (approved):
Thank you for the great work, this looks fine to me. I posted 2 small questions where I was not quite sure about the rationale behind the change, but in principle, this PR is good to go independent of that.

.github/workflows/build_wheels.yml
CozySocksAlways: Yup you are right, but I did this as a temporary measure with the commit d8ffec52454f7592eeacc32902e20c014f7fa5b1 for development and will revert it before merging

.github/workflows/build_wheels.yml
CozySocksAlways: So it's not actually loosening constraints here, but rather changing the source of truth to the toml file (which will be taken up in the next PR #608 ). And as a side note `CIBW_PROJECT_REQUIRES_PYTHON` is only meta data, wheel building is done based on `CIBW_SKIP`, which on next minor release can also be refactored to skip 3.9

As for `auto`, it technically resolves to `auto64` on 64 bit machines. We don't actually build for 32 bit. 'auto' is more generic, and will pick the preferred target arch based on the host runner.

PR #564, opened 2023-05-12
[Fix] spike time tiling coefficient for unsorted spiketrains, added validation test

Moritz-Alexander-Kern:
This is related to Issue #563 .

The example from the Issue now returns:

- STTC for E7 and E3: 1.0  (Spiketrains identical)
- STTC for E8 and B3: -0.02 (Spiketrains different)

Previously, this function encountered a bug when using unsorted spiketrains as input, leading to the generation of values exceeding the desired range of -1 to 1.

This fix involves several improvements: 
- the spiketrains used as input are now sorted 
- additional input checks were added 
- added a regression test, verifying that the sttc function returns values within the desired range of -1 to 1, with the example given in the Issue

Example Code:
```python
import neo

import quantities as pq

from elephant.spike_train_correlation import spike_time_tiling_coefficient

spiketrain_E7 = neo.SpikeTrain([ 1678. , 23786.3, 34641.8, 71520.7, 73606.9, 78383.3,
97387.9, 144313.4, 4607.6, 19275.1, 152894.2, 44240.1], units='ms', t_stop=160000*pq.ms)

spiketrain_E3 = neo.SpikeTrain([ 1678. , 23786.3, 34641.8, 71520.7, 73606.9, 78383.3,
97387.9, 144313.4, 4607.6, 19275.1, 152894.2, 44240.1], units='ms', t_stop=160000*pq.ms)

print(spike_time_tiling_coefficient(spiketrain_E7, spiketrain_E3, dt=0.10 * pq.s))

spiketrain_E8 = neo.SpikeTrain([ 20646.8, 25875.1, 26154.4, 35121. , 55909.7, 79164.8,
110849.8, 117484.1, 3731.5, 4213.9, 119995.1, 123748.1,
171016.8, 172989. , 185145.2, 12043.5, 185995.9, 186740.1,
12629.8, 23394.3, 34993.2], units='ms', t_stop=210000*pq.ms)

spiketrain_B3 = neo.SpikeTrain([ 10600.7, 19699.6, 22803. , 40769.3, 121385.7, 127402.9,
130829.2, 134363.8, 1193.5, 8012.7, 142037.3, 146628.2,
165925.3, 168489.3, 175194.3, 10339.8, 178676.4, 180807.2,
201431.3, 22231.1, 38113.4], units='ms', t_stop=210000*pq.ms)

print(spike_time_tiling_coefficient(spiketrain_E8, spiketrain_B3, dt=0.10 * pq.s))
```

## todo:
[...]

PR #706, opened 2026-08-05
Use an exact +- dt window in spike_time_tiling_coefficient

adityasingh2400:
`run_p` builds the synchronicity window with `np.isclose(t_j, t_i, atol=dt)`. `np.isclose` also applies its default relative tolerance `rtol=1e-5`, so the test performed is `|t_j - t_i| <= dt + 1e-5 * |t_i|` rather than the documented `|t_j - t_i| <= dt`.

Spike times are absolute, so the window grows with the distance of the recording from t = 0, and it is asymmetric between the two trains because `rtol` multiplies only the second argument. `run_t` computes the T terms with `dt` exactly, so P and T were being derived from different windows.

The result is a silently wrong coefficient rather than an error:

| scenario | dt | elephant | correct |
|---|---|---|---|
| two spikes 12 ms apart, recording starts at t=0 | 5 ms | -0.0005 | -0.0005 |
| same two spikes, recording starts at t=1000 s | 5 ms | +1.0 | -0.0005 |
| independent 5 Hz Poisson pair, 600 s, t_start=0 | 5 ms | +0.027 | -0.003 |
| same pair with wall-clock timestamps (t_start=86400 s) | 5 ms | +0.9998 | -0.003 |

The effective window is `dt + 1e-5*|t|`. A requested 5 ms is already 11 ms at the end of an ordinary 10 minute recording, and 869 ms at t = 86400 s.

This compares the times directly instead. It matches the documented `[-dt, +dt]`, it matches what `run_t` already does, and it makes the result invariant under a time shift of the whole recording. `test_sttc_validation_test`, which checks against C. Cutts' original C implementation, still passes.

The relative tolerance came in with the vectorization in `8bac14c` (#564) and looks like an artifact of moving to `np.isclose` rather than a deliberate tolerance.

Tested with two new cases in `elephant/test/test_spike_train_correlation.py`, one pinning the window to exactly dt and one asserting invariance under a time shift. Reverting `elephant/spike_train_correlation.py` to master makes both fail, the first with `np.float64(1.0) != -0.0005`. The full file passes after at 37 passed and 12 subtests. `pycodestyle` is clean.

Note the module doctest shows `np.float64(0.4958601655933762)` against an expected `0.4958601655933762`. The value is identical and the difference is repr only, which is known issue #686, and CI forces legacy numpy printing.

I have not added myself to `doc/authors.rst`. It is a numbered institutional affiliation list and I did not want to presume, so please let me know if you would like an entry there.

### DIFF
--- a/elephant/spike_train_correlation.py
+++ b/elephant/spike_train_correlation.py
@@ -900,10 +900,14 @@ def spike_time_tiling_coefficient(spiketrain_i: neo.core.SpikeTrain,
         """
         # Create a boolean array where each element represents whether a spike
         # in spiketrain_j lies within +- dt of any spike in spiketrain_i.
-        tiled_spikes_j = np.isclose(
-            spiketrain_j.times.magnitude[:, np.newaxis],
-            spiketrain_i.times.magnitude,
-            atol=dt.item())
+        # The comparison is done explicitly rather than with `np.isclose`:
+        # `np.isclose` also applies its default relative tolerance rtol=1e-5,
+        # which widens the synchronicity window by 1e-5 * |spike time|. Because
+        # spike times are absolute, that made the window grow with the distance
+        # of the recording from t = 0 instead of staying at the requested dt.
+        tiled_spikes_j = np.abs(
+            spiketrain_j.times.magnitude[:, np.newaxis]
+            - spiketrain_i.times.magnitude) <= dt.item()
         # Determine which spikes in spiketrain_j satisfy the time window
         # condition.
         tiled_spike_indices = np.any(tiled_spikes_j, axis=1)
--- a/elephant/test/test_spike_train_correlation.py
+++ b/elephant/test/test_spike_train_correlation.py
@@ -839,6 +839,43 @@ class SpikeTimeTilingCoefficientTestCase(unittest.TestCase):
                                     spiketrain_B3, dt=0.10 * pq.s)
         self.assertAlmostEqual(sttc_unsorted_E8_B3, sttc_sorted_E8_B3)
 
+    def test_sttc_synchronicity_window_is_exactly_dt(self):
+        # Two spikes 12 ms apart are not synchronous for dt = 5 ms, no matter
+        # how far the recording is from t = 0. Previously the window was built
+        # with np.isclose, whose default rtol=1e-5 widened it by
+        # 1e-5 * |spike time|, so the same pair of spikes was counted as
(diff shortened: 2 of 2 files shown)


adityasingh2400:
The three red jobs here are not from this branch.

All three fail on the same two tests, `test_datasets.py::TestDownloadDatasets::test_valid_data_with_integrity_check` and `test_valid_data_with_failed_integrity_check`, both with `SSLCertVerificationError: certificate verify failed: unable to get local issuer certificate` while fetching the sample datasets over HTTPS. The rest of the run is green at 773 passed, 5 skipped, including the spike train correlation tests this PR touches.

This branch only changes `elephant/spike_train_correlation.py` and its test module, nothing under `elephant/datasets` and nothing involved in downloading. A rerun once the certificate or network issue clears should be all it needs.

adityasingh2400:
Small addition to my earlier note: the `docs` job is red here too now, and it is the same remote-data problem seen from the sphinx side, where a notebook that downloads sample data fails during execution.

Also, the Coveralls number on this PR, about 45.7 percent against a base of 88.6, is consistent with the failed jobs never uploading their coverage rather than with a real regression. The passing run still reports 773 passed and 5 skipped, with only the two `TestDownloadDatasets` cases failing.

CozySocksAlways:
Hello there,
And thank you once again for 3 straight PRs :) :).  Same story here on the CI/coverage front as the other two. As mentioned I'll review them and post my comments here.

adityasingh2400:
Correcting my note above, since the rerun has now happened and did not do what I predicted.

I said a rerun should clear this once the network settled. Attempt 2 on `8075b6a` cleared `pip` and `conda` but not `mpi` or `docs`, and those two are now separate problems.

`mpi` is still the data host, but refusing rather than dropping. Five `TestDownloadDatasets` cases fail with `HTTP Error 403: Forbidden`. `git ls-remote https://gin.g-node.org/NeuralEnsemble/elephant-data.git` succeeds from my machine right now, returning HEAD `666b920d`, so the host is up and is refusing the runner specifically. `pip` and `conda` passed on that same run, which looks more like rate limiting than a block.

`docs` is no longer a fetch failure at all:

```
```

That is what #709 pins `ipython<9.16` for, and `docs` plus `readthedocs` are both green on that branch. #708 hits the same import. My note on #707 saying its docs job would pass once the remote recovered was wrong for the same reason, that run predates the IPython release.

